PR #10, opened 2020-06-12
Update the top-level README.md to point directly to build/run instructions

mkavulich:
## DESCRIPTION OF CHANGES: 
Now that this is the top-level app that all users will check out, this README file will be where users may look for build/run instructions. We can do this by pointing directly to the quick-start guide on the wiki (https://github.com/ufs-community/ufs-srweather-app/wiki/Getting-Started) rather than the wiki page, where the build/run instructions are harder to find.

## TESTS CONDUCTED: 
None required, text-only change.


mkavulich:
@JulieSchramm This was in response to comments by the DA group who found it difficult to find the actual instructions on building and running. The wiki is kind of a wall of text and the "Getting started" guide was just one link in the middle of a paragraph on that page.

I made the link more prominent, but maybe this README should just point directly to that page? https://github.com/ufs-community/ufs-srweather-app/wiki/Getting-Started

JulieSchramm:
Pointing directly to the wiki/Getting-Started is a good compromise.  Cecila
DeLuca set up the wiki to look like Mr Wx App.

PR #23, opened 2020-09-10
Update manage_externals for Python 3+ bug fix, support for 3.7 and 3.8

JeffBeck-NOAA:
## DESCRIPTION OF CHANGES: 
Updates manage_externals to the most recent release which includes a critical Python 3+ bug fix and supports 3.7 and 3.8.  This new version avoids the previous incompatibility between manage_externals and the YAML namelist generation/Jinja templating for XML and diag_table files.  Users will no longer have to load different versions of Python in-between manage_externals and running the workflow.

## TESTS CONDUCTED: 
Tested successfully on Hera.  All repos are cloned correctly.  @JulieSchramm tested on Cheyenne.

### DIFF
--- a/manage_externals/.gitignore
+++ b/manage_externals/.gitignore
@@ -12,3 +12,6 @@ components/
 
 # generated python files
 *.pyc
+
+# test tmp file
+test/tmp
--- a/manage_externals/.travis.yml
+++ b/manage_externals/.travis.yml
@@ -1,7 +1,3 @@
-# NOTE(bja, 2017-11) travis-ci dosen't support python language builds
-#  on mac os. As a work around, we use built-in python on linux, and
-#  declare osx a 'generic' language, and create our own python env.
-
 language: python
 os: linux
 python: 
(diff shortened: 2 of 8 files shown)


JulieSchramm:
I still need to run

ncar_pylib /glade/p/ral/jntp/UFS_CAM/ncar_pylib_20200427

on Cheyenne to generate the workflow.  Is this expected behavior?

JeffBeck-NOAA:
@JulieSchramm If that Cheyenne command is to load a version of Python for both manage_externals and the workflow, then yes, that's still required.

JulieSchramm:
I can run manage_externals without the ncar_pylib command and it works
fine, then I need the ncar_pylib command to run ./generate_FV3SAR_wflow.sh,
as we were doing before this PR.

JeffBeck-NOAA:
@JulieSchramm That sounds like normal behavior.  Did you previously need the "ncar_pylib" command for manage_externals?  The Python version required for the generate script may not be the default on Cheyenne, and may still require you to run the ncar_pylib command.

JulieSchramm:
I didn't need the ncar_pylib command previously to run manage_externals.

JeffBeck-NOAA:
@JulieSchramm OK, then everything is working as expected.  Thanks for testing!

PR #32, opened 2020-10-14
Remove hard-coded executables directory names

gsketefian:
## DESCRIPTION OF CHANGES:
* Remove hard-coded executable directory names by replacing them with the new variable BIN_DIR that gets set only in one place.
* This removes the bug where an empty "exec" directory is created in addition to the "bin" directory in which the executables are placed.

## TESTS CONDUCTED: 
Built on hera successfully.  The executables are identical to the ones created before these changes, and an empty "exec" directory is no longer created.

### DIFF
--- a/src/build_post.sh
+++ b/src/build_post.sh
@@ -11,11 +11,6 @@ else
   export MOD_PATH=${cwd}/lib/modulefiles
 fi
 
-# Check final exec folder exists
-if [ ! -d "../exec" ]; then
-  mkdir ../exec
-fi
-
 cd EMC_post
 
 if [ "$target" = "jet" ] ; then
--- a/src/build_all.sh
+++ b/src/build_all.sh
@@ -32,7 +32,7 @@ fi
 # Set the full path of the directory where the binaries (executables) 
 # will be placed (and where the workflow scripts will look for them).
 #
-BIN_DIR=$( readlink -m "../bin" )
+export BIN_DIR=$( readlink -m "../bin" )
 #
 # If the binaries directory doesn't already exist, create it.
 #
--- a/src/build_all.sh
+++ b/src/build_all.sh
@@ -28,10 +28,17 @@ if [ ! -d $logs_dir  ]; then
   mkdir $logs_dir
 fi
 
-# Check final bin folder exists
-if [ ! -d "../" ]; then
-  echo "Creating ../bin folder"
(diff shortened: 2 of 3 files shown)


gsketefian:
Can I have someone take a look at this PR? It's pretty straightforward. Thx.

JeffBeck-NOAA:
> Can I have someone take a look at this PR? It's pretty straightforward. Thx.

@gsketefian, I'll take a look.

src/build_post.sh
```diff
@@ -11,11 +11,6 @@ else
   export MOD_PATH=${cwd}/lib/modulefiles
 fi
 
-# Check final exec folder exists
-if [ ! -d "../exec" ]; then
-  mkdir ../exec
-fi
-
```
JeffBeck-NOAA, line 18: Shouldn't there still be a check for the ../bin directory?

JeffBeck-NOAA: Oh, it looks like this is now in the install_all.sh script?

gsketefian: It's not even in the install_all.sh script.  I assume that users are always going to run the top-level script build_all.sh, not the ones that it in turn calls (e.g. build_post.sh).  They can control which codes are build via the configuration file regional_build.cfg.

JeffBeck-NOAA: Yeah, I'm seeing that now.  Wasn't there a way previously that users could call the individual scripts in a stand-alone fashion?  I thought there was, but if we're going to require users to modify regional_build.cfg only, then run build_all.sh, I'm OK with that option.  I guess that should be explicitly mentioned in the documentation.

gsketefian: Well all these scripts are not even going to be in the release, so we don't need to document it.  I guess they're eventually going to removed from develop as well, so there's even less incentive to document!  I could put the directory check/create stanzas in the sub-scripts, but that requires retesting, which may not be worth it (since the configuration file already provides a mechanism for limiting what you want to build).

JeffBeck-NOAA: Nope, it's fine as is.  Approving now.

JeffBeck-NOAA (approved):

src/build_post.sh
gsketefian: Thanks, merging now.

PR #48, opened 2020-11-04
Release/public v1 Load module NCEPLIBS before module esmf

JulieSchramm:
## DESCRIPTION OF CHANGES: 
module load NCEPLIBS/2.0.0 before module load esmf/8.0.0

## TESTS CONDUCTED: 
Builds successfully on Cheyenne


climbfuji:
You should remove the explicit load of ESMF entirely - `ESMFMKFILE` is set in the NCEPLIBS umbrella module, and the `bin` directory is added to `PATH`. Can you try to remove it?

JulieSchramm:
Testing now....

RatkoVasic-NOAA:
Sure, right awaySent from my Samsung Galaxy smartphone.
-------- Original message --------From: Dom Heinzeller <notifications@github.com> Date: 11/4/20  4:08 PM  (GMT-05:00) To: ufs-community/ufs-srweather-app <ufs-srweather-app@noreply.github.com> Cc: Subscribed <subscribed@noreply.github.com> Subject: Re: [ufs-community/ufs-srweather-app] Release/public v1 Load module NCEPLIBS before module esmf (#48) 
You should remove the explicit load of ESMF entirely - ESMFMKFILE is set in the NCEPLIBS umbrella module, and the bin directory is added to PATH. Can you try to remove it?

mkavulich (approved):
Thanks for taking care of this Julie, I also tested these changes earlier and they are good to go in my book

mkavulich:
Somehow I did not see @climbfuji's comment, sorry about that. I thought that I had tested without loading esmf in the past and it did not work on Cheyenne, but I will wait to see what Julie's test finds.

JulieSchramm:
It doesn't build on Cheyenne without the module load esmf/8.0.0.  Something must be missing.

climbfuji:
What is the error message?

mkavulich:
Okay so it sounds like the issue I've seen in the past is still there. @climbfuji I found that, on Cheyenne, unless I manually specified ESMFMKFILE or loaded the esmf/8.0.0 module, the build failed. I was never quite sure why, because ESMFMKFILE should have been set by loading the NCEPLIBS module

climbfuji:
> Okay so it sounds like the issue I've seen in the past is still there. @climbfuji I found that, on Cheyenne, unless I manually specified ESMFMKFILE or loaded the esmf/8.0.0 module, the build failed. I was never quite sure why, because ESMFMKFILE should have been set by loading the NCEPLIBS module

In this case we should find out why and fix it. How can I reproduce this problem?

JulieSchramm:
@climbfuji You should be able to use the head of the release/public-v1 branch, which module loads ESMF before NCEPLIBS, and it will fail on Cheyenne.  Then make changes from here.  It is tedious since it takes to long to build.

climbfuji:
> @climbfuji You should be able to use the head of the release/public-v1 branch, which module loads ESMF before NCEPLIBS, and it will fail on Cheyenne. Then make changes from here. It is tedious since it takes to long to build.

This only works if I manually change the CCPP suites in `src/CMakeLists.txt`. We should really be pointing to hashes and not branches in `Externals.cfg`! This allows for controlled updates.

Anyway, doing this now and let the model compile.

climbfuji:
Superseded by #50, please close this PR #48.

climbfuji:
As an explanation for the issues you had, this is what I typed this morning and forgot to send off:

The wrong NCEPLIBS installation is used on cheyenne (according to `doc/README_cheyenne_{gnu,intel}.txt`). These must be the official installations in
```
module use /glade/p/ral/jntp/GMTB/tools/NCEPLIBS-ufs-v2.0.0/intel-19.1.1/mpt-2.19/modules
```
and
```
module use /glade/p/ral/jntp/GMTB/tools/NCEPLIBS-ufs-v2.0.0/gnu-9.1.0/mpt-2.19/modules
```
Otherwise you get all sorts of interesting problems with the ufs-weather-model's `CMakeLists.txt` is loading its `modulefiles/machine.compiler/fv3` module, which for Cheyenne points to the official installations. (You see several cmake reconfigure steps, and at some point the ufs-weather-model's `FV3/ccpp/CMakeLists.txt` says it can't find NetCDF.

Second, the ones used here are wrong, because `ESMFMKFILE` was not set correctly after building NCEPLIBS-external and before building NCEPLIBS. This information ends up in the NCEPLIBS umbrella module.

This line
```
setenv ESMFMKFILE "/glade/p/ral/jntp/UFS_SRW_app/temp/NCEPLIBS-ufs-v2.0.0/intel-19.1.1/mpt-2.19/lib64"
```
should be
```
setenv ESMFMKFILE "/glade/p/ral/jntp/UFS_SRW_app/temp/NCEPLIBS-ufs-v2.0.0/intel-19.1.1/mpt-2.19/lib64/esmf.mk"
```
This is correct in the official installation.

This was fixed in #50.

Issue #49, opened 2020-11-05
Top-level CMakeLists.txt: need to use cmake 3.15 or later


climbfuji:
https://github.com/ufs-community/ufs-srweather-app/blob/c6aaee2fbf470e1a926e7a10791037bbdde83338/CMakeLists.txt#L7

The ufs-weather-model however requires 3.15, I would assume that the srweather app needs to do the same.

climbfuji:
This was fixed in #50

PR #50, opened 2020-11-05
release/public-v1: fix Cheyenne build, update CCPP suites etc

climbfuji:
This PR fixes a number of issues with building the latest version of the ufs-weather-model and supersedes https://github.com/ufs-community/ufs-srweather-app/pull/48. It also fixes https://github.com/ufs-community/ufs-srweather-app/issues/49.

Changes:
- Use fixed hash instead of branch for ufs-weather-model in Externals.cfg
- CMakeLists.txt: cmake 3.15 is required by the submodules
- src/CMakeLists.txt: update CCPP suites
- Update documentation for cheyenne.intel and cheyenne.gnu
- Also: convert tabs in various `CMakeLists.txt` files into spaces and remove unnecessary CMake modules import (this directory doesn't exist)

This has been tested successfully on Cheyenne with the Intel compiler. Somebody else should try to test this with GNU on Cheyenne, and on other platforms if possible.

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -17,8 +17,8 @@ required = True
 protocol = git
 repo_url = https://github.com/ufs-community/ufs-weather-model
 # Specify either a branch name or a hash but not both.
-branch = release/public-v2
-#hash = 8165575
+#branch = release/public-v2
+hash = fa29a21
 local_path = src/ufs_weather_model
 required = True
 
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -2,14 +2,16 @@
 protocol = git
 repo_url = https://github.com/NOAA-EMC/regional_workflow
 # Specify either a branch name or a hash but not both.
-branch = release/public-v1
+#branch = release/public-v1
+hash = 71c2058
 local_path = regional_workflow
 required = True
 
 [ufs_utils]
 protocol = git
 repo_url = https://github.com/NOAA-EMC/UFS_UTILS
-tag = release/public-v2
+#branch = release/public-v2
+hash = 6891c8a
 local_path = src/UFS_UTILS
 required = True
 
@@ -26,8 +28,8 @@ required = True
 protocol = git
 repo_url = https://github.com/NOAA-EMC/EMC_post
 # Specify either a branch name or a hash but not both.
-branch = release/public-v2
-#hash =
+#branch = release/public-v2
+hash = c0899ed
 local_path = src/EMC_post
 required = True
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,13 +1,6 @@
 # This is the main cmake file for ufs-srweather-app.
-#
-# Ed Hartnett 5/14/20
 
-# This will use any cmake between 3.1 and 3.15, prefering later
-# versions with updated policies.
-cmake_minimum_required(VERSION 3.1...3.15)
-if (${CMAKE_VERSION} VERSION_LESS 3.12)
-  cmake_policy(VERSION ${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION})
-endif()
+cmake_minimum_required(VERSION 3.15)
 
 # Set the project name and version.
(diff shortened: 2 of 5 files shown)


JulieSchramm (approved):
GNU build works on Cheyenne.  Thanks for fixing this.

climbfuji:
> GNU build works on Cheyenne. Thanks for fixing this.

Yay. Thanks for trying it out.

mkavulich:
@climbfuji Can you incorporate the changes from #47 into this PR? Specifically, using `if(NOT CCPP_SUITES)` so that the default CCPP_SUITES value can be overwritten by the cmake command?

Externals.cfg
```diff
@@ -17,8 +17,8 @@ required = True
 protocol = git
 repo_url = https://github.com/ufs-community/ufs-weather-model
 # Specify either a branch name or a hash but not both.
-branch = release/public-v2
-#hash = 8165575
```
JeffBeck-NOAA, line 21: @climbfuji Is this how the MRW App handles their Externals.cfg file for the release?  Since we're still under rapid-development (even for bug fixes for the release), I'm concerned that this change will require us to constantly update Externals.cfg to stay up to date.  

Note that we've had discussions about creating a stable branch of regional_workflow that would contain specific hashes in the Externals.cfg file, while the develop branch would point to the HEAD of the branch.  New hashes would then have to be extensively tested before they're updated in the Externals.cfg file for the stable branch, while the develop branch would always point to the HEAD and users would be made aware that it may not work.

JeffBeck-NOAA:
> @climbfuji Can you incorporate the changes from #47 into this PR? Specifically, using `if(NOT CCPP_SUITES)` so that the default CCPP_SUITES value can be overwritten by the cmake command?

@climbfuji @mkavulich, I assume #47 be closed when those changes are migrated to this PR?

Externals.cfg
climbfuji: In my opinion a certain hash of the srweather-app code is guaranteed to work with exactly one hash of each of its submodules / externals, everything else is gambling. Yes, everytime you update one of the submodules you need to test it with the app and as part of the testing update the hash. The PR that is created to update it documents the successful testing with those particular hashes.

climbfuji: But note that you don't have to update to a new hash every time, you can do that whenever necessary and as such accumulate a few changes to the submodules (although this increases the risk to break stuff).

JeffBeck-NOAA: OK, that makes sense.  I see that the ufs-mrweather-app points to tags for each repo in the Externals.cfg file.  I'm assuming those correspond to specific hashes, so that would be equivalent.

climbfuji:
> @climbfuji Can you incorporate the changes from #47 into this PR? Specifically, using `if(NOT CCPP_SUITES)` so that the default CCPP_SUITES value can be overwritten by the cmake command?

Done, please check. If #50 gets merged, #47 should be flagged as merged, too (i.e. no need to close).

Externals.cfg
climbfuji: Yes, tags and hashes are equivalent in a sense that they are static; tags however can be updated by force to a new hash (re-tagging) - not recommended unless absolutely necessary.

climbfuji: Should I also change the branches to hashes for EMC_post and UFS_UTILS?

JeffBeck-NOAA: Yes, if you don't mind, that would be great.  I'll approve right afterward!

climbfuji: Done. I'll try to check out the PR again to make sure I got all the hashes correct.

JeffBeck-NOAA: OK, I'll preliminarily approve this PR.

JeffBeck-NOAA (approved):
Changes to Externals.cfg hashes and from PR #47 now incorporated, so approving.

Externals.cfg
climbfuji: The hashes do match the head of the branches at this point in time, so all good.

JeffBeck-NOAA: OK, merging!

PR #52, opened 2020-11-12
Update hash for ufs_utils to point to latest version of release/public-v2

JulieSchramm:
## DESCRIPTION OF CHANGES: 
This latest release builds only the executables needed by the SRW App.

## TESTS CONDUCTED: 
Built on Cheyenne with Intel and GNU compilers.  Intel out-of-the-box case workflow runs to completion, a fix to run with the GNU build is in PR "https://github.com/NOAA-EMC/regional_workflow/pull/363"

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -11,7 +11,7 @@ required = True
 protocol = git
 repo_url = https://github.com/NOAA-EMC/UFS_UTILS
 #branch = release/public-v2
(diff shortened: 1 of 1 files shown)


mkavulich (approved):
This looks good to me, we need this hash updated to remove the extra utilities from UFS_UTILS

PR #155, opened 2021-07-19
Add arguments to devbuild.sh script

danrosen25:
## DESCRIPTION OF CHANGES: 
* devbuild.sh documented with --help
* automatically determines compiler based on machine name
* added options for app, ccpp suite, enable options, disable options
* added option for build type DEBUG, RELEASE, RELWITHDEBINFO
* added option for build jobs, default 4
* added verbosity option for make

## TESTS CONDUCTED: 
Tested each devbuild.sh option in isolation. Tested on Jet (needs testing and confirmation of platform/compiler defaults on other machines)

## DEPENDENCIES:
None

## DOCUMENTATION:
```
Usage: ./devbuild.sh PLATFORM [OPTIONS]...

PLATFORM
      name of machine you are building on
      (e.g. cheyenne | hera | jet | orion | wcoss)

OPTIONS
  -h, --help
      show this help guide
  --compiler=COMPILER
      compiler to use; default depends on platform
      (e.g. intel | gnu | cray | gccgfortran)
  --app=APPLICATION
      weather model application to build
      (e.g. ATM | ATMW | S2S | S2SW)
  --ccpp="CCPP_SUITE1,CCPP_SUITE2..."
      CCCP suites to include in build; delimited with ','
  --enable-options="OPTION1,OPTION2,..."
      enable ufs-weather-model options; delimited with ','
      (e.g. 32BIT | INLINE_POST | UFS_GOCART | MOM6 | CICE6 | WW3 | CMEPS)
  --disable-options="OPTION1,OPTION2,..."
      disable ufs-weather-model options; delimited with ','
      (e.g. 32BIT | INLINE_POST | UFS_GOCART | MOM6 | CICE6 | WW3 | CMEPS)
  --continue
      continue with existing build
  --clean
      removes existing build; overrides --continue
  --build-dir=BUILD_DIR
      build directory
  --install-dir=INSTALL_DIR
      installation prefix
  --build-type=BUILD_TYPE
      build type; defaults to RELEASE
      (e.g. DEBUG | RELEASE | RELWITHDEBINFO)
  --build-jobs=BUILD_JOBS
      number of build jobs; defaults to 4
  -v, --verbose
      build with verbose output

NOTE: This script is for internal developer use only;
See User's Guide for detailed build instructions
```

## CONTRIBUTORS: 
@danrosen25


devbuild.sh
```diff
@@ -1,76 +1,235 @@
...
 usage () {
-  echo "Usage: "
-  echo "  $0 PLATFORM COMPILER"
-  echo ""
-  echo "PLATFORM: Name of machine you are building on"
-  echo "COMPILER: (optional) compiler to use; valid options are 'intel', 'gnu'"
-  echo ""
-  echo "NOTE: This script is for internal developer use only;"
-  echo "See User's Guide for detailed build instructions"
+  printf "Usage: $0 [OPTIONS]...\n"
```
gsketefian: You can actually group this all into one `printf` statement like this (and you won't need the \n's because they're part of your string):
```
printf "\
Usage: $0 [OPTIONS]...

OPTIONS
  -h, --help
...
NOTE: This script is for internal developer use only;
See User's Guide for detailed build instructions
"
```

devbuild.sh
```diff
@@ -1,76 +1,235 @@
...
-if [ ! -f "$ENV_FILE" ]; then
-  echo "ERROR: environment file ($ENV_FILE) does not exist for this platform/compiler combination"
-  echo "PLATFORM=$PLATFORM"
-  echo "COMPILER=$COMPILER"
-  echo ""
-  echo "See User's Guide for detailed build instructions"
+# set ENV_FILE for this platform/compiler combination
+ENV_FILE="${SRC_DIR}/env/build_${PLATFORM}_${COMPILER}.env"
+if [ ! -f "${ENV_FILE}" ]; then
+  printf "ERROR: environment file does not exist for this platform/compiler\n"
```
gsketefian: Combine `printf`s?  Same elsewhere.

devbuild.sh
```diff
@@ -1,76 +1,235 @@
...
-  echo "NOTE: This script is for internal developer use only;"
-  echo "See User's Guide for detailed build instructions"
+  printf "Usage: $0 [OPTIONS]...\n"
+  printf "\n"
+  printf "OPTIONS\n"
+  printf "  -h, --help\n"
+  printf "      show this help guide\n"
+  printf "  --platform=PLATFORM\n"
+  printf "      name of machine you are building on; defauts to hostname\n"
+  printf "      (.e.g. cheyenne | hera | jet | orion | wcoss)\n"
```
gsketefian: Why the leading dot in `.e.g.`?

devbuild.sh
```diff
@@ -1,76 +1,235 @@
...
+  printf "  --compiler=COMPILER\n"
+  printf "      compiler to use; default depends on platform\n"
+  printf "      (.e.g. intel | gnu | cray | gccgfortran)\n"
+  printf "  --app=APPLICATION\n"
+  printf "      weather model application to build\n"
+  printf "      (.e.g. ATM | ATMW | S2S | S2SW)\n"
+  printf "  --ccpp=\"CCPP_SUITE1,CCPP_SUITE2...\"\n"
+  printf "      CCCP suites to include in build; delimited with ','\n"
+  printf "  --components=\"COMPONENT1,COMPONENT2...\"\n"
```
gsketefian: What are components in this context?  Like the atmosphere component, ocean component, etc?  And what are the valid values that can be included in the list?

devbuild.sh
```diff
@@ -1,76 +1,235 @@
...
+  printf "  CLEAN=${CLEAN}\n"
+  printf "  CONTINUE=${CONTINUE}\n"
+  printf "  BUILD_TYPE=${BUILD_TYPE}\n"
+  printf "  BUILD_JOBS=${BUILD_JOBS}\n"
+  printf "  VERBOSE=${VERBOSE}\n"
+  printf "\n"
+}
 
+# find system name
+find_system () {
```
gsketefian: There is already a script (from EMC, I think) in the UFS_UTILS repo that does this.  It's at:
`ufs-srweather-app/src/UFS_UTILS/sorc/machine-setup.sh
`
To avoid code duplication, maybe you can use that here to get the platform instead of having this new function.  Or find_system can be a wrapper around machine_setup.sh that echos the final result.

devbuild.sh
danrosen25: @gsketefian I'm listing example platforms. I don't know if anyone plans to maintain the help document and keep it sync'd up with all the supported platforms. Do you want me to remove "e.g. " here? Everywhere?

devbuild.sh
danrosen25: @gsketefian This feature is actually the reason I modified the devbuild script. I found this helpful because this feature allows me to add components to the build without modifying the build script or any CMake files.  I will be adding a "WRFHYDRO" component for coupling RRFS-NWM. I tested this and it doesn't conflict with --app, so you can say --app=ATM --components=WW3 and it will build the standard ATM system and add WW3. I can add an example list here ("e.g. "). The link below gives you an example of the available build components.

[https://github.com/ufs-community/ufs-weather-model/blob/develop/CMakeLists.txt#L25](https://github.com/ufs-community/ufs-weather-model/blob/develop/CMakeLists.txt#L25)

devbuild.sh
danrosen25: @gsketefian Would you prefer that I did this through directing a block into cat (see following code) or write one large printf (as suggested)?

```
cat << EOF
Usage: $0 [OPTIONS]...

OPTIONS
  -h, --help
      show this help guide
...
EOF
```

devbuild.sh
danrosen25: My concern is that src/UFS_UTILS is an external that is dependent on using manage externals. I can write this subroutine to look for machine-setup.sh and fail if it doesn't find it or try to determine the system name on its own if it doesn't find it. What's your preference?

devbuild.sh
gsketefian: Oh sorry, the examples are fine.  I was just talking about the "." before the "e" in ".e.g.".  Should it just be "e.g."?  Don't remove the listed examples, they're very useful!

devbuild.sh
gsketefian: Yes, it would be great if you add some examples.  In addition, it would be good if you can mention the location of that CMakeLists.txt file so people can see what the latest components are.  Thx.

devbuild.sh
danrosen25: That's funny that I didn't even notice the leading ".". I will remove it.

devbuild.sh
gsketefian: The second approach sounds good to me -- look for machine-setup.sh, otherwise try to determine the machine on its own.

devbuild.sh
gsketefian: I like to use printf, but up to you.

devbuild.sh
gsketefian: There's 3 more of those below!

danrosen25:
@gsketefian I was looking at the ufs-weather-model CMakeList.txt and they use -DDEBUG to set CMAKE_BUILD_TYPE.  Do you want me to replace --build-type=? with --debug?
[https://github.com/ufs-community/ufs-weather-model/blob/develop/CMakeLists.txt#L95](https://github.com/ufs-community/ufs-weather-model/blob/develop/CMakeLists.txt#L95)

gsketefian:
> @gsketefian I was looking at the ufs-weather-model CMakeList.txt and they use -DDEBUG to set CMAKE_BUILD_TYPE. Do you want me to replace --build-type=? with --debug?
> https://github.com/ufs-community/ufs-weather-model/blob/develop/CMakeLists.txt#L95

I like --build-type because --debug sounds to me like a true/false flag (but isn't).

devbuild.sh
```diff
@@ -1,76 +1,235 @@
...
+CLEAN=false
+CONTINUE=false
+VERBOSE=false
+
+# process arguments
+while :; do
+  case $1 in
+    --help|-h) usage; exit 0 ;;
+    --platform=?*) PLATFORM=${1#*=} ;;
+    --platform) printf "ERROR: $1 requires an argument.\n"; usage; exit 1 ;;
```
gsketefian: Suggestion (that you don't have to take):  To make this list shorter (and have less repeated code), you can combine the cases without and with equal signs:
`--platform|--platform=) printf ...
`
You can also make a function for 
`printf ""ERROR: $1 requires an argument.\n"; usage; exit 1
`
and call that for the appropriate cases.

devbuild.sh
danrosen25: I went with `cat << EOF` so that I can type the usage and setting messages WYSIWYG. I kept other printf statements for lines that I wanted to indent.

devbuild.sh
danrosen25: Removed

devbuild.sh
danrosen25: Added examples and note about CMakeLIsts.txt

devbuild.sh
danrosen25: I'm now using machine-setup.sh unless the file is not found.

devbuild.sh
danrosen25: I've combined and cleaned up error messages.

devbuild.sh
danrosen25: @gsketefian I only combined printf statements for the large blocks in usage and settings. I like that the printf statements in ENV_FILE and BUILD_DIR are indented to follow the code. Is this okay?

gsketefian: Yes, no problem.

devbuild.sh
gsketefian: That's great, thanks!

gsketefian:
@danrosen25 This looks good now.  I guess it should be tested on one more machine, e.g. Hera.  I would volunteer but have some other things to finish.  I can get to it say Thursday, but if someone else can test it beforehand, that would be great.

src/CMakeLists.txt
```diff
@@ -11,19 +11,44 @@ if(NOT CCPP_SUITES)
...
+
+list(APPEND UFS_WEATHER_MODEL_ARGS
+  "-DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX}"
+  "-DCCPP_SUITES=${CCPP_SUITES}"
+  "-DCMAKE_C_COMPILER=${MPI_C_COMPILER}"
+  "-DCMAKE_CXX_COMPILER=${MPI_CXX_COMPILER}"
+  "-DCMAKE_Fortran_COMPILER=${MPI_Fortran_COMPILER}"
+  "-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}"
+  "-DNETCDF_DIR=$ENV{NETCDF}"
+  "-D32BIT=ON"
```
mkavulich, line 30: I like this modification to allow command-line arguments to CMake! While you're adding this, could we get the ability to have D32BIT and DINLINE_POST also be controlled by flags passed to CMake (similar to how you handled DATM)?

danrosen25: @mkavulich I can add any options you'd like. Just verify what you'd like the user interface to look like
--32bit=<ON|OFF>?
--inline-post=<ON|OFF>?

Just an FYI the --component=<COMPONENT1,COMPONENT2,...> just takes the transforms this list into -DCOMPONENT1=ON -DCOMPONENT2=ON, do you want me to change this to --enable-option=<ITEM1,ITEM2,...> and --disable-option=<ITEM1,ITEM2,...> (disable macro would set -DITEM1=OFF -DITEM2=OFF. Doing so would cover all weather model options.

JulieSchramm:
@danrosen25 I can test this on Hera and Cheyenne.  I am out Friday, and Cheyenne is down next week, so I will start with Hera.  Let me know when it is ready for testing.

danrosen25:
@JeffBeck-NOAA @JulieSchramm @mkavulich @gsketefian 
The only question I have right now is 
Do you want me to change `--component=<COMPONENT1,COMPONENT2,...>` to `--enable-option=<OPTION1,OPTION2,...>` and `--disable-option=<OPTION1,OPTION2,...>`, which better represents enabling and disabling options in [https://github.com/ufs-community/ufs-weather-model/blob/develop/CMakeLists.txt](https://github.com/ufs-community/ufs-weather-model/blob/develop/CMakeLists.txt).  This can be used to enable/disable any option for the ufs-weather-model including 32BIT and INLINE_POST

JulieSchramm:
When I run ``./devbuild.sh hera --build-dir=build_intel --install-dir=../bin_intel``, the build goes in the ``build`` directory and the executables are installed in the ``bin`` directory.  Is this the expected behavior?

JulieSchramm:
Oops, need the --platform.  I'm not a big fan of the machine-setup.sh script.  It figures out the machine name from the available disk names.  This caused problems when "yellowstone" became "cheyenne" and the disk names changed, and this had to be fixed in the many instances of this script.  Can the platform be a required argument, and then we wouldn't need machine-setup.sh?

danrosen25:
@JulieSchramm 
I removed the platform requirement. `./devbuild.sh hera` will throw an error because the argument is unknown. `./devbuild.sh` will automatically determine that you're running on hera or `./devbuild.sh --platform=hera` will manually specify hera.

`--build-dir=build_intel` will build into <CURRENT_DIRECTORY>/build_intel
and `--install-dir=../bin_intel` sets CMAKE_INSTALL_PREFIX=../bin_intel, which will install binaries into <CURRENT_DIRECTORY>/build_intel/../bin_intel/bin, libraries into <CURRENT_DIRECTORY>/build_intel/../bin_intel/lib, share into <CURRENT_DIRECTORY>/build_intel/../bin_intel/share, and include into <CURRENT_DIRECTORY>/build_intel/../bin_intel/include
Is that what you want?

JulieSchramm:
@danrosen25 When I type ``./devbuild.sh hera``, I don't get an error
message and the build starts.  The --build-dir and --install-dir work as
expected if I specify ``--platform=hera``.

danrosen25:
@JulieSchramm 
Oh, I see that I'm just ignoring arguments that don't start with `-`, so yes you can type `./devbuild.sh hera` but hera is ignored.  Do you want me to make PLATFORM a required argument since the machine_setup.sh script can be buggy? I would change usage to `Usage: .devbuild.sh PLATFORM [OPTIONS]...` and remove --platform=PLATFORM.

JulieSchramm:
@danrosen25 Making PLATFORM a required argument would be great.  Thanks for
doing this.

danrosen25:
@JulieSchramm I made `PLATFORM` a required argument.
@mkavulich I changed `--components` to `--enable-options` and `--disable-options`

src/CMakeLists.txt
danrosen25: I added --enable-options and --disable-options

mkavulich (approved):
Looks good, my concerns have been addressed. Once @JulieSchramm is okay with the updates I believe this PR is good to go.

JulieSchramm:
I will test on Cheyenne next week Aug 2 or 3 when the computer is back up.

JulieSchramm (approved):
Tested on Jet, Hera and Cheyenne.

PR #156, opened 2021-07-19
Feature/add build test

JulieSchramm:
## DESCRIPTION OF CHANGES: 
Add ``build.sh`` to test directory to provide the build step for automated testing.  A ``README.md`` file is included with information on how to run the build.sh script.

## TESTS CONDUCTED: 
build.sh script tested on Cheyenne (Intel and GNU), Hera (Intel) and Jet (Intel).  All builds succeeded.

## DEPENDENCIES:
None

## DOCUMENTATION:
Provided in the README.md file.  Will be included in the User's Guide when testing is added.

### DIFF
--- a/test/build.sh
+++ b/test/build.sh
@@ -1,0 +1,146 @@
+#!/bin/bash -l
+#=======================================================================
+# Description:  This script runs a build test for the
+#               UFS Short-Range Weather App. The executables
+#               built are listed below in $executables_created.
+#               A pass/fail message is printed at the end of the output.
+#
+# Necessary input parameters: machine name (jet hera or cheyenne)
+#
+# Usage:  see function usage below
+#
+# Examples: ./build.sh $machine >& test.out &
+#
+set -eux    # Uncomment for debugging
+#=======================================================================
+
+fail() { echo -e "\n$1\n" >> ${TEST_OUTPUT} && exit 1; }
+
+function usage() {
+  echo
+  echo "Usage: $0 machine | -h"
+  echo
+  echo "       machine       [required] is one of: ${machines[@]}"
+  echo "       -h            display this help"
+  echo
+  exit 1
+}
+
+machines=( hera jet cheyenne )
+
+[[ $# -eq 0 ]] && usage
+if [ "$1" = "-h" ] ; then usage ; fi
+
+export machine=${1}
+machine=`echo "${machine}" | tr '[A-Z]' '[a-z]'`  # scripts in sorc need lower case machine name
+
+#-----------------------------------------------------------------------
+# Check that machine is valid
+#-----------------------------------------------------------------------
+if [[ "${machines[@]}" =~ "$machine" ]]; then
+  echo "machine ${machine} is valid"
+else
+  echo "ERROR: machine ${machine} is NOT valid"
+  exit 1
+fi
+
+#-----------------------------------------------------------------------
+# Set compilers to be tested depending on machine
+#-----------------------------------------------------------------------
+if [ "${machine}" == "cheyenne" ] ; then
+  compilers=( intel gnu )
+else
+  compilers=( intel )
+fi
+
+#-----------------------------------------------------------------------
+# Set some directories
+#-----------------------------------------------------------------------
+PID=$$
+TEST_DIR=$( pwd )                   # Directory with this script
+TOP_DIR=${TEST_DIR}/..              # Top level (umbrella repo) directory
+TEST_OUTPUT=${TEST_DIR}/build_test${PID}.out
+
+build_it=0        # Set to 1 to skip build (for testing pass/fail criteria)
+#-----------------------------------------------------------------------
+# Create the output file if it doesn't exist
+#-----------------------------------------------------------------------
+if [ ! -f "$TEST_OUTPUT" ]; then
+   touch ${TEST_OUTPUT}
+fi
+
+cd ${TOP_DIR}
+
+ENV_DIR=${TOP_DIR}/env
+#-----------------------------------------------------------------------
+# Arrray of all executables built
+#-----------------------------------------------------------------------
+declare -a executables_created=( chgres_cube \
+                                 emcsfc_ice_blend \
+                                 emcsfc_snow2mdl \
+                                 filter_topo \
+                                 fregrid \
+                                 fvcom_to_FV3 \
+                                 global_cycle \
+                                 global_equiv_resol \
+                                 make_hgrid \
+                                 make_solo_mosaic \
+                                 ncep_post \
+                                 orog \
+                                 orog_gsl \
+                                 regional_esg_grid \
+                                 sfc_climo_gen \
+                                 shave \
+                                 ufs_model \
+                                 vcoord_gen )
+
+#-----------------------------------------------------------------------
+# Set up the build environment and run the build script.
+#-----------------------------------------------------------------------
+  for compiler in "${compilers[@]}"; do
+    BUILD_DIR=${TOP_DIR}/build_${compiler}
+    BUILD_OUTPUT=${BUILD_DIR}/build.out
+    BIN_DIR=${TOP_DIR}/bin_${compiler}
+    EXEC_DIR=${BIN_DIR}/bin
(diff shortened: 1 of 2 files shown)


test/build.sh
```diff
@@ -0,0 +1,146 @@
+#!/bin/bash -l
```
gsketefian: Does it work without -l?  That would be better.

gsketefian:
How is build.sh different from the devbuild.sh that @mkavulich put together?

test/build.sh
```diff
@@ -0,0 +1,146 @@
...
+  exit 1
+}
+
+machines=( hera jet cheyenne )
+
+[[ $# -eq 0 ]] && usage
+if [ "$1" = "-h" ] ; then usage ; fi
+
+export machine=${1}
+machine=`echo "${machine}" | tr '[A-Z]' '[a-z]'`  # scripts in sorc need lower case machine name
```
christinaholtNOAA: In general, the `$()` syntax is easier to read/debug.

test/build.sh
```diff
@@ -0,0 +1,146 @@
...
+if [ "${machine}" == "cheyenne" ] ; then
+  compilers=( intel gnu )
+else
+  compilers=( intel )
+fi
+
+#-----------------------------------------------------------------------
+# Set some directories
+#-----------------------------------------------------------------------
+PID=$$
```
christinaholtNOAA, line 59: Just for my enrichment, what does this syntax do?

test/build.sh
```diff
@@ -0,0 +1,146 @@
...
+                                 vcoord_gen )
+
+#-----------------------------------------------------------------------
+# Set up the build environment and run the build script.
+#-----------------------------------------------------------------------
+  for compiler in "${compilers[@]}"; do
+    BUILD_DIR=${TOP_DIR}/build_${compiler}
+    BUILD_OUTPUT=${BUILD_DIR}/build.out
+    BIN_DIR=${TOP_DIR}/bin_${compiler}
+    EXEC_DIR=${BIN_DIR}/bin
```
christinaholtNOAA, line 103: Just curious why a subdirectory is needed.

JulieSchramm:
The -l is handy if you call this script from crontab.  I can remove it, and
it can be added in the crontab file if someone wants to use it that way.

And how is the build.sh different from the devbuild.sh that @danrosen25 is
putting together?

test/build.sh
```diff
@@ -0,0 +1,146 @@
...
+    if [ $build_it -eq 0 ] ; then
+      . ${ENV_DIR}/build_${machine}_${compiler}.env
+#-----------------------------------------------------------------------
+# Remove old and bin directories if they exist
+#-----------------------------------------------------------------------
+      if [ -d "${BUILD_DIR}" ] ; then rm -rf ${BUILD_DIR} ; fi
+      if [ -d "${BIN_DIR}" ] ; then rm -rf ${BIN_DIR} ; fi
+      mkdir ${BUILD_DIR}
+      cd ${BUILD_DIR}
+      cmake .. -DCMAKE_INSTALL_PREFIX=../bin_${compiler}
```
christinaholtNOAA: Can this line use `${BIN_DIR}` to reduce duplication of information?

christinaholtNOAA (approved):
Just a couple of very minor suggestions, and a quick question for my curiosity (hard to google).

JeffBeck-NOAA:
@JulieSchramm, @danrosen25's PR only improves the devbuild.sh script that @mkavulich created.  But I think @gsketefian is asking why we need this script in addition to devbuild.sh.

test/build.sh
JulieSchramm: Agree, will change.

test/build.sh
JulieSchramm: ``$$`` is a process ID (PID) used to create a unique output filename each time the script is run so the previous output is saved.

test/build.sh
JulieSchramm: In the top-level CMakeLists.txt file, the install directory is set as ``bin`` under what  ``-DCMAKE_INSTALL_PREFIX`` is set to.  To make multiple builds and save all the executables for testing, I set ``-DCMAKE_INSTALL_PREFIX=../bin_${compiler}``, which creates ``../bin_intel/bin``, ``../bin_gnu/bin``, etc.

test/build.sh
christinaholtNOAA: Thanks, @JulieSchramm!

test/build.sh
JulieSchramm: Yes!

JulieSchramm:
The build.sh script is for testing.  It builds all executables for all supported compilers on a given machine.  It checks for the existence of all executables and writes a pass/fail message at the end.  This will be the build step for Christina et al.'s testing.

gsketefian:
I'm wondering if it's possible to call the `devbuild.sh` script within the `build.sh` script introduced in this PR so that the same build script that is normally used to build the code is used to test the build, i.e. `build.sh` would be a wrapper around `devbuild.sh`.  @danrosen25 is adding a bunch of new arguments to `devbuild.sh` in PR #155.  If those changes aren't enough, `devbuild.sh` can probably be tweaked further so that it can be called by `build.sh` (instead of `build.sh` issuing its own `cmake` and `make` commands).

test/build.sh
danrosen25: If possible I think it's best to design the script to be independent of the shell invocation selected. So it's best to eliminate any dependence on the environment settings at the time the script is called.

JulieSchramm:
I would leave the decision about build.sh vs devbuild.sh to @christinaholtNOAA, as the group at GSL will be setting up the automated testing.

christinaholtNOAA:
I have no strong opinions about one way or the other. If any users rely on `devbuild.sh` for building, I could see how calling it (instead of cmake directly) would essentially be testing an additional piece of used code, which is always better than not testing code that is used. Maybe that is enough to tip the scale toward calling `devbuild.sh` instead of using cmake?

danrosen25:
Just a heads up the devbuild.sh changes will take arguments for
- `--platform=<PLATFORM>` , if omitted this will use the machine_setup.sh script
- `--compiler=<COMPILER>`, defaults to intel on cheyenne, hera, jet
- `--build-dir=<BUILD_DIR>`, the CMake build directory
- `--install-dir=<INSTALL_DIR>` the CMAKE_INSTALL_PREFIX directory, which is put into the build directory if you don't give a full path.

More can be added as needed. I'm trying to make a user|developer|sysadmin friendly build script. We needed a way to build different configurations of the model i.e. `--app=<APPLICATION>`

JulieSchramm:
All builds passed on Cheyenne, Jet and Hera.

test/build.sh
```diff
@@ -0,0 +1,137 @@
...
+if [ ! -f "$TEST_OUTPUT" ]; then
+   touch ${TEST_OUTPUT}
+fi
+
+cd ${TOP_DIR}
+
+ENV_DIR=${TOP_DIR}/env
+#-----------------------------------------------------------------------
+# Arrray of all executables built
+#-----------------------------------------------------------------------
```
JeffBeck-NOAA, line 77: Typo: Arrray to Array.

test/build.sh
```diff
@@ -0,0 +1,137 @@
...
+    if [ -f ${exec_file} ]; then
+      echo "SUCCEED: ${compiler} executable file ${exec_file} exists" >> ${TEST_OUTPUT}
+    else
+      echo "FAIL: ${compiler} executable file ${exec_file} does NOT exist" >> ${TEST_OUTPUT}
+      let "n_fail=n_fail+1"
+    fi
+  done
+  #-----------------------------------------------------------------------
+  # Rename build and bin dirs for next build
+  #-----------------------------------------------------------------------
```
JeffBeck-NOAA: I'm confused by this comment block above.  What code does it apply to?

test/build.sh
JulieSchramm: Good catch

test/build.sh
JulieSchramm: Removed

PR #166, opened 2021-08-25
Release/public v1

panll:
Changing the tag from ufs-1.0.0 to ufs-1.0.1 in Externals.cfg for pointing to regional workflow release tag ufs-1.0.1
Passed the test on Hera.

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -22,7 +22,7 @@ required = True
 [EMC_post]
 protocol = git
 repo_url = https://github.com/NOAA-EMC/EMC_post
-tag = upp-v9.0.0
+tag = upp_v9.0.1
 local_path = src/EMC_post
 required = True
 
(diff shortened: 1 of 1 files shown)


fossell:
@panll and @jwolff-ncar - UPP has created a bug fix release tag called "upp_v9.0.1".  This should point to hash 8cf4f3b. 
 Please update the Externals.cfg file for EMC_post to point to the updated take.  Also please be careful with the tag name.  In the current Externals.cfg the tag name for upp used a dash "-" but this has changed and the formal tag naming standard uses an underscore "_" in the tag name. Please make sure the tag name is exactly: upp_v9.0.1

panll:
I'll update that, thanks! @fossell @jwolff-ncar

panll:
Updated Externals.cfg with the post tag upp_v9.0.1 and tested on Hera. @fossell @jwolff-ncar @JeffBeck-NOAA

fossell (approved):
I am reviewing for the UPP tag modification only. The update to the EMC_post tag in Externals.cfg is correct.

jwolff-ncar (approved):
I think this can be approved, merged, and the testing team should commence as assigned.

PR #173, opened 2021-09-07
Fix for manage_externals.

christinaholtNOAA:
## DESCRIPTION OF CHANGES: 
The existing snapshot of the manage_externals package builds a Python dictionary of the git submodules of each package where the keys are the relative paths where the submodules are checked out on disk. The code that checks status of these paths is looking for the key to be the same as the section name. That is a bad assumption for submodules like CMEPS. 

This is why we might often see the last line of the checkout_externals display `u"CMEPS"` and then the utility didn't actually do anything else. 

With this fix, the user should be able to clone, run checkout_externals successfully, switch branches and re-run checkout_externals and get the appropriate hashes throughout.

The code in [ESMCI/manage_externals](https://github.com/ESMCI/manage_externals/blob/master/manic/externals_description.py) has a slightly different fix implemented, however, it doesn't seem that the logic surrounding that fix is necessary given the way the dictionary is populated (use of output from `git submodule status`). Unnecessary logic is less efficient, especially for dictionaries that are already O(1) operations, and can slow the process down for projects with a large submodule tree.

## TESTS CONDUCTED: 
Sequence went like this before fix:
- Run manage_externals/checkout_externals once on fresh clone.
- Run manage_externals/checkout_externals -S to see status of current checkout. Stopped running with u'CMEPS' and did nothing. (Same behavior has been observed on prior occasions without the-S flag)

Sequence after fix:
- Run manage_externals/checkout_externals once on fresh clone.
- Run manage_externals/checkout_externals -S to see status of current checkout. Loops through all .gitmodules submodules and then shows clean status of all externals.
- Checkout different branch. Run manage_externals/checkout_externals and all submodules are updated to expected hashes for new branches. Note: submodules that may not be tracked in current branch (e.g., python_utilities), will still exist on disk, untracked.

### DIFF
--- a/manage_externals/manic/externals_description.py
+++ b/manage_externals/manic/externals_description.py
@@ -253,7 +253,7 @@ def read_gitmodules_file(root_dir, file_name):
                                           ExternalsDescription.REPO_URL, url)
                 externals_description.set(sec_name,
                                           ExternalsDescription.REQUIRED, 'True')
-                git_hash = submods[sec_name]['hash']
+                git_hash = submods[path]['hash']
                 externals_description.set(sec_name,
                                           ExternalsDescription.HASH, git_hash)
 


JulieSchramm (approved):
Good catch, works as described.

llpcarson:
Would it be worthwhile to provide this change to the authoritative repository for their consideration?  Less code divergence, if it makes sense, even in a fairly static sub-tree.

mkavulich:
A couple questions:

1. In cases where Externals.cfg points to a branch rather than a hash, will this change allow manage_externals to pull recent changes to that branch that the user does not have locally? Currently manage_externals does not do this so it would be a great bonus to have.
2. With this change, will manage_externals be able to handle cases where local changes have been made to one or more of the repositories?

christinaholtNOAA:
@llpcarson I think that could work. It may also be nice to ensure that I'm not missing something with the logic that the instantiated. Does it makes sense to tag them here to see if my change makes sense compared to theirs?

@mkavulich 
1. Yes, it does seem that an update to a branch will be pulled with the next instantiation of checkout_externals. I doesn't seem to be able to tell you that the remote branch has additional changes in it when you do a status check, which could result in you getting unexpected changes from externals pointing to branches. 
2. A local change will not be overwritten. If a module is in state "modified", nothing will be updated (by default). The utility provides you several options for how to manage the rest of the externals in this case. For example, I wouldn't be able to pull updates from regional_workflow as long as I have local mods to ufs-weather-model without some additional flags like `--exclude`.

christinaholtNOAA:
@mkavulich 
One other scenario I tested (inadvertently) was modifying Externals.cfg and pulling the modified section. In this case, I edited Externals.cfg to point to my test branch of regional_workflow from a hash of develop. That worked as expected, too!

JulieSchramm:
Would this be a fix to the problem of having to "git clone" the entire umbrella each time the nightly build is run?  I couldn't find an automated way to update all sub-repositories without removing the code and re-cloning and running manage_externals.

llpcarson:
I would suggest opening an issue in the upstream repository
ESMCI/manage_externals
<https://github.com/ESMCI/manage_externals/blob/master/manic/externals_description.py>
and linking this PR in that issue along with the description of your
alternate solution.

christinaholtNOAA:
@JulieSchramm I think you might be right about that!

@llpcarson I can open the issue. Should I tag all of you on that?

christinaholtNOAA:
@llpcarson I was doing a couple of timing runs to compare the two implementations and they are really close in performance for these smaller dictionaries, even swapping "winners" when I time them together and cache hits help out the one that comes second. Maybe it's more important to just adopt the implementation they have in the current version of ESMCI?

I am not sure of the best way forward. Does the group prefer I just take the changes from that one PR that fixed this issue? Do the code managers of this repo want to do a full "merge" with the current manage_externals?

JeffBeck-NOAA:
After discussion of this PR at our weekly DTC UFS SRW App meeting, we've decided to pull in the latest manage_externals that will fix this issue to avoid code divergence.  @christinaholtNOAA, following up on @llpcarson's comment, it would be good to open an issue for this problem in ESMCI/manage_externals, then hopefully a modification can be made to their code to make it more efficient.

Issue #182, opened 2021-10-14
Workflow environment needs to be updated for Orion


chan-hoo:
# Description

- Regional workflow doesn't work due to a conflict of miniconda on Orion.
- The build and workflow environment files in 'ufs-srweather-app/env' need to be updated for Orion.

PR #183, opened 2021-10-14
Update build and workflow environment for Orion

chan-hoo:
## DESCRIPTION OF CHANGES: 
- Since the regional workflow does not run due to a version conflict of miniconda on Orion, the workflow environment for Orion is updated.
- Some build environment is updated for Orion.

## TESTS CONDUCTED: 
- GST test on Orion

## ISSUE: 
Fixes issue mentioned in #182

### DIFF
--- a/env/wflow_orion.env
+++ b/env/wflow_orion.env
@@ -1,9 +1,6 @@
 # Python environment for workflow on Orion
 
-module purge
-
-module load contrib/0.1
-module load rocoto/1.3.3
+module load contrib rocoto
 
 module use -a /apps/contrib/miniconda3-noaa-gsl/modulefiles
 module load miniconda3/3.8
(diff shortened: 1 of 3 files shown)


env/wflow_orion.env
```diff
@@ -1,7 +1,10 @@
 # Python environment for workflow on Orion
 
-module load rocoto
+module purge
+
+module load contrib/0.1
+module load rocoto/1.3.3
```
JeffBeck-NOAA, line 4: @chan-hoo, should contrib and rocoto have version numbers when loaded?  @christinaholtNOAA's [PR](https://github.com/ufs-community/ufs-srweather-app/pull/184) that I just merged does not.  I tend to avoid version numbers as they cause the module load commands to break more often than when just specifying the module name.

chan-hoo: @JeffBeck-NOAA , removed!

JeffBeck-NOAA: Thanks @chan-hoo!  Merging now.

JeffBeck-NOAA (approved):

Issue #202, opened 2022-01-11
RRFS-CMAQ for air quality modeling needs to be merged into the UFS SRW App.


chan-hoo:
## Description
Since RRFS-CMAQ, based on the UFS SRW App, has been established for air quality modeling successfully, it should be merged into the authoritative repository.

PR #203, opened 2022-01-11
Merge RRFS-CMAQ into UFS SRW App

chan-hoo:
## DESCRIPTION OF CHANGES: 
RRFS-CMAQ for air quality modeling is ready to be merged into the app.

## QUICK START GUIDE:
1. For air quality modeling (RRFS-CMAQ):
- Free-forecast (non-DA): run './build_UFS-SRW-App.sh RRFS-CMAQ'
- DA: run './build_UFS-SRW-App.sh RRFS-CMAQ YES' 
- https://github.com/ufs-community/ufs-srweather-app/wiki/Air-Quality-Modeling

2. For RRFS: run './build_UFS-SRW-App.sh'

## TESTS CONDUCTED: 
1. Air quality modeling:
- RRFS-CMAQ: non-DA (free forecast), warm start, 2 day 2 cycles, RRFS-CONUS 13km on hera and wcoss_dell_p3
- RRFS-CMAQ-DA: DA with GSI and JEDI, warm start, 1 day 4 cycles, RRFS-CONUS 13km on hera

2. WE2E test for RRFS on hera and wcoss_dell_p3:
- grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
- grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15_thompson_mynn_lam3km
- MET_ensemble_verification

## DEPENDENCIES:
- NOAA-EMC/regional_workflow/pull/665

## ISSUE: 
Fixes issue mentioned in #202 

## CONTRIBUTOR:
@rmontuoro


JeffBeck-NOAA (commented):
@chan-hoo, this PR just contains build scripts, environment files, and config sample scripts for the RRFS-CMAQ, correct?  Anything else we should know about this PR?  Thanks!

christinaholtNOAA:
@chan-hoo Will you let us know when you're happy with another review of this PR and its regional_workflow counterpart?

chan-hoo:
@JeffBeck-NOAA, you are correct. Since RRFS-CMAQ needs extra external components such as Nexus, GSI, and JEDI and uses the different forecast model, FV3GFS-AQM, the list of the external components, the build environment and configuration files should be replaced with the default ones for FV3-LAM. Moreover, FV3GFS-AQM, GSI, and JEDI use different modules, their build scripts are separated from other components.

chan-hoo:
@christinaholtNOAA, sure. I'll let you know when the PRs are ready for another review.

chan-hoo:
@christinaholtNOAA, this PR and its regional workflow are ready for another review.

christinaholtNOAA:
Thanks for letting me know. I will try to get to it early next week. Is that acceptable?

mkavulich:
@chan-hoo Sorry for the delay in getting to this review. 

I have been looking at the code here and in regional_workflow, and I have some concerns about the lack of integration with the rest of the SRW App. It seems as if this capability has been mostly added in parallel to the existing build process, which would make support and maintenance of this capability (especially across different platforms) nearly impossible. In addition, the requirement for gerrit/VLAB access to even check out the code for testing makes this inaccessible to the majority of the community (myself included). I think these are both problems that will need to be addressed prior to discussing the specifics of the code merger.

I know that none of these points are specifically written anywhere as requirements for contributions to the UFS SRW App, which makes this PR a good test case for starting a discussion about setting requirements and standards for contributions to the App. If you are willing, I'd like to start this discussion at tomorrow's SRW App Code Management Meeting.

chan-hoo:
@mkavulich, yes, we can discuss this tomorrow. I understand what you concern about. 1) build process: what is your recommendation? Instead of replacing the existing configuration and environment files of FV3-LAM with those of RRFS-CMAQ, do you mean that it would be better to put separate configuration and environment files with different suffixes for each model? 2) gerrit/VLab: I agree with you. CMAQ will be added to the UFS Weather Model as a new component shortly. This means the code of the new version of FV3GFS-AQM can be checked out from GitHub soon. I'll ask the code manager of FV3GFS-AQM if it can be copied to GitHub. 3) Most of the problems (configuration, environment, input files) are caused by the fact that FV3GFS-AQM couples with the old version of FV3GFS (v.15). As I mentioned in 1), this will be resolved shortly (hopefully by this summer). However, the management group of air quality modeling wants to use the latest regional workflow and put RRFS-CMAQ into UFS SRW App. This is not something I can decide on. We should put RRFS-CMAQ coupling with the trouble maker, old version of FV3GFS, into UFS SRW App. Let me know if you have any other brilliant ideas :) Let's talk about this tomorrow.

chan-hoo:
@christinaholtNOAA, @JeffBeck-NOAA , @mkavulich, can you review this PR with the PR of the regional workflow? I've modularized the components to simplify the build steps. With this change, we don't have to worry about the environment files any more. They are copied from the source code and loaded from the regional workflow regardless of the format of the module files (e.g. lua) or the version of the code. Moreover, RRFS-DA can be added to the app easily with this approach. This looks much more reasonable to me. Can we speed up the review process? The air quality modeling group has been waiting for a long time. They have the timeline for their projects. Once this is merged into the app, they will be able to move forward to their next steps.

chan-hoo:
This is my design concept: ufs_srw_app_comp

JeffBeck-NOAA:
@chan-hoo, your design concept is great!  Just a minor modification, the RRFS will have DA, with v1 being GSI based (the plan is to have v2 be JEDI based).

chan-hoo:
ufs_srw_app_comp

JeffBeck-NOAA:
> ufs_srw_app_comp

@chan-hoo, I think you can just replace RRFS-DA with RRFS, and it will be good.  Sorry for not clarifying previously!

chan-hoo:
ufs_srw_app_comp

build_RRFS-CMAQ.sh
```diff
@@ -0,0 +1,64 @@
...
+
+###########################################################
+## User specific parameters                              ##
+###########################################################
+##
+## DA option ("NO" or "YES")
+##    NO: hera, wcoss_dell_p3
+##    YES: hera
+##
+DA_opt="NO"
```
gsketefian: @chan-hoo You can make `DA_opt` an optional argument so you don't have to change the script to reset it. e.g.:
`
DA_opt=${$1:-NO}
`

chan-hoo: @gsketefian, good idea! Modified.

build_RRFS.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
mkavulich: I am not fully informed on this subject, but from my understanding the "RRFS" moniker specifically refers to operational implementations. The top-level community build in this repository should not be referred to as "RRFS"; it is the "UFS Short-Range Weather Application" (UFS SRW App).

build_RRFS.sh
```diff
@@ -0,0 +1,46 @@
...
+## User specific parameters                              ##
+###########################################################
+##
+## Compiler:
+##
+export COMPILER="intel"
+##
+## CCPP_SUITES:
+##
+export CCPP_SUITES="FV3_CPT_v0,FV3_GFS_2017_gfdlmp,FV3_GFS_2017_gfdlmp_regional,FV3_GSD_SAR,FV3_GSD_v0,FV3_GFS_v15p2,FV3_GFS_v16,FV3_RRFS_v1beta,FV3_HRRR,FV3_RRFS_v1alpha,FV3_GFS_v15_thompson_mynn_lam3km"
```
mkavulich: We should be careful which of these we include by default; I believe some of these are due to be removed from support in the top of ufs-weather-model. It could be we want the default suites to be handled there rather than in the top-level app build.

build_RRFS-CMAQ.sh
```diff
@@ -0,0 +1,71 @@
...
+## Compiler
+##
+export COMPILER="intel"
+##
+###########################################################
+echo "DA option:" ${DA_opt}
+
+# Check out the external components for RRFS
+echo "... Checking out the common external components: regional workflow, UFS_UTILS, UPP ..."
+./manage_externals/checkout_externals -e comp_conf/COMMON/Externals.cfg
```
mkavulich: Can you explain the naming convention behind this new directory structure? I do not know what "comp_conf" should stand for, for example.

comp_conf/RRFS-CMAQ-DA/build_scripts/build_GSI.sh
```diff
@@ -0,0 +1,35 @@
...
+SRW_APP_DIR="${MYDIR}/../../.."
+COMP_DIR="${SRW_APP_DIR}/comp_conf"
+SRC_DIR="${SRW_APP_DIR}/src/gsi"
+BUILD_DIR="${SRW_APP_DIR}/build/gsi"
+BIN_DIR="${SRW_APP_DIR}/bin/"
+
+# Detect MACHINE
+source ${COMP_DIR}/detect_machine.sh
+
+echo "MACHINE:" ${MACHINE}
```
mkavulich: As was mentioned in the weekly meeting, we should have a way to build GSI on all machines if it is going to be included in develop. At a *minimum* it should work on all [tier-1 platforms](https://github.com/ufs-community/ufs-srweather-app/wiki/Getting-Started#pre-staged-raw-initial-conditions)

comp_conf/RRFS/build_scripts/build_ufs-weather-model.sh
```diff
@@ -0,0 +1,79 @@
...
+module list
+
+# Copy module file to env
+cp "${SRC_DIR}/modulefiles/${MODULE_FN}" "${COMP_DIR}/env/modulefile.ufs-weather-model"
+cp "${SRC_DIR}/modulefiles/ufs_common" "${COMP_DIR}/env/ufs_common"
+
+# Set cmake environment
+source ../../cmake_env_machine.sh
+
+export ESMFMKFILE=${ESMFMKFILE:?"Please set ESMFMKFILE environment variable"}
```
mkavulich: If any variables are required to build, this should be handled at the top-level, no incomplete builds for environment problems.

comp_conf/COMMON/build_scripts/build_UFS_UTILS.sh
```diff
@@ -0,0 +1,66 @@
...
+echo "Moduel file name:" ${MODULE_FN}
+
+# Load modules
+module purge
+module use ${SRC_DIR}/modulefiles
+module load ${MODULE_FN}
+module list
+
+# Copy module file to env
+cp "${SRC_DIR}/modulefiles/${MODULE_FN}" "${COMP_DIR}/env/modulefile.UFS_UTILS${FN_SFX}"
```
mkavulich: I worry that this "copy at build-time" method of setting environment files and the like will complicate the ability to "clean" an existing build and start from scratch

devbuild.sh
```diff
@@ -1,244 +0,0 @@
-#!/bin/bash
```
mkavulich: A lot of functionality has been lost with the removal of this script; these options should be restored or justified as to why they were removed.

comp_conf/env/modulefile.build.hera.JEDI
```diff
@@ -0,0 +1,11 @@
+export JEDI_OPT=/scratch1/NCEPDEV/jcsda/jedipara/opt/modules
+module use $JEDI_OPT/modulefiles/core
+module load jedi/intel-impi/2020.2
+
+export SLURM_ACCOUNT=da-cpu
```
mkavulich: Runtime account information should not be included in the build system modulefiles; these should be handled at runtime in configuration files.

comp_conf/cmake_env_machine.sh
```diff
@@ -0,0 +1,80 @@
...
+set -eu
+
+MACHINE_upper=${MACHINE^^}
+
+echo "MACHINE:" ${MACHINE}
+echo "COMPILER:" ${COMPILER}
+
+case "${MACHINE_upper}" in
+#
+  "WCOSS_CRAY")
```
mkavulich: These options should not be hard-coded; the user should be able to set these in their environment and have them be preserved through the build.

mkavulich (commented):
@chan-hoo Again, sorry if it seems I have been ignoring your PR, 7000+ lines added across two repositories is a lot to review!

I'll also add that I am sympathetic to your group's timeline and, while I think we should not merge this PR before the March "code slush" (at least, not in its current form), I will put in the effort required to ensure that the capability can be merged prior to your May deadline for realtime runs.

Consider this an initial review of concerns I have; I presume the discussion will cause a lot of the specifics to evolve in the coming days and weeks. I will also leave comments in the regional_workflow repository relevant to that PR.

@JeffBeck-NOAA @gsketefian and others please feel free to jump in on any of my comments as well; a change this big needs wider feedback than just me.

devbuild.sh
chan-hoo: returned

build_RRFS.sh
chan-hoo: This is not the default value. It is set in "ufs-srweather-app/components/RRFS/build_scripts/build_ufs-weather-model.sh". This is an example.

build_RRFS-CMAQ.sh
chan-hoo: Renamed 'comp_conf' to "components".

build_RRFS.sh
chan-hoo: If RRFS-CMAQ is merged into the UFS SRW App, we'll need to distinguish between the non-coupled (FV3 stand-alone) regional modeling (current UFS SRW App) and the coupled regional air quality modeling (RRFS-CMAQ). Those two are on the same level. The non-coupled regional modeling will not be the only top-level any more. About the naming of RRFS, I think @JeffBeck-NOAA has an answer to this. Do you mean that we should combine the two scripts 'build_RRFS.sh' and 'build_RRFS_CMAQ.sh' into one 'build_UFS_SRW_App.sh'? I'll try to merge them.

comp_conf/RRFS-CMAQ-DA/build_scripts/build_GSI.sh
chan-hoo: This GSI is not the same as the authoritative one. It was modified only for RRFS-CMAQ. The developer will merge his change into the authoritative repo in the future. Until then, this GSI should be separated from the authoritative one. Currently, RRFS-CMAQ works only on Hera and WCOSS Dell, and RRFS-CMAQ-DA works only on Hera because JEDI is only available on Hera now.

comp_conf/COMMON/build_scripts/build_UFS_UTILS.sh
chan-hoo: Yes. Each has its own pros and cons. As I've managed UFS SRW App at EMC, I felt that the one integrated env module file for all the components was not efficient. Whenever we update the hashes of the components, we should check their env files and test them again. Each component already has been validated with its built-in module file. Another issue is that a certain version of a specific module is sometimes not available on WCOSS. In this case, we should separate the component from the integrated external comp list and env file, and compile it manually and load the module separately in the regional workflow. In my opinion, this new approach is much easier to manage the components at least on the tier-1 machines and is more efficient to install the app especially for RRFS-CMAQ than before.

JeffBeck-NOAA: We've had this discussion before, and while I see the utility of having a single env file per platform housed within the umbrella SRW App repo, my personal feeling is that it puts unnecessary work on the SRW App managers (us) to keep it working and current.  The other authoritative UFS repo managers keep these env files up to date and tested on all platforms already, why not take advantage of that and just copy them over at build/run time?  Then our modules are always in sync with whatever is being used for that hash of the repo in question.   With the variety of HPC platforms that exist, and the growing number of required modules, I'd much rather the other code managers keep things working, and then we just leverage that work for the SRW App.  If there are separate build scripts for each component, then a component-specific "clean" shouldn't be a problem, right?  I also like the idea of being able to build one component separately within the App if I choose, instead of all together, as I've wanted that functionality when I only needed to rebuild ufs-weather-model, for example.

Issue #215, opened 2022-02-18
Some directories should be renamed to meet the NCO standards


chan-hoo:
## Description
- The directories of docs, env, src, and bin should be renamed to meet the NCO standards (pp 19, Directory Structures).
- The workflow env files in the 'env' directory needs to be moved to the regional workflow.

PR #216, opened 2022-02-18
Rename dirs and make executable dir switchable to meet NCO standards

chan-hoo:
## DESCRIPTION OF CHANGES: 
- Some directories are renamed to meet the NCO standards (Ver 11.0.0, January 19, 2022, pp 19 Directory Structures):
> docs -> doc
> env -> modulefiles
> src -> sorc
> bin -> exec
- The executable directory name can be switched between `exec` and `bin`. A PR pre-required for this is under review in NOAA-EMC/UPP.
- The workflow environment files in the 'env' directory are moved to the 'regional_workflow/env' directory.

## DEPENDENCIES:
- ufs-community/regional_workflow/pull/679
- NOAA-EMC/UPP/pull/448

## ISSUE: 
Fixes issue mentioned in #215


sorc/CMakeLists.txt
```diff
@@ -1,13 +1,30 @@
 include(ExternalProject)
 
+##### CMAKE Arguments depending on exec_bin #########################
+
+if(${exec_bin} STREQUAL exec)
```
gsketefian: How about something like this to avoid repetition:
```
list(APPEND COMP_ARGS "-DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX}")
if(${exec_bin} STREQUAL exec)
  list(APPEND COMP_ARGS "-DEMC_EXEC_DIR=ON")
endif()
```
My cmake syntax might be wrong, but you get the idea.

chan-hoo: @gsketefian, It worked. updated.

gsketefian: Great, thanks!

gsketefian (approved):

CMakeLists.txt
```diff
@@ -91,6 +88,10 @@ SET(host_os "${osrel}")
 SET(abs_top_builddir "${CMAKE_CURRENT_BINARY_DIR}")
 SET(abs_top_srcdir "${CMAKE_CURRENT_SOURCE_DIR}")
 
+##### Executable directory name (exec_bin): exec or bin #############
+SET(exec_bin exec)
```
christopherwharrop-noaa: This seems hard-coded.  The PR description says it is switchable, but I don't see any option for doing that at build time.  Also, what is the point of having a configurable directory name for executables?  Everything should be adjustable using `$CMAKE_PREFIX_PATH` with its subdirs using standard software industry names.  It seems to me that having a configurable directory name for the executables directory has no advantages and increases the complexity a lot.  The complexity increases dramatically because now every other component in the App tree will need to support this.  I need to be convinced that there is a good reason to make this configurable rather than just setting it to what it needs to be.

devbuild.sh
```diff
@@ -89,12 +89,12 @@ CCPP=""
...
 DISABLE_OPTIONS=""
 BUILD_TYPE="RELEASE"
-BUILD_JOBS=4
+BUILD_JOBS=8
 CLEAN=false
 CONTINUE=false
 VERBOSE=false
 # detect PLATFORM (MACHINE)
-source ${SRC_DIR}/env/detect_machine.sh
+source ${SRC_DIR}/modulefiles/detect_machine.sh
```
christopherwharrop-noaa: The contents of a directory named `modulefiles` should probably contain module files.  This is confusing.

devbuild.sh
```diff
@@ -161,7 +161,7 @@ if [ "${VERBOSE}" = true ] ; then
 fi
 
 # set ENV_FILE for this platform/compiler combination
-ENV_FILE="${SRC_DIR}/env/build_${PLATFORM}_${COMPILER}.env"
+ENV_FILE="${SRC_DIR}/modulefiles/build_${PLATFORM}_${COMPILER}.env"
```
christopherwharrop-noaa, line 164: If you're going to name this directory `modulefiles` then all of the `*.env` files should be converted to actual module files and should be loaded with, for example, `module load build_hera_intel` rather than sourced.  This would also remove the user's shell from the equation as the `module load` would work regardless of the shell the user is using.

sorc/CMakeLists.txt
```diff
@@ -1,13 +1,24 @@
 include(ExternalProject)
 
+##### CMAKE Arguments depending on exec_bin #########################
+
+list(APPEND COMP_ARGS "-DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX}")
+if(${exec_bin} STREQUAL exec)
```
christopherwharrop-noaa: What is the point of supporting multiple names for the executable install directory?

christopherwharrop-noaa:
I understand that the point of this PR is to make the ufs-srweather-app conform to NCO naming conventions.  However, I can't in good conscience let this go by without expressing a few thoughts about that.  I'd like to remind everyone that this is a community model, and as such, its development should always be guided by standard software industry practices and principles.  Departures from well accepted ubiquitous industry standards should only be done for reasons that can be objectively evaluated and justified on their technical merits and which clearly demonstrate why the departure improves the codebase.  Departures from industry standards present customers with unexpected attributes, features, and behaviors.  This has at least two consequences: (1) Usability by the community is decreased because customers have to learn non-standard behaviors and quirks (2) Nonstandard nomenclature and naming conventions undermines the otherwise professional quality of the product; it may be harmful to the organization's reputation because it will not appear "professional".  This may sound like a minor point, but the software industry has well established conventions for how the various parts of a code base are organized.  NCO is absolutely an outlier when it comes to their naming convention standards. Literally no one else on Earth uses those conventions (e.g. `sorc/` ).  As a community model, ufs-srweather-app is public facing and how it is perceived and used by customers in the research community matters.

CMakeLists.txt
christopherwharrop-noaa: To clarify, I think this should look like:

```
set(exec_bin "exec" CACHE STRING "Executable directory name")
```

or perhaps

```
set(exec_bin "exec" CACHE INTERNAL "")
```

christopherwharrop-noaa (changes requested):
I can't approve until the following are addressed:

1. A CMake option is added for the value of `exec_bin` needs to be chosen at build time with a CMake option or it is 
2. Either the contents of `modulefiles` directory are changed to be actual module files or the directory is not named "modulefiles"

CMakeLists.txt
christopherwharrop-noaa: I would also argue that the default should be `bin` not `exec` for the simple reason that NCO is the only place where `exec` would be desired.  The default should be the value that the community expects.

chan-hoo: @christopherwharrop-noaa, I am not familiar with cmake. Can you help me for this? I'll invite you to my repo.

devbuild.sh
chan-hoo: @christopherwharrop-noaa, ok. I'll move it to another dir.

devbuild.sh
chan-hoo: @christopherwharrop-noaa, I agree. This will be my next PR. I don't want to put several topics into one PR. I'll modify the env files into the regular module files in another PR.

chan-hoo:
@christopherwharrop-noaa, I understand what your point is. However, most users of the UFS SRW App and the regional workflow, including myself, are not familiar with the standard software industry practices and principles because their backgrounds are not software engineering or computer science. Only some knowledgeable people like you perceive it. I've never heard that 'bin' is the standard name for executables before. This might be my lack of knowledge. I don't care about the name of the executable directory. However, the NCO standards describe that the executable directory should be named to 'exec'. Simply put yourself in my position. There exists one standards describing the rules specifically even though it is only for one organization. Other institutes don't have standards like this. In this case, it will be more professional to follow the existing standards. I agree that it may be harmful to the organization's reputation. However, it is not something we can decide at this moment. I suggest you to meet the NOAA management group and update the standards with the software industry principles :) :)  Can you share the standard software industry practices and principles with us if you have any documents? It will be definitely helpful at least for me to be a professional software engineer in the far future.

CMakeLists.txt
chan-hoo: @christopherwharrop-noaa, Now I got your point. I'll modify exec_bin soon.

devbuild.sh
chan-hoo: @christopherwharrop-noaa, if you want this change in this PR, I'll do it soon.

sorc/CMakeLists.txt
chan-hoo: @christopherwharrop-noaa, the UFS SRW App will be used as an operational model. We should change the app to meet the NCO standards for this. If you and other reviewers agree, we can use only one name as 'exec'. But I don't think this is what you want.

CMakeLists.txt
chan-hoo: @christopherwharrop-noaa, a new cmake argument '-DEMC_EXEC_DIR' was introduced to change the dir name to 'exec' for  the NCO standard. This is the same approach as UFS_UTILS and UPP. Its default is 'bin'. To change the name to 'exec', we just need to add "-DEMC_EXEC_DIR=ON" to CMAKE_SETTINGS as described in devbuild.sh (ln 213).

devbuild.sh
chan-hoo: @christopherwharrop-noaa, this file has been moved into the 'test' dir.

CMakeLists.txt
aerorahul: The appropriate solution would be to provide at cmake configuration time:
`-DCMAKE_INSTALL_BINDIR=exec`

aerorahul: > I would also argue that the default should be `bin` not `exec` for the simple reason that NCO is the only place where `exec` would be desired.  The default should be the value that the community expects.

That is the default that CMake assigns.

CMakeLists.txt
```diff
@@ -91,6 +88,12 @@ SET(host_os "${osrel}")
 SET(abs_top_builddir "${CMAKE_CURRENT_BINARY_DIR}")
 SET(abs_top_srcdir "${CMAKE_CURRENT_SOURCE_DIR}")
 
+# Executable directory name (exec_bin): exec or bin
+SET(exec_bin "bin")
+IF(EMC_EXEC_DIR)
+  SET(exec_bin "exec")
+ENDIF()
```
aerorahul, line 95: This should not be necessary.

CMakeLists.txt
```diff
@@ -119,7 +122,7 @@ MESSAGE(${UFS-SRWEATHER-APP_SETTINGS})
 # Install ufs_srweather_app.settings file into same location
 # as the app.
 INSTALL(FILES "${CMAKE_BINARY_DIR}/ufs_srweather_app.settings"
-  DESTINATION bin)
+  DESTINATION ${exec_bin})
```
aerorahul, line 125: Use ${CMAKE_INSTALL_BINDIR}

devbuild.sh
```diff
@@ -210,6 +210,7 @@ fi
 # cmake settings
 CMAKE_SETTINGS="-DCMAKE_INSTALL_PREFIX=${INSTALL_DIR}"
 CMAKE_SETTINGS="${CMAKE_SETTINGS} -DCMAKE_BUILD_TYPE=${BUILD_TYPE}"
+#CMAKE_SETTINGS="${CMAKE_SETTINGS} -DEMC_EXEC_DIR=ON"
```
aerorahul, line 213: Instead use
-DCMAKE_INSTALL_BINDIR=exec 
where exec is desired.

aerorahul (commented):
A few comments on the discussion of bin vs exec. 
BTW, NCO does not object to src or bin vs sorc or exec. 
I am not sure about the envs case. 
Hope this helps.

christopherwharrop-noaa:
If @aerorahul is correct that NCO doesn't object to the use of `src` and `bin` then I would like to simply close this pull request (or limit it to doc and modulefiles).  How would we get positive confirmation of that?

chan-hoo:
@christopherwharrop-noaa, @aerorahul is the right person who can confirm this. Since he said ok, I'll close this pr and submit a new one for modulefiles soon.

PR #232, opened 2022-03-30
Add python module to cheyenne build environments

willmayfield:
## DESCRIPTION OF CHANGES: 
Build fails on Cheyenne if no python environment is loaded, so loading a python module to fix it.

## TESTS CONDUCTED: 
- grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2 passed on intel
- build now succeeds on gnu

### DIFF
--- a/env/build_cheyenne_gnu.env
+++ b/env/build_cheyenne_gnu.env
@@ -7,6 +7,7 @@ module load ncarenv/1.3
 module load gnu/10.1.0
 module load mpt/2.22
 module load ncarcompilers/0.5.0
+module load python/3.7.9
 module unload netcdf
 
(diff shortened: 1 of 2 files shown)


gsketefian:
@willmayfield Just curious, is the need to load the python module due to a recent change in the build system (since it was apparently not needed before)?

mkavulich (approved):
Looks good. I also tested these and was able to build successfully with both intel and gnu compilers with these changes.

gsketefian (approved):

JeffBeck-NOAA (approved):

christopherwharrop-noaa (approved):

PR #238, opened 2022-04-12
Replace bash env files with modules 

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 

- Replaces environment files with modules. This should make build and workflow execution SRW app from both `bash` and `tcsh` login shells possible.

- Removes `detect_machine.sh` script and forces the user to pass `--platform=PLATFORM` instead. Auto-detection of machine is error prone and an impediment to new systems, e.g. Gaea

## TESTS CONDUCTED: 
Tested build on **Hera**,  **Jet**, **Orion**, **Gaea**, **WCOSS** and **Singularity** :

    ./devbuild.sh --platform=hera

---

Cheyenne, Macosx and Odin need testing

---

To build directly with cmake, as is the current recommendation, some systems such as Odin/Gaea/Macosx need
to properly initialize `Lmod` first. This is taken care of automatically if you build with `devbuild.sh` but has to be done
manually otherwise. Assuming you are using `bash` login shell:

    source etc/lmod-setup.sh gaea
    module use modulefiles
    module load build_gaea_intel
    ...

## DEPENDENCIES:
None

## DOCUMENTATION:
Updated `docs/INSTALL` and `docs/RUNTIME` instructions, and `BuidRunSRW.rst` as well

## ISSUE (optional): 
None

## CONTRIBUTORS (optional): 
None


chan-hoo:
@danielabdi-noaa, can you explain why you try to remove 'detect_machine.sh' and go back to the platform argument option? I personally like this approach because I don't need to think of which machine I am working on and I just need to type 'devbuild.sh' on the terminal. This approach is also used in the UFS Weather Model. Do you have any specific reasons to remove it?

RatkoVasic-NOAA:
> @danielabdi-noaa, can you explain why you try to remove 'detect_machine.sh' and go back to the platform argument option? I personally like this approach because I don't need to think of which machine I am working on and I just need to type 'devbuild.sh' on the terminal. This approach is also used in the UFS Weather Model. Do you have any specific reasons to remove it?

We definitely need detect_machine.sh both in UFS Weather Model App and SRW app.

christopherwharrop-noaa:
@chan-hoo / @RatkoVasic-NOAA  - I disagree.  Detecting the machine is inherently unportable and unmaintainable and provides no added value.  Not to mention that it publishes FQDN of our machines in a public repo.  There is absolutely no reason for it because every user always knows what machine they are using and can easily provide that information.  Having a lot of fragile code to provide a service that save a human mere seconds, is not worth it.  Furthermore, it makes working on additional platforms less intuitive and gives the incorrect impression to users that the system only works in certain places.  Replacing automated detection of machine with a human input makes more sense and removes portability issues.  Automated detection of the machine is not a scalable, maintainable,  portable, or secure, feature and should be removed.

chan-hoo:
@christopherwharrop-noaa, I disagree with you. Saving even a few seconds is worth it. I assume that most users of the UFS SRW App/Regional Workflow are working on the tier-1 and tier-2 machines. Using the script can save their precious time. You can find some specific machines in 'ufs-srweather-app/env/' and 'regional_workflow/ush/machine/'. If an anonymous user wants to use the UFS SRW App on a machine that is not listed there, the user should add scripts to these directories. If the user can do this, it will not be hard to add its machine to the list in 'detect_machine.sh'. Why do you think it is not scalable, maintainable, portable, or secure? I do not think that just using one auto-detection script is such a big deal. From this point of view, the UFS SRW App / Regional workflow are not scalable, maintainable, portable, or secure.

christopherwharrop-noaa:
"saving even a few seconds is worth it. "  - No it isn't.  Sacrificing portability and maintainability of the system to avoid typing "hera" or "wcoss" is a very bad trade.  The benefit is almost non-existent, while the disadvantages are plentiful.

"I assume that most users of the UFS SRW App/Regional Workflow are working on the tier-1 and tier-2 machines" - This assumption is directly responsible for some very poor design choices which many of us are now trying to reverse.  It doesn't matter where "most" users run because it still needs to be functional and portable, even to non-Tier-1 machines.  A user should on a non-tier-1 platform should NEVER have to edit code or the build system to make the system work, they should only have to provide a env file that configures the environment.

The auto-detection feature scales linearly with O(N) where N is the number of machines.  Unfortunately the number of machines is unlimited, so keeping that up-to-date means adding more and more machines.  The ufs-community should not be responsible for maintaining an unbounded growing list of stanza's in that script.

It is unmaintainable because the detection feature's correctness depends on outside entities beyond the control of ufs-community.  Names of machines change, and the details may change over time without anyone noticing.  Every time a machine changes or is added, the code (not just configuration) in the repository has to change.  It is fragile code that depends on magic, hard-coded, constants.

It is fundamentally unportable because the auto-detection feature only works on machines for which someone has added code.  The instant any part of the UFS system relies on detection of specific machines, it becomes unportable.  Portability means it works everywhere without changing the code.  Use of auto-detection of machine fundamentally breaks that.  Portability is also a critical requirement for community-based software.

It is insecure because it embeds FQDN into code published in a public repository.  While those details are probably easily accessible to would-be attackers via other means, it seems irresponsible to advertise an entire list of hostnames for all of our Tier-1 platforms.

chan-hoo:
@christopherwharrop-noaa, I still disagree with you :)  However, if other reviewers agree with it, I'll respect their opinions.

env/set_machine.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+MACHINE="$1"
+PLATFORM="${MACHINE}"
+
+printf "PLATFORM(MACHINE)=${PLATFORM}\n" >&2
```
christinaholtNOAA: It's not completely obvious if we need this any more. Maybe I'm missing the point, though?

env/wflow_odin
```diff
@@ -0,0 +1,47 @@
...
+__conda_setup="$('/scratch/software/Odin/python/anaconda2/bin/conda' 'shell.bash' 'hook' 2> /dev/null)"
+if [ $? -eq 0 ]; then
+    eval "$__conda_setup"
+else
+    if [ -f "/scratch/software/Odin/python/anaconda2/etc/profile.d/conda.sh" ]; then
+        . "/scratch/software/Odin/python/anaconda2/etc/profile.d/conda.sh"
+    else
+        export PATH="/scratch/software/Odin/python/anaconda2/bin:$PATH"
+    fi
+fi
```
christinaholtNOAA: I am not sure this section would work. It looks fairly bash-like.

It may be useful for someone to install a module to handle this Python installation. There seems to be no way to "unload" it and that can lead to issues on some platforms.

env/wflow_jet
```diff
@@ -0,0 +1,21 @@
...
+module use -a /contrib/miniconda3/modulefiles
+module load miniconda3
+
+if { [module-info mode load] } {
+  puts "conda activate regional_workflow;"
+}
+
+if {[module-info mode remove]} {
+  puts "conda deactivate;"
+}
```
christinaholtNOAA: Ugh. I forgot about this problem. The modules on jet and hera won't allow you to activate the environment from the module. It's quite annoying. 

@danielabdi-noaa did you find that it works when you load it? I am happy to test, too.

christinaholtNOAA (changes requested):
@danielabdi-noaa Thanks for doing this. It is super helpful for those of us running and testing on the cloud. We definitely have reached a state of running on a whole variety of platforms of our own creation that are not the Tier-1 platforms, and this is an important step toward reducing machine dependency for all that work.

It is also important to capture the idea that `/bash` is not the default for every user, and this transition to module files is a huge help in streamlining the build/deploy process. 

I did make a couple of comments below about some issues that might need some follow up tests. Someone who has access to Odin might be able to help test that, and I can help with the Jet/Hera tests. Those wflow_ module files aren't a part of the workflow from what I can tell, so they just need to be tested manually to make sure you get the python environment you need.

christinaholtNOAA:
@chan-hoo We are fundamentally in a new age of spreading to all sorts of non-Tier-1/Tier-2 machines, and it is unbelievably time consuming to be able to handle these sorts of machine dependencies each time we spin up new cloud resources. The cloud resources, for the most part, are not meant to be "supported platforms" and the system will be much more portable if we can have more flexible ways for the user to specify the platform from the command line. 

I believe that being able to specify a machine on the command line is a tiny compromise for enabling smooth portability to platforms that could free up some resources on research platforms.

env/set_machine.sh
danielabdi-noaa: I guess it is only used to set the `MACHINE` and `PLATFORM` env variables now so I might as well move it to the script that calls set_machine.

env/wflow_jet
danielabdi-noaa: It seems to work on hera and just now tested on jet, but with a caveat...
For some reason, loading the `wflow_hera/jet` module needs to be called twice for `conda activate` to work.
The first call to `module load wflow_hera/jet` in different errors on the machines. Hera complains `conda init $SHELL` need to be called first, while jet couldn't even find `conda` command at all as shown below.
Note that after a `module purge` and executing the commands again everything goes smoothly, but I guess I need to find a solution so that it works first time.

Here is how it went on Jet for example:

```
Daniel.Abdi@fe2 ufs-srweather-app $ module use env
Daniel.Abdi@fe2 ufs-srweather-app $ module load build_jet_intel
Daniel.Abdi@fe2 ufs-srweather-app $ module list

Currently Loaded Modules:
  1) slurm/18.08.7p1    6) hpc-intel/18.0.5.274  11) png/1.6.35   16) g2tmpl/1.10.0  21) gftl-shared/v1.3.3      26) nemsiogfs/2.5.3  31) nemsio/2.5.2  36) srw_common
  2) sutils/default     7) impi/2018.4.274       12) pio/2.5.2    17) ip/3.3.3       22) yafyaml/v0.5.1          27) sfcio/1.4.1      32) w3emc/2.7.3   37) build_jet_intel
  3) cmake/3.20.1       8) hpc-impi/2018.4.274   13) esmf/8_2_0   18) sp/2.3.3       23) mapl/2.11.0-esmf-8_2_0  28) sigio/2.3.2      33) hdf5/1.10.6
  4) hpc/1.1.0          9) jasper/2.0.25         14) fms/2021.03  19) crtm/2.4.0     24) gfsio/1.4.1             29) bacio/2.4.1      34) netcdf/4.7.4
  5) intel/18.0.5.274  10) zlib/1.2.11           15) g2/3.4.3     20) upp/10.0.10    25) landsfcutil/2.4.1       30) w3nco/2.4.1      35) wgrib2/2.0.8

Daniel.Abdi@fe2 ufs-srweather-app $ module load wflow_jet
conda: Command not found.
Daniel.Abdi@fe2 ufs-srweather-app $ module list

Currently Loaded Modules:
  1) slurm/18.08.7p1    6) hpc-intel/18.0.5.274  11) png/1.6.35   16) g2tmpl/1.10.0  21) gftl-shared/v1.3.3      26) nemsiogfs/2.5.3  31) nemsio/2.5.2  36) srw_common
  2) sutils/default     7) impi/2018.4.274       12) pio/2.5.2    17) ip/3.3.3       22) yafyaml/v0.5.1          27) sfcio/1.4.1      32) w3emc/2.7.3   37) build_jet_intel
  3) cmake/3.20.1       8) hpc-impi/2018.4.274   13) esmf/8_2_0   18) sp/2.3.3       23) mapl/2.11.0-esmf-8_2_0  28) sigio/2.3.2      33) hdf5/1.10.6   38) rocoto/1.3.3
  4) hpc/1.1.0          9) jasper/2.0.25         14) fms/2021.03  19) crtm/2.4.0     24) gfsio/1.4.1             29) bacio/2.4.1      34) netcdf/4.7.4  39) miniconda3/4.5.12
  5) intel/18.0.5.274  10) zlib/1.2.11           15) g2/3.4.3     20) upp/10.0.10    25) landsfcutil/2.4.1       30) w3nco/2.4.1      35) wgrib2/2.0.8  40) wflow_jet


Daniel.Abdi@fe2 ufs-srweather-app $ module load wflow_jet
usage: conda [-h] [-V] command ...
....
(regional_workflow) Daniel.Abdi@fe2 ufs-srweather-app $ 
```

env/wflow_odin
danielabdi-noaa: Yeah, wflow_odin seem to be the most challenging. TCL language allows to call shell scripts or commands via a puts to stdout so was just hoping it will work. Recent modules versions have `sh-to-mod`, but couldn't find it on any system to test ...

chan-hoo:
@christinaholtNOAA, Thank you for your explanation. This is pretty much convincing to me. Sorry, Chris :) :) I'll agree with this change from now on. I think that the directory 'env' should be renamed to 'modulefiles' because all other external components have their module files in the directory 'modulefiles'. What do you think of it?

christopherwharrop-noaa:
@chan-hoo - Renaming `env` to `modulefiles` seems reasonable to me.

danielabdi-noaa:
@chan-hoo @christopherwharrop-noaa Thanks for the suggestion! I have renamed `env` to `modulefiles` now since it only contains modulefiles now.

devbuild.sh
```diff
@@ -228,10 +237,10 @@ if [ "${VERBOSE}" = true ]; then
   MAKE_SETTINGS="${MAKE_SETTINGS} VERBOSE=1"
 fi
 
-# source the environment file for this platform/compiler combination, then build the code
-printf "... Source ENV_FILE and create BUILD directory ...\n"
-module use ${SRC_DIR}/env
-. ${ENV_FILE}
+# source the module file for this platform/compiler combination, then build the code
+printf "... Load MODULE_FILE and create BUILD directory ...\n"
+module use ${SRC_DIR}/modulefiles
```
chan-hoo, line 250: @danielabdi-noaa, the build process doesn't go further without "module purge" before module load on wcoss_dell_p3. If this is an issue only on wcoss_dell_p3, can you add if statement for wcoss_dell_p3 and put "module purge" there?

chan-hoo (changes requested):

devbuild.sh
chan-hoo: @danielabdi-noaa, I confirmed that the build process was completed successfully on wcoss_dell_p3, hera, and orion.

chan-hoo (approved):

devbuild.sh
```diff
@@ -228,10 +237,11 @@ if [ "${VERBOSE}" = true ]; then
...
+module purge
```
christinaholtNOAA: I think that a purge here was causing some issues on Cheyenne in the past. @mkavulich Can you confirm that is still problematic.

modulefiles/build_macos_gnu
```diff
@@ -1,32 +1,42 @@
...
 # Option 2 (uncomment if needed) Intel chip, x86_64
 # 
 # Option 1 (default):
-export BASH_ENV=/opt/homebrew/opt/lmod/init/bash
+setenv BASH_ENV "/opt/homebrew/opt/lmod/init/bash"
+
 # Option 2 (uncomment if needed):
-#export BASH_ENV=/usr/local/opt/lmod/init/bash  
-source $BASH_ENV
+#setenv BASH_ENV "/usr/local/opt/lmod/init/bash"
```
christinaholtNOAA: Can the commented line be removed?

modulefiles/build_odin_intel
```diff
@@ -0,0 +1,67 @@
...
+module load hpc-cray-mpich
+
+#module load srw_common
+
+module load jasper
+module load zlib
+module load png
+
+#module load cray-hdf5
+#module load cray-netcdf
```
christinaholtNOAA, line 24: I assume @ywangwof will be able to test whether these commented lines should/should not be needed here.

modulefiles/wflow_macos
```diff
@@ -0,0 +1,26 @@
...
+if { [module-info mode load] } {
+   system "source $env(VENV)/bin/activate;"
+}
+
+# Uncomment if Rocoto workflow manager is used
+# This path should point to your Rocoto module location
+# set ROCOTOmod "/Users/username/modules"
+# module use ${ROCOTOmod}
+# module load rocoto
+# module list
```
christinaholtNOAA, line 25: @mkavulich @mark-a-potts Are module files ever supported/used on macos?

christinaholtNOAA (commented):
@danielabdi-noaa Thanks for the updates. I have a few questions down below for others. The code looks good, and I'm happy to approve depending on their answers.

ywangwof (changes requested):
For platform Odin, I would like to keep the bash env settings, but not the module files you suggested. This is because the following reasons.

1. Odin is a Tier-2 support platform, all users are from NSSL locally.
2. I cannot find a method to load those Lmod module files (**hpc-stack** installation) through the TCL module file. Odin is using the "**modulecmd**" by Cray, but not the Lmod. Using the bash env, I can load those **hpc-stack** module files explicitly. I cannot find a good way to load those **hpc-stack** module files through the TCL/C module file (as in your modification). 

Anyway, if there is not a resolution to load lua module files within a TCL/C module file using the Cray **modulecmd**, I would prefer to restoring the original bash env files on Odin for a temporarily working around.

```
build_jet_intel.env
wflow_odin.env
```
I am also open to any other solution if there is one. Thanks.

danielabdi-noaa:
@ywangwof Thanks for checking the PR on Odin and making suggestion! I agree Tier-2 platforms such as odin and maybe even macosx are difficult to deal with pure TCL modulefiles. I suppose you have not used `./devbuild.sh` directly to build on odin before, because there are `module use/load` commands there before you apply your workaround. How about we keep the `env` folder to keep shell scripts for odin, macosx etc that are hard to deal with pure TCL/Lua modulefiles?
@christinaholtNOAA @chan-hoo What do you think about this?

danielabdi-noaa:
@ywangwof I have restored the env files for odin under the `env/` directory and deleted corresponding modulefiles.
I am assuming that builds on odin were never done with `devbuild.sh` so I am not going to modify that to accomodate odin.
If you come up with a modulefile solution for odin in the future, we can replace them.

For macosx, I am going to try and install hpc-stack etc to see if I can get it working with modufiles.

ywangwof (approved):
@danielabdi-noaa Thanks. Yes, I did not use "_devbuild.sh_" before. It can be easily fixed for Odin, but it will make _devbuild.sh_ unreasonably long. Anyway, the **cmake** generalized building process works well and is simple enough on Odin.

danielabdi-noaa:
> @danielabdi-noaa Thanks. Yes, I did not use "_devbuild.sh_" before. It can be easily fixed for Odin, but it will make _devbuild.sh_ unreasonably long. Anyway, the **cmake** generalized building process works well and is simple enough on Odin.

@ywangwof I have gone back to modulefiles for Odin, because I think atleast the build should work with modulefiles. Gaea is similar to Odin in this regard and the approach I used to get it working there should work for odin too. The first part of `build_odin_intel.env` where you switch the cray modules app to Lmod is now in a separate script `lmod-setup.sh` and is sourced in `devbuild.sh`. If one is using cmake to build SRW directly, then this script has to be sourced manually to setup Lmod. 

While I expect `build_odin_intel` to work now, I am not sure about `wflow_odin` where you have some bash script for conda init. However, I think getting that working somehow is better than having two folders ( `env` and `modulefiles` ) together.

Could you please un-approve the PR and help me get it working on odin with modulefiles instead?

christinaholtNOAA (approved):
@danielabdi-noaa Thanks for digging in to make sure all the supported machines are working appropriately. What you have here looks good when you finish with those last few items.

natalie-perlin:
@danielabdi-noaa - a couple of issues found when using the TCL modulefiles `build_macos_gnu `and `wflow_macos` for the MacOS would you mind sharing some thoughts whether they could be fixed, were there any similar issues on other platforms?

- `build_macos_gnu`: setting LDFLAGS environment variable is not working. This seems to be a known issue (discussed elsewhere) with the TCL modules (#%Modules), because the LDFLAGS uses other env. variable $MPI_ROOT that is only available after loading another module, and not being a system environment variable. As far as I understand, TCL file does not actually load the module, it only contains the instructions. Thus the environment variables of modules being loaded (i.e., in "module load xxxx") are not available within the TCL file when this file is interpreted by Lua. However, the variable LDFLAGS could be successfully set _after_  the module load of `build_macos_gnu` is completed. My workaround suggestion for the moment is to comment out the line `setenv LDFLAGS "-L$env(MPI_ROOT)/lib" `, and suggest a user to run this line in the terminal! Maybe more elegant solution could be later, but this works for now.

- `wflow_macos`, which supposed to activate python environment "regional_workflow" created earlier, does not seem to work for me. It does OK if I repeat the instructions manually, or when _sourcing_ a similar `wflow_macos.sh` script in sh/bash - compatible form (instead of _loading_ the module/script).  
Are there any similar issues with other platforms, or why this could be the case with MacOS?. - any comments are welcome! I could include this `wflow_macos.sh` script along with the original TCL `wflow_macos` in the repository, as a temporary workaround or at least for the PR to the SRW repo with the changes for the MacOS, until a better solution is found.
 Attached is a file `wflow_macos.sh`  except renamed to `wflow_macos.sh.txt` to conform to the attachable formats allowed by GitHub
[wflow_macos.sh.txt](https://github.com/ufs-community/ufs-srweather-app/files/8721147/wflow_macos.sh.txt)

danielabdi-noaa:
@natalie-perlin Thanks for testing it out. I was not able to test modulefiles on macosx because I lack admin privilages on my NOAA mac laptop.

There are similar issues on other systems, for example, one can't load miniconda3 and execute `conda activate` in the same modulefile because the environment changes made by `module load miniconda3` will only be available after the module is loaded. But for macos I think the problem is solvable atleast for `wflow_macos`. It looks like the problem is that the environment variable set with `setenv` is not immediately available within the same modulefile, so maybe using a local variable will work instead using something like:

```sh
set VENV_t "/Users/username/venv/regional_workflow"
setenv VENV $VENV_t
if { [module-info mode load] } { 
  system "source $VENV_t/bin/activate;"
}
```
Let me know if that works.

For `build_macos_gnu`, I think the problem is similar i.e the environment changes that loading the mpi module makes such as `MPI_ROOT` are not yet available in the modulefile. If you know the path where MPI is installed, you can set `LDFLAGS` directly with it but that is not an ideal solution. I will look for other solutions for this one ...

natalie-perlin:
@danielabdi-noaa  - Hi Daniel, 
Yes, thank you, I did test using a local variable, it did not help. Only when/if this command "source $VENV/bin/activate" is launched outside the script it does start the python environment.

So maybe this file wflow_macos could be left unused until it is more clear on how to activate the python env. from this module. At the present moment,  MacOS users could then be instructed just to load the python env. manually from the command line.

danielabdi-noaa:
@natalie-perlin I just installed lmod with a local installation of homebrew and figured out the solution for `wflow_macos`.
Using `puts` instead of `system` seem to solve the problem and venv is loaded first time.
On linux systems though, the opposite is true which is why I have been using `system` instead of `puts` there.
Using TCL modulefile with Lmod causes problems depending on version numbers

Change
```Bash
if { [module-info mode load] } { 
   puts "source $env(VENV)/bin/activate;"
}
```

```Bash
abdi-lt:~ daniel.abdi$ source homebrew/opt/lmod/init/bash 
abdi-lt:~ daniel.abdi$ module --version

Modules based on Lua: Version 8.7.2  2022-05-04 13:42 -05:00
    by Robert McLay mclay@tacc.utexas.edu

abdi-lt:~ daniel.abdi$ ls modulefiles/
wflow_macos
abdi-lt:~ daniel.abdi$ module use modulefiles
abdi-lt:~ daniel.abdi$ module load wflow_macos
(venv) abdi-lt:~ daniel.abdi$ deactivate
abdi-lt:~ daniel.abdi$ 
```

danielabdi-noaa:
The same goes for the system command that sets stack size to unlimited in `build_macos_gnu`.
```Bash
puts "ulimit -S -s unlimited;"
```

natalie-perlin:
@danielabdi-noaa - Thank you for your comments, all worked out!
 setting env. variable LDFLAGS in terminal/cli, immediately after loading "build_macos_gnu" did solve the issue to compile the code.
 
 $ module load build_macos_gnu
 $ export LDFLAGS="-L${MPI_ROOT}/lib"

With the workflow environment activation: using "puts" did solved the issue.

So all is good with the modulefiles! :)

PR #254, opened 2022-05-11
tweaks for build/run on gaea

mark-a-potts:
This fixes problems with building and running on the regional workflow on Gaea.

## TESTS CONDUCTED: 
Will test out using Jenkins on multiple platforms.

### DIFF
--- a/etc/lmod-setup.sh
+++ b/etc/lmod-setup.sh
@@ -20,7 +20,7 @@ elif [ "$L_MACHINE" = singularity ]; then
    module purge
 
 elif [ "$L_MACHINE" = gaea ]; then
-   export BASH_ENV="/lustre/f2/pdata/esrl/gsd/contrib/lua-5.1.4.9/init/init_lmod.sh"
+   export BASH_ENV="/lustre/f2/pdata/esrl/gsd/contrib/lua-5.1.4.9/init/bash"
    source $BASH_ENV
 
 elif [ "$L_MACHINE" = odin ]; then
--- a/etc/lmod-setup.sh
+++ b/etc/lmod-setup.sh
@@ -20,7 +20,7 @@ elif [ "$L_MACHINE" = singularity ]; then
    module purge
 
 elif [ "$L_MACHINE" = gaea ]; then
-   export BASH_ENV="/lustre/f2/pdata/esrl/gsd/contrib/lua-5.1.4.9/init/bash"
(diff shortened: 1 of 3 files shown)


etc/lmod-setup.sh
```diff
@@ -20,7 +20,7 @@ elif [ "$L_MACHINE" = singularity ]; then
    module purge
 
 elif [ "$L_MACHINE" = gaea ]; then
-   export BASH_ENV="/lustre/f2/pdata/esrl/gsd/contrib/lua-5.1.4.9/init/init_lmod.sh"
+   export BASH_ENV="/lustre/f2/pdata/esrl/gsd/contrib/lua-5.1.4.9/lmod/lmod/init/bash"
```
danielabdi-noaa, line 23: Could you do similar modification in `etc/lmod-setup.csh` for those who use tcsh login?

modulefiles/wflow_gaea
```diff
@@ -12,4 +12,7 @@ module load rocoto
 module load miniconda3
 
 setenv CONDA_DEFAULT_ENV "regional_workflow"
-
+setenv PROJ_LIB /lustre/f2/pdata/esrl/gsd/contrib/miniconda3/4.8.3/envs/regional_workflow/share/proj
+if { [module-info mode load] } {
+   system "conda activate regional_workflow;"
```
danielabdi-noaa: I have found that activating the regional_workflow environment in the same modulefile that miniconda3 is loaded doesn't work the first time you do `module load wflow_gaea` (conda init doesn't work). It works the second time you do it though. Thus, I chose to leave it  to the user to activate the environment manually with something like:
```
  puts stderr "Please do the following to activate conda:
       > conda activate regional_workflow"
```
The only system that direct activation of conda environment worked with a direct conda activate call is singularity, and I think that is because the miniconda3 modulefile there is written in Lua instead of TCL, which Lmod likes better. In most other systems it didn't work.

etc/lmod-setup.sh
danielabdi-noaa: Also, now that the modulefiles that the cray module app loads seem to be unloaded instead of re-loaded with Lmod, which the previous `init_lmod.sh` did, maybe it is  appropriate to add `module purge` to it just like macos and singularity do?

danielabdi-noaa (approved):

EdwardSnyder-NOAA (approved):

PR #258, opened 2022-05-16
Feature/cheyenne fix

mark-a-potts:
Cheyenne's version of LMOD does not support the "load-any" option, so the recent changes to srw_common have broken the build. A request has been made to CISL to update LMOD, but until that is completed, these changes will load the missing modules for both intel and gnu builds.


.github/workflows/build.yml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+      - run: echo "🎉 The job was automatically triggered by a ${{ github.event_name }} event."
+      - run: echo "🐧 This job is now running on a ${{ runner.os }} server hosted by GitHub!"
+      - run: echo "🔎 The name of your branch is ${{ github.ref }} and your repository is ${{ github.repository }}."
+      - name: Check out repository code
+        uses: actions/checkout@v2
+      - run: |
+             cd ufs-srweather-app
+             ./manage_externals/checkout_externals
+             source env/build_cheyenne_intel.env
```
willmayfield:
```suggestion
             module use modulefiles
             module load build_cheyenne_intel
```

willmayfield (approved):

mkavulich:
@mark-a-potts I am having trouble testing this fix for GNU on Cheyenne because there appears to be a bug in the crontab generation of the WE2E tests that overwrites the crontab rather than appending to it. Is this a known problem in the NOAA-EPIC repository? I did not encounter it when testing from develop in ufs-community a couple weeks ago, but there have been some commits since then, so I'm not sure when this problem started.

mark-a-potts:
@mkavulich, unfortunately, that sounds like a new bug. I am not sure what might be causing that. Will have to investigate...

JeffBeck-NOAA (approved):

PR #266, opened 2022-05-20
Update srw_common and hashes of UFS_UTILS and UPP 

chan-hoo:
## DESCRIPTION OF CHANGES: 
- The hashes of UFS_UTILS and UPP are updated with the latest ones.
- the module file `srw_comm` is updated with the latest modules.
- Since the latest hash of the ufs weather model requires changes in the input namelists, it will be updated in a separate PR later. However, I confirmed that the latest hash of the ufs weather model was built successfully with `srw_common` in this PR on Orion and Hera.

## TESTS CONDUCTED: 
- successfully built on Orion and Hera.
- WE2E tests on Orion and Hera:
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_2017_gfdlmp_regional
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_RAP_suite_HRRR
grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15_thompson_mynn_lam3km
grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_RRFS_v1beta
grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1alpha

## DEPENDENCIES:
- ufs-community/regional_workflow/pull/775

Issue #249, opened 2022-05-04
Port to WCOSS2


chan-hoo:
## Description
With WCOSS1 going away at the end of June, we will need to be able to run the RRFS on WCOSS2.

PR #270, opened 2022-05-24
Port SRW App to WCOSS2

chan-hoo:
## DESCRIPTION OF CHANGES: 
- Port UFS SRW App to WCOSS2
- Since HPC-stack is not available on WCOSS2 yet, this is a temporary module file.
- work with the latest hashes of UFS_UTILS and UPP (shown in PR #266)

## TESTS CONDUCTED: 
- Tested hashes:
UFS_UTILS: b6efa86
ufs weather model: 96dffa1
UPP: fbd41a5

- Build steps:
1. git clone -b feature/wcoss2 https://github.com/chan-hoo/ufs-srweather-app
2. cd ufs-srweather-app
3. ./manage_externals/checkout_externals
4. ./devbuild.sh --platform=wcoss2

## ISSUE: 
Fixes issue mentioned in #249

## CONTRIBUTORS: 
@BenjaminBlake-NOAA

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -12,7 +12,7 @@ protocol = git
 repo_url = https://github.com/ufs-community/UFS_UTILS
 # Specify either a branch name or a hash but not both.
 #branch = develop
-hash = 31271f7
+hash = b6efa86
 local_path = src/UFS_UTILS
 required = True
 
@@ -30,7 +30,7 @@ protocol = git
 repo_url = https://github.com/NOAA-EMC/UPP
 # Specify either a branch name or a hash but not both.
 #branch = develop
-hash = 394917e
+hash = fbd41a5
 local_path = src/UPP
 required = True
 
--- a/devbuild.sh
+++ b/devbuild.sh
@@ -10,7 +10,7 @@ OPTIONS
       show this help guide
   -p, --platform=PLATFORM
       name of machine you are building on
-      (e.g. cheyenne | hera | jet | orion | wcoss_dell_p3)
+      (e.g. cheyenne | hera | jet | orion | wcoss_dell_p3 | wcoss2)
   -c, --compiler=COMPILER
       compiler to use; default depends on platform
       (e.g. intel | gnu | cray | gccgfortran)
@@ -155,6 +155,7 @@ if [ -z "${COMPILER}" ] ; then
     jet|hera|gaea) COMPILER=intel ;;
     orion) COMPILER=intel ;;
     wcoss_dell_p3) COMPILER=intel ;;
+    wcoss2) COMPILER=intel ;;
     cheyenne) COMPILER=intel ;;
     macos,singularity) COMPILER=gnu ;;
     odin) COMPILER=intel ;;
--- a/devbuild.sh
+++ b/devbuild.sh
@@ -136,6 +136,12 @@ while :; do
   shift
 done
 
+# Ensure platform name from variance ======================================
+if [[ "${PLATFORM}" == "wcoss2" || "${PLATFORM}" == "cactus" ||
+      "${PLATFORM}" == "dogwood" ]]; then
+  PLATFORM="wcoss2"
+fi
+
 # check if PLATFORM is set
 if [ -z $PLATFORM ] ; then
   printf "\nERROR: Please set PLATFORM.\n\n"
--- a/devbuild.sh
+++ b/devbuild.sh
@@ -136,12 +136,6 @@ while :; do
   shift
 done
 
-# Ensure platform name from variance ======================================
-if [[ "${PLATFORM}" == "wcoss2" || "${PLATFORM}" == "cactus" ||
-      "${PLATFORM}" == "dogwood" ]]; then
-  PLATFORM="wcoss2"
-fi
-
 # check if PLATFORM is set
 if [ -z $PLATFORM ] ; then
   printf "\nERROR: Please set PLATFORM.\n\n"
--- a/etc/lmod-setup.csh
+++ b/etc/lmod-setup.csh
@@ -35,8 +35,7 @@ else if ( "$L_MACHINE" == odin ) then
    setenv MODULEPATH "/oldscratch/ywang/external/hpc-stack/modulefiles/mpi/intel/2020/cray-mpich/7.7.16:/oldscratch/ywang/external/hpc-stack/modulefiles/compiler/intel/2020:/oldscratch/ywang/external/hpc-stack/modulefiles/core:/oldscratch/ywang/external/hpc-stack/modulefiles/stack:/opt/cray/pe/perftools/21.02.0/modulefiles:/opt/cray/ari/modulefiles:/opt/cray/pe/craype-targets/default/modulefiles:/opt/cray/pe/modulefiles:/opt/cray/modulefiles:/opt/modulefiles"
 
 else if ( "$L_MACHINE" == wcoss2 ) then
-   module purge
-   module load envvar/1.0
+   module reset
 
 else
    module purge
--- a/etc/lmod-setup.csh
+++ b/etc/lmod-setup.csh
@@ -34,6 +34,10 @@ else if ( "$L_MACHINE" == odin ) then
    module --initial_load --no_redirect restore
    setenv MODULEPATH "/oldscratch/ywang/external/hpc-stack/modulefiles/mpi/intel/2020/cray-mpich/7.7.16:/oldscratch/ywang/external/hpc-stack/modulefiles/compiler/intel/2020:/oldscratch/ywang/external/hpc-stack/modulefiles/core:/oldscratch/ywang/external/hpc-stack/modulefiles/stack:/opt/cray/pe/perftools/21.02.0/modulefiles:/opt/cray/ari/modulefiles:/opt/cray/pe/craype-targets/default/modulefiles:/opt/cray/pe/modulefiles:/opt/cray/modulefiles:/opt/modulefiles"
 
+else if ( "$L_MACHINE" == wcoss2 ) then
+   module purge
+   module load envvar/1.0
+
 else
    module purge
 endif
--- a/etc/lmod-setup.csh
+++ b/etc/lmod-setup.csh
@@ -34,9 +34,6 @@ else if ( "$L_MACHINE" == odin ) then
    module --initial_load --no_redirect restore
    setenv MODULEPATH "/oldscratch/ywang/external/hpc-stack/modulefiles/mpi/intel/2020/cray-mpich/7.7.16:/oldscratch/ywang/external/hpc-stack/modulefiles/compiler/intel/2020:/oldscratch/ywang/external/hpc-stack/modulefiles/core:/oldscratch/ywang/external/hpc-stack/modulefiles/stack:/opt/cray/pe/perftools/21.02.0/modulefiles:/opt/cray/ari/modulefiles:/opt/cray/pe/craype-targets/default/modulefiles:/opt/cray/pe/modulefiles:/opt/cray/modulefiles:/opt/modulefiles"
 
-else if ( "$L_MACHINE" == wcoss2 ) then
-   module reset
-
 else
    module purge
 endif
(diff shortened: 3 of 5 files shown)


Externals.cfg
```diff
@@ -30,7 +30,7 @@ protocol = git
 repo_url = https://github.com/NOAA-EMC/UPP
 # Specify either a branch name or a hash but not both.
 #branch = develop
-hash = 394917e
+hash = fbd41a5
 local_path = src/UPP
 required = True
```
JeffBeck-NOAA, line 35: Have these hashes of UFS_UTILS and UPP been tested with the SRW App on other platforms than WCOSS2?

chan-hoo: @JeffBeck-NOAA, this hash has been tested with PR #266 on other machines. Once this PR is approved, I'll change this hash to the original in this PR. This hash is just for testing.

RatkoVasic-NOAA (approved):

Externals.cfg
RatkoVasic-NOAA: Tested on wcoss2. Build was successful.

JeffBeck-NOAA: What is the status of these hashes on Cheyenne?  @mkavulich, @mark-a-potts, do we need to update hpc-stack or otherwise install newer libraries for this PR and for PR #266?

devbuild.sh
```diff
@@ -136,6 +136,12 @@ while :; do
   shift
 done
 
+# Ensure platform name from variance ======================================
+if [[ "${PLATFORM}" == "wcoss2" || "${PLATFORM}" == "cactus" ||
+      "${PLATFORM}" == "dogwood" ]]; then
+  PLATFORM="wcoss2"
+fi
```
danielabdi-noaa: I think it is better to leave this to the user. This looks like we are going back to auto-detecting machine just like `detect_machine.sh` used to do. If the user is on 'cactus/dogwood' and wants to use wcoss2 modulefiles, he can just pretend to be on wcoss2 and pass `-p=wcoss2` instead.

etc/lmod-setup.sh
```diff
@@ -35,6 +35,10 @@ elif [ "$L_MACHINE" = odin ]; then
    module --initial_load --no_redirect restore
    export MODULEPATH="/oldscratch/ywang/external/hpc-stack/modulefiles/mpi/intel/2020/cray-mpich/7.7.16:/oldscratch/ywang/external/hpc-stack/modulefiles/compiler/intel/2020:/oldscratch/ywang/external/hpc-stack/modulefiles/core:/oldscratch/ywang/external/hpc-stack/modulefiles/stack:/opt/cray/pe/perftools/21.02.0/modulefiles:/opt/cray/ari/modulefiles:/opt/cray/pe/craype-targets/default/modulefiles:/opt/cray/pe/modulefiles:/opt/cray/modulefiles:/opt/modulefiles"
 
+elif [ "$L_MACHINE" = wcoss2 ]; then
+   module purge
+   module load envvar/1.0
```
danielabdi-noaa: Can this go to the modulefiles instead? Unless it is loading Lmod and it is something that can be loaded in modulefiles, I think it should go there.

devbuild.sh
chan-hoo: @danielabdi-noaa, the wcoss system is weird. It switches between `cactus` and `dogwood` on a regular basis. So both `cactus` and `dogwood` are called `wcoss2`. This if-statement prevents a user on wcoss2, either on cactus or on dogwood, from looking into this script or typing --help to figure out what is wrong.

etc/lmod-setup.sh
chan-hoo: @danielabdi-noaa, when a user logs in wcoss2, the module `envvar/1.0` is automatically loaded by default. In the wcoss2 user wiki, the following is described in the section of "Lmod and User Environments": `Please note that when doing a module purge you need to reload envvar, module load envvar/1.0, in your environment (interactive or batch) to get the appropriate default modules. NCO has implemented the standard of doing a “module reset”. This implements the module purge and module load of envvar/1.0 in one step.` So I think this command would be better to be placed here. What do you think about this?

devbuild.sh
danielabdi-noaa: @chan-hoo But then you would have to make the same changes in `test/build.sh` as well. I still think it is better to keep it simple and let the user explicitly specify he wants to use wcoss2 modulefiles for the build. 
I was about to suggest that instead of changing `devbuild.sh` and `test/build.sh`, to provide modulefiles for `cactus/dogwood` that are sym-linked to `wcoss2` ones, but there is lots of code specific for wcoss2 `if [ platform = "wcoss" ]` that it probably complicates the code even more ...

chan-hoo: @danielabdi-noaa, I understand what your concern is. We might have a different viewpoint. I stand on a user's point of view, while you are on a developer's point of view :) :) If you feel uncomfortable about this statement, I'll remove it.

etc/lmod-setup.sh
danielabdi-noaa: I think `module load envar/1.0` can go to the top of `build_wcoss2_*`. Have you tested this change and found it to cause a problem? It looks like it is loading default modules which should not be any different from loading custom modulefiles of SRW. It is not changing the module app, as is the case for example on `gaea/odin` going from Cray->Lmod or on macos where you have no Lmod when you start. Also, the way you have it now if the user does a regular `module purge` on the command line, he has to follow it up with `module load envvar/1.0`, which would be avoided if you have the latter in `build_wcoss2`. But for this to be the case, you have to test  loading envvar in `build_wcoss2` works in the first place ...

devbuild.sh
christopherwharrop-noaa: @chan-hoo - I agree with @danielabdi-noaa, the user needs to be responsible for knowing which machine they are on and specify the correct machine.  The use of if-then-else clauses and checking for specific machines is an anti-pattern that leads to overly complicated hard-to-understand and unmaintainable code.  The names of machines should not appear anywhere in the UFS code or scripts, and I will not approve of changes that includes that.  Please remove this check.

etc/lmod-setup.sh
christopherwharrop-noaa: Once again, machine specific things need to be isolated to the machine-specific configuration.  I agree with @danielabdi-noaa. This belongs in the wcoss configuration module file.  It doesn't make sense here, and pollutes the code with machine-specific things in places users and developers don't expect.

christopherwharrop-noaa (changes requested):

etc/lmod-setup.sh
chan-hoo: @danielabdi-noaa, okay, I'll test it and move it to `build_wcoss2_*` unless it cases any issues.

chan-hoo: @danielabdi-noaa , @christopherwharrop-noaa , what do you think of replacing "module purge and module load envvar" with "module reset"?

devbuild.sh
chan-hoo: @christopherwharrop-noaa, okay, no problem. I'll remove it. My concern is that all users are not as smart as you expect :) :)

etc/lmod-setup.sh
christopherwharrop-noaa: If you are putting it in the `build_wcoss*` modulefile, then you can do whatever you want there as that is completely isolated behavior for that particular machine configuration.  My objection is the test for machine name in a non-machine-specific config/module file.

chan-hoo: Do you mean you don't like `module reset`, either? It is an official nco standard on wcoss2.

devbuild.sh
christopherwharrop-noaa: @chan-hoo - Thank you for making that change.  One lesson I've learned the hard way is that making things stupid-proof is a fool's errand.  It's just not possible and causes more problems than it solves.  Knowing what machine you are using is a reasonable expectation.

etc/lmod-setup.csh
```diff
@@ -34,6 +34,9 @@ else if ( "$L_MACHINE" == odin ) then
    module --initial_load --no_redirect restore
    setenv MODULEPATH "/oldscratch/ywang/external/hpc-stack/modulefiles/mpi/intel/2020/cray-mpich/7.7.16:/oldscratch/ywang/external/hpc-stack/modulefiles/compiler/intel/2020:/oldscratch/ywang/external/hpc-stack/modulefiles/core:/oldscratch/ywang/external/hpc-stack/modulefiles/stack:/opt/cray/pe/perftools/21.02.0/modulefiles:/opt/cray/ari/modulefiles:/opt/cray/pe/craype-targets/default/modulefiles:/opt/cray/pe/modulefiles:/opt/cray/modulefiles:/opt/modulefiles"
 
+else if ( "$L_MACHINE" == wcoss2 ) then
+   module reset
+
```
christopherwharrop-noaa: @chan-hoo - This file is problematic from a design point of view because it is unscalable/unmaintainable/unportable/untestable.  This stuff should be moved to some kind of individual machine-specific configuration file.  But, since you are just following the existing template, I can't argue with this change right now.  But this should be flagged as something that needs rethinking with respect to the overall design.

christopherwharrop-noaa (approved):

etc/lmod-setup.csh
chan-hoo: @christopherwharrop-noaa, thank you for the comment. If you still think that it is better to separate two module commands and move envvar to build_wcoss2_intel, please let me know. I'll talk about this with my EMC colleagues.

etc/lmod-setup.sh
christopherwharrop-noaa: There are really two problems.  First, manipulation of modules really belongs in the machine-specific modulefiles.  The other big problem is the `elif [ "$L_MACHINE" = wcoss2 ]`.  That is bad design and should be fixed.  The conditional tests in this file for which machine you are on needs to be removed and moved from the code/scripts to machine-specific configurations.  But, it appears that you are just following the existing template here, and so I can approve this as is for now because fixing this design problem is beyond the scope of your proposed changes.

dshawul (approved):

danielabdi-noaa (approved):

etc/lmod-setup.sh
chan-hoo: @christopherwharrop-noaa, now I got what your point was. You pointed out that the if-statement itself in the 'lmod-setup.sh' file would cause the issues you mentioned. Is this correct? I thought that you were talking about the command 'module load envvar'. I'll remove 'wcoss2' from the if-statement. Sorry for my late grasp! :) :)

danielabdi-noaa: @chan-hoo I think it is Ok both ways. It is desirable if sourcing `etc/lmod-setup.sh/csh` can be avoided, one less step. On most systems that is actually the case and all you get if you do source it is the `module purge` in the `else:` condition.  In my opinion, those two etc files should be restricted to setting up Lmod, and loading default modules (envvar) can be considered part of setting up Lmod. If it was upto me, I would rather have the user load Lmod manually before building SRW :), but I know your stand on this. Thanks for addressing the requests.

natalie-perlin:
@gspetro-NOAA - is WCOSS2 being supported as Level 1 systems? It seems like WCOSS2 has been removed from the supported platforms, following the discussion on SRW App meeting on May 24, 2022 (but I may be wrong).

chan-hoo:
@natalie-perlin, WCOSS1 (wcoss_dell_p3) will go away in June. So it will be removed from SRW App soon (maybe in July). Instead, WCOSS2 (cactus/dogwood) will be on the list of the tier-1 platforms.

mark-a-potts:
> @natalie-perlin, WCOSS1 (wcoss_dell_p3) will go away in June. So it will be removed from SRW App soon (maybe in July). Instead, WCOSS2 (cactus/dogwood) will be on the list of the tier-1 platforms.

For the release, I think we should leave WCOSS2 off. It is not yet operational and I am not sure we have all the required conda packages installed there, yet. If you like, you can add a "coming soon..." for it, though.

chan-hoo:
@mark-a-potts, yes. I agree. Both WCOSS1 (wcoss_dell_p3) and WCOSS2 will not be available on this release version.

BenjaminBlake-NOAA (approved):
I was able to build the codes successfully on WCOSS2.

PR #276, opened 2022-05-25
Fix artifact creation for branches with / in name (#275)

mark-a-potts:
Jenkins fix for branches with slashes.


mark-a-potts:
I merged this directly.

PR #263, opened 2022-05-19
Updated hpc-stack modules and compiler flags for building SRW release/public-v2 on MacOS

natalie-perlin:
## DESCRIPTION OF CHANGES: 
Changed a build template for the MacOS system to use the hpc-stack modules built for the release/public-v2 SRW version, a file `./modulefiles/build_macos_gnu` .
Recipe for building the hpc-stack for this release on a MacOS using the GNU compiler suite, for either M1/arm64 or Intel/x86_64 chips could be found in HPC-stack repo:
https://github.com/NOAA-EPIC/hpc-stack.git, release/public-v2 branch; it will be later merged with the hpc-stack NOAA-EMC repository.  Currently it could be cloned as following:
`git clone -b release/public-v2 https://github.com/NOAA-EPIC/hpc-stack.git hpc-stack.git`

## TESTS CONDUCTED: 
The SRW was build and tested (single test 2019051600) based on the hpc-stack for release/public-v2. Tho MacOS systems tested: M1/arm64 with Darwin 20/Big Sure, Intel/x86_64 with Darwin 19/Catalina, build has been also tested on M1/arm64 with Darwin 21/Monterey.

## DEPENDENCIES:
HPC-stack recipe built for this release:
- NOAA-EPIC/hpc-stack/release/public-v2
- NOAA-EMC/hpc-stack/pull/250

## DOCUMENTATION:
The documentation is under a separate PR

## ISSUE (optional): 
1. This resolves dependence of syncing the modules loaded to build and run the SRW release/public-v2 on MacOS.
2. Modulefiles build_macos_gnu, wflow_macos could be loaded. Env. variable LDFLAGS needs to be set up (exported) after loading build_macos_gnu; wflow_macos needs to have "puts" in place of "system" calls.
  
  See discussion under the [PR #238 ](https://github.com/ufs-community/ufs-srweather-app/pull/238)
  
  Contributors:
  @danielabdi-noaa

PR #240, opened 2022-04-20
Include documentation for building/running SRW App on Mac

gspetro-NOAA:
## DESCRIPTION OF CHANGES: 
- Updated hpc-stack submodule to reflect new instructions for building the hpc-stack on MacOS
- Added section to BuildRunSRW.rst on configuring the build on MacOS 

## TESTS CONDUCTED: 
None required. @natalie-perlin and @laurenfrederick have built the hpc-stack on MacOS using the submodule info; Natalie has build the SRW on MacOS following the updated procedure in BuildRunSRW.rst. 

## DEPENDENCIES:
None. 

## DOCUMENTATION:
If this PR is contributing new capabilities that need to be documented, please also include updates to the RST files (docs/UsersGuide/source) as supporting material.

## ISSUE: 
Resolves Issue #222 .

## CONTRIBUTORS (optional): 
@natalie-perlin @laurenfrederick

Issue #222, opened 2022-03-08
ufs-srweather-app no longer builds successfully on MacOS


mkavulich:
# Description

See also https://github.com/ufs-community/regional_workflow/issues/684

The original PR implementing MacOS capabilities on the develop branch (#539), instructions were provided in the form of a Google Doc (https://docs.google.com/document/d/14jlvL3nOi85NJCWSNnhHjjBsrAsR-mr8JfGFvVIybFI/edit). These tests worked at the time on my MacOS Catalina 10.15.7 machine, using gnu 9.4.0 compilers installed with Homebrew. However, the app will no longer build, specifically due to a UFS_UTILS compilation error as described below.

## Steps to Reproduce

1. Install hpc-stack on a MacOS machine using [the MacOS instructions](https://docs.google.com/document/d/14jlvL3nOi85NJCWSNnhHjjBsrAsR-mr8JfGFvVIybFI/edit)
2. Attempt to build ufs-srweather-app using the same instructions
3. Observe that UPP and ufs-weather-model build successfully, but UFS_UTILS fails with the following error:

```
[ 42%] Building Fortran object sorc/fvcom_tools.fd/CMakeFiles/fvcom_tools_lib.dir/module_nwp_base.f90.o
/Volumes/d1/workdir/UFS/test_develop_gnu_9/ufs-srweather-app/src/UFS_UTILS/sorc/fvcom_tools.fd/module_nwp_base.f90:16:37:

   16 |    use kinds, only: r_kind, r_single, rmissing
      |                                     1
Error: Symbol 'rmissing' referenced at (1) not found in module 'kinds'
make[5]: *** [sorc/fvcom_tools.fd/CMakeFiles/fvcom_tools_lib.dir/module_nwp_base.f90.o] Error 1
make[4]: *** [sorc/fvcom_tools.fd/CMakeFiles/fvcom_tools_lib.dir/all] Error 2
make[4]: *** Waiting for unfinished jobs....
```

## Additional Context

I have not been able to isolate exactly why this error is occurring. Nothing relevant in the fvcom_tools_lib.dir directory of UFS_UTILS seems to have changed since the last successful tests. I saw the same failure with both the develop branch and tag v1.2.0 of hpc-stack (the tag originally tested), and when I attempted to use the older hash from when this capability was first merged, I was met with CMake errors:

```
-- Configuring done
CMake Error at sorc/chgres_cube.fd/CMakeLists.txt:30 (add_executable):
[...]

PR #281, opened 2022-05-27
Updated the Introduction, build for MacOS

natalie-perlin:
## DESCRIPTION OF CHANGES: 
- Updated the Introduction, systems and storage requirements, software packages
- Updated instructions for the SRW build on MacOS systems

## TESTS CONDUCTED: 
Document changes only

## DOCUMENTATION:
If this PR is contributing new capabilities that need to be documented, please also include updates to the RST files (docs/UsersGuide/source) as supporting material.

## ISSUE (optional): 
Issue mentioned in #222, #238, #240, #263

## CONTRIBUTORS (optional): 
@gspetro-NOAA

### DIFF
--- a/docs/UsersGuide/source/Introduction.rst
+++ b/docs/UsersGuide/source/Introduction.rst
@@ -14,9 +14,6 @@ The SRW App v1.0.0 citation is as follows and should be used when presenting res
 
 UFS Development Team. (2021, March 4). Unified Forecast System (UFS) Short-Range Weather (SRW) Application (Version v1.0.0). Zenodo. https://doi.org/10.5281/zenodo.4534994
 
-..
-   COMMENT: Update version numbers/citation for release! Also update release date for citation!
-
 
 How to Use This Document
 ========================
@@ -112,9 +109,6 @@ The following software is also required to run the SRW Application, but the :ter
 * MPI (MPICH, OpenMPI, or other implementation)
 
    * Only **MPICH** or **OpenMPI** can be built with HPC-Stack. Other implementations must be installed separately by the user. 
-
-..
-    COMMENT: All the software libraries are removed from the list, as they are built by the HPC stack.
 
 For MacOS systems, some additional software is needed. It is recommended that users install this software using the `Homebrew <https://brew.sh/>`__ package manager for MacOS:
 
--- a/docs/UsersGuide/source/Introduction.rst
+++ b/docs/UsersGuide/source/Introduction.rst
@@ -70,12 +70,14 @@ The UFS SRW Application has been designed so that any sufficiently up-to-date ma
 
 * POSIX-compliant UNIX-style operating system
 
-* >102 GB disk space
+* >82 GB disk space
 
-   * 54 GB input data from GFS, RAP, and HRRR for "out-of-the-box" SRW App case described in :numref:`Chapter %s <BuildRunSRW>`
+   * 53 GB input data for a standard collection of global database, or "fix" data (topography, climatology, observational database), and boundary conditions for a short 12-h test forecast on CONUS 25km domain. See data download instructions in :numref:`Step %s <DownloadingStagingInput>`.
    * 8 GB for :term:`HPC-Stack` full installation
    * 3 GB for ufs-srweather-app installation
-   * 17 GB for 12hr forecast on CONUS 25km domain
+   * 1 GB boundary conditions for a short 12-h test forecast on CONUS 25km domain. See data download instructions in :numref:`Step %s <DownloadingStagingInput>`
+   * 17 GB for a 12-h test forecast on CONUS 25km domain, with model output saved hourly, see :numref:`Step %s <GridSpecificConfig>`
+
 
 * 4GB memory (CONUS 25km domain)
 
--- a/docs/UsersGuide/source/Introduction.rst
+++ b/docs/UsersGuide/source/Introduction.rst
@@ -51,7 +51,7 @@ The instructions in this documentation assume that users have certain background
 * Familiarity with LINUX/UNIX systems
 * Command line basics
 * System configuration knowledge (e.g., compilers, environment variables, paths, etc.)
-* Numerical Weather Prediction
+* Numerical Weather Prediction (concepts of parameterizations: physical, microphysical, convective)
 * Meteorology (particularly meteorology at the scales being predicted)
 
 ..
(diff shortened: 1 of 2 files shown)


danielabdi-noaa:
Looks good to me

danielabdi-noaa (approved):

docs/UsersGuide/source/Introduction.rst
```diff
@@ -73,12 +73,12 @@ The UFS SRW Application has been designed so that any sufficiently up-to-date ma
...
 * POSIX-compliant UNIX-style operating system
 
-* >40 GB disk space
+* >102 GB disk space
 
-   * 18 GB input data from GFS, RAP, and HRRR for "out-of-the-box" SRW App case described in :numref:`Chapter %s <BuildRunSRW>`
-   * 6 GB for :term:`HPC-Stack` full installation
-   * 1 GB for ufs-srweather-app installation
-   * 11 GB for 48hr forecast on CONUS 25km domain
+   * 54 GB input data from GFS, RAP, and HRRR for "out-of-the-box" SRW App case described in :numref:`Chapter %s <BuildRunSRW>`
```
gspetro-NOAA: @natalie-perlin If users are only using one of these (e.g., GFS data only), would it still be 54 GB? This should all reflect the _minimum_ requirements.

docs/UsersGuide/source/Introduction.rst
```diff
@@ -88,58 +88,58 @@ The UFS SRW Application has been designed so that any sufficiently up-to-date ma
...
 
-   * netCDF (C and Fortran libraries)
-   * HDF5 
-   * ESMF 8.2.0
-   * Jasper
-   * libJPG
-   * libPNG
-   * zlib
+..
+    COMMENT: All the software libraries are removed from the list, as they are built by the HPC stack.
```
gspetro-NOAA: @natalie-perlin Makes sense! Can you delete this comment before you merge?

docs/UsersGuide/source/Introduction.rst
```diff
@@ -88,58 +88,58 @@ The UFS SRW Application has been designed so that any sufficiently up-to-date ma
...
 
 For MacOS systems, some additional software is needed. It is recommended that users install this software using the `Homebrew <https://brew.sh/>`__ package manager for MacOS:
 
+* brew install bash  
+
+  * OR:  brew upgrade bash
```
gspetro-NOAA, line 119: @natalie-perlin Should we just list the software instead of the commands? The commands already appear in the instructions for installing HPC-Stack and in the MacOS-specific SRW instructions. At a minimum, I don't think we need the `brew upgrade bash` command because we're just listing what they should have on their system. The upgrade statement appears elsewhere in the docs.

docs/UsersGuide/source/Introduction.rst
```diff
@@ -88,58 +88,58 @@ The UFS SRW Application has been designed so that any sufficiently up-to-date ma
...
 For MacOS systems, some additional software is needed. It is recommended that users install this software using the `Homebrew <https://brew.sh/>`__ package manager for MacOS:
 
+* brew install bash  
+
+  * OR:  brew upgrade bash 
 * brew install gcc@11
 * brew install cmake
 * brew install make
 * brew install wget
+* brew install lmod
```
gspetro-NOAA, line 124: @natalie-perlin I thought lmod was recommended, not required. Is it required for MacOS?

docs/UsersGuide/source/Introduction.rst
```diff
@@ -88,58 +88,58 @@ The UFS SRW Application has been designed so that any sufficiently up-to-date ma
...
 Optional but recommended prerequisites for all systems:
 
 * Conda for installing/managing Python packages
 * Bash v4+
 * Rocoto Workflow Management System (1.3.1)
 * Python packages ``scipy``, ``matplotlib``, ``pygrib``, ``cartopy``, and ``pillow`` for graphics
-* Lmod
 
+..
+   COMMENT: Lmod is listed as required
```
gspetro-NOAA, line 138: @mark-a-potts said that Lmod wasn't required to build/run the SRW. Recommended, yes, but you don't have to have it for the SRW to be able to run. It's just more complicated and less supported.

docs/UsersGuide/source/Introduction.rst
natalie-perlin: @gspetro-NOAA - yes, it's only the fix and GFS data for the 12-hour run. The tar archive for the new release is over 53 GB (and this was before addition of the missing 12-h file, another 324MB)

docs/UsersGuide/source/Introduction.rst
natalie-perlin: The important point here is that bash used to be included with the MacOS system until recently, when Apple stopped supporting and upgrading to a newer bash versions. The latest recent version of bash distributed with the macOS is 
`GNU bash, version 3.2.57(1)-release (arm64-apple-darwin20)`
whereas at least bash v.4 is recommended. 
So while it is present in MacOS, it cannot be simply upgraded using homebrew, because there is no previous `brew` installation of bash .
Thus the comments about 
`brew install bash` and `brew upgrade bash`

docs/UsersGuide/source/Introduction.rst
natalie-perlin: Lmod is needed to handle all the SRW modules, so it is needed. It ma. y not be needed for the HPC-stack, if users use the stack for some other software and not SRW, and compile the HPC-stack without the "-m" option. In this case no modules  are created, and all the binaries, libraries, include files from different software packages would be placed together under corresponding `bin/` , `lib/ `, `lib64/`, or `include/` directories

docs/UsersGuide/source/Introduction.rst
natalie-perlin: @gspetro-NOAA  It may not be required, but in current instructions we only show how to build the SRW when Lmod is installed. Note that all the Level 1 systems use the "srw_common" module, which requres use of Lmod.

gspetro-NOAA: @natalie-perlin That's true re Level 1 systems, but Lmod is already pre-installed on all Level 1 systems, so it's a non-issue there. The idea is to provide the minimum requirements you'd need to make SRW work on _any_ system. I think we still need to say that Lmod is highly recommended but not required to run the SRW, but we can say that there is limited (or no) support for using the SRW without Lmod. The point of this section is not really to say what will run best. It's so users can know whether it is even possible to run the SRW on their system. If the team wants to change the section to be "minimum _supported_ requirements," I think that would be reasonable, but you should probably ask @mark-a-potts to see what he thinks. Maybe you can bring it up at the SRW meeting today? I think that would be a reasonable change and easier to maintain. @laurenfrederick @mark-a-potts

docs/UsersGuide/source/Introduction.rst
gspetro-NOAA: @natalie-perlin Then the prerequisite we list should be bash v4.x. The logistics of installing it are in other places. If you prefer, I can make that change in my PR tho.

docs/UsersGuide/source/Introduction.rst
natalie-perlin: Yes, removing this comment

docs/UsersGuide/source/Introduction.rst
natalie-perlin: Some more changes will need to be done, after some data updates. Let me follow with this as we prepare the data for the upload!

gspetro-NOAA (approved):

PR #284, opened 2022-05-31
Update unified test script to run e2e tests

jessemcfarland:
## DESCRIPTION OF CHANGES: 
Execute the end-to-end test workflow during the test stage and archive
the test results.

## TESTS CONDUCTED: 
Add the _run_ci_ label to this PR to execute the end-to-end test workflow on all platforms. Jenkins will report back a passed or failed status based on the results of all platform/compiler combinations. Individual results can be retrieved from the `test_results-<platform>-<compiler>.txt` artifact file on Jenkins.

### DIFF
--- a/.cicd/scripts/srw_test.sh
+++ b/.cicd/scripts/srw_test.sh
@@ -17,5 +17,120 @@ else
     workspace="$(cd -- "${script_dir}/../.." && pwd)"
 fi
 
+we2e_experiment_base_dir="${workspace}/experiments"
+we2e_test_dir="${workspace}/regional_workflow/tests/WE2E"
+
+we2e_test_file="${we2e_test_dir}/experiments.txt"
+
+# The default set of end-to-end tests to run.
+# TODO: Create a list of additional tests that can be run when a parameter
+# is set to true.
+declare -a we2e_default_tests
+we2e_default_tests=('grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16'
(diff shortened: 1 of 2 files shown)


mark-a-potts (approved):

.cicd/scripts/srw_test.sh
```diff
@@ -17,5 +17,120 @@ else
     workspace="$(cd -- "${script_dir}/../.." && pwd)"
 fi
 
+we2e_experiment_base_dir="${workspace}/experiments"
```
gsketefian, line 20: @jessemcfarland The script `get_expts_status.sh` in `ufs-srweather-app/regional_workflow/tests/WE2E` does things similar to this script (but not exactly the same).  You can get help for that by using `get_expts_status.sh --help`.  It would be good to combine the two so the automated tests and any manual tests the user runs use the same set of scripts.

jessemcfarland: @gsketefian thanks! I'll check it out and try to incorporate it here.

dshawul: There is also `setup_WE2E_tests.sh` under `regional_workflow/tests/WE2E` that is used to run 10 regression tests in `machine_suites` on hera/jet using ci-hera/jet-* labels. I think modifying that for other platforms should avoid duplicate work.

danielabdi-noaa: There is also `setup_WE2E_tests.sh` under `regional_workflow/tests/WE2E` that is used to run 10 regression tests in `machine_suites` on hera/jet using ci-hera/jet-* labels. I think modifying that for other platforms should avoid duplicate work.

PR #293, opened 2022-06-02
updated version of ufs to include WoFS suite and ccpp-v6

mark-a-potts:
This change updates the version of the UFS Weather Model to a cherry-picked version based off the hash from April 4 but also includes the WoFS suite.

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -19,11 +19,11 @@ required = True
 
 [ufs-weather-model]
 protocol = git
-#repo_url = https://github.com/ufs-community/ufs-weather-model
-repo_url = https://github.com/NOAA-EPIC/ufs-weather-model-1
+repo_url = https://github.com/ufs-community/ufs-weather-model
+#repo_url = https://github.com/NOAA-EPIC/ufs-weather-model-1
 # Specify either a branch name or a hash but not both.
 #branch = develop
-hash = release/public-v3
+hash = release-public-v3-SRW
 local_path = src/ufs-weather-model
 required = True
 
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -1,6 +1,7 @@
 [regional_workflow]
 protocol = git
-repo_url = https://github.com/ufs-community/regional_workflow
+#repo_url = https://github.com/ufs-community/regional_workflow
+repo_url = https://github.com/NOAA-EPIC/regional_workflow
 # Specify either a branch name or a hash but not both.
 #branch = develop
 hash = release/public-v2
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -4,7 +4,7 @@ protocol = git
 repo_url = https://github.com/NOAA-EPIC/regional_workflow
 # Specify either a branch name or a hash but not both.
 #branch = develop
-branch = release/public-v2
+branch = feature/hera-fix
 local_path = regional_workflow
 required = True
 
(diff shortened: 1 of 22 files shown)


Externals.cfg
```diff
@@ -18,19 +18,19 @@ required = True
 
 [ufs-weather-model]
 protocol = git
-repo_url = https://github.com/ufs-community/ufs-weather-model
+repo_url = https://github.com/NOAA-EPIC/ufs-weather-model-1
 # Specify either a branch name or a hash but not both.
```
JeffBeck-NOAA, line 23: @mark-a-potts, should this point to ufs-community and not NOAA-EPIC?  Or is this just for testing?

JeffBeck-NOAA (approved):

Externals.cfg
mark-a-potts: Just for testing right now.

mark-a-potts: I tried to update the pointers to ccpp/physics and ccpp/framework, but that broke the build and I had to revert to the hashes we had on April 4. I am worried that that means we will not have release/public-v6 of ccpp in our release. Is this a reason to delay?

JeffBeck-NOAA: OK, we'll just have to see if the WoFS SDF XML will work without the pointer updates.  I've emailed @ligiabernardet  and @grantfirl to ask about the status of the v6 CCPP release branch and whether it includes the major module updates that went into the ufs-weather-model repo recently.

mark-a-potts: Thanks, @JeffBeck-NOAA  I emailed Grant as well. He had commits to release-v6 today, so I am thinking that even if they are ready on their side, the changes at the fv3atm level have not been merged yet. @arunchawla-NOAA recommended on slack that we should probably delay the release of the SRW. I am on the fence, but if we do the release, I don't think it can be with ccpp-v6.

JeffBeck-NOAA: @mark-a-potts, just read through the email threads related to the physics/fv3atm release branches.  I would agree with your assessment that it's best to delay by about a week.  I'd hope it won't take too long to get the branch situation straightened out, but there are several submodule branches/pointers that have to be lined up with the correct cherry-picked PRs.  It looks like Grant has taken a big step already toward sorting this out.  However, I'm wondering if the resulting release branch tree will still require the major module updates that currently don't work in the App.  If that does turn out to be the case, I'm happy to quickly track down the required changes to srw_common and get a PR open into the App-level release branch.

mark-a-potts: I agree. Let's postpone the release and focus on getting the physics/branch/submodule situation lined up to everyone's satisfaction. I think we add considerable value to the release by incorporating the latest physics package.

devbuild.sh
```diff
@@ -136,6 +136,11 @@ while :; do
   shift
 done
 
+# Ensure uppercase / lowercase ============================================
```
gsketefian, line 139: FYI the `^^` and `,,` bashisms don't work on a Mac because Apple apparently doesn't/can't update its version of bash on MacOs (and whatever version they're frozen on doesn't support these features).  That's why in `regional_workflow`, we instead use the `echo_uppercase` and `echo_lowercase` functions, e.g. `COMPILER=$(echo_lowercase $COMPILER)`.  To be able to do that, you have to first source `source_util_funcs.sh` in `regional_workflow/ush`.

PR #305, opened 2022-06-13
[release/public-v2] fixes for cheyenne and addition of FV3_WoFS_v0 to build

mark-a-potts:
Fixes modules for Cheyenne.

### DIFF
--- a/modulefiles/build_cheyenne_gnu
+++ b/modulefiles/build_cheyenne_gnu
@@ -8,11 +8,11 @@ proc ModulesHelp { } {
 module-whatis "Loads libraries needed for building SRW on Cheyenne"
 
 module load cmake/3.22.0
+module load python/3.7.9
 module load ncarenv/1.3
 module load gnu/11.2.0
 module load mpt/2.25
 module load ncarcompilers/0.5.0
-module load python/3.7.9
 module unload netcdf
 
 module use /glade/work/epicufsrt/GMTB/tools/gnu/11.2.0/hpc-stack-v1.2.0/modulefiles/stack
--- a/modulefiles/build_cheyenne_intel
+++ b/modulefiles/build_cheyenne_intel
@@ -7,11 +7,13 @@ proc ModulesHelp { } {
 
 module-whatis "Loads libraries needed for building SRW on Cheyenne"
 
-module purge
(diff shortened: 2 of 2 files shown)


christopherwharrop-noaa:
My initial test of this failed for Intel on Cheyenne:

```
CMake Error at /glade/u/apps/ch/opt/cmake/3.22.0/share/cmake-3.22/Modules/CMakeFindDependencyMacro.cmake:47 (find_package):
  Could not find a package configuration file provided by "w3emc" with any of
  the following names:

    w3emcConfig.cmake
    w3emc-config.cmake

  Add the installation prefix of "w3emc" to CMAKE_PREFIX_PATH or set
  "w3emc_DIR" to a directory containing one of the above files.  If "w3emc"
  provides a separate development package or SDK, be sure it has been
  installed.
Call Stack (most recent call first):
  /glade/work/epicufsrt/GMTB/tools/intel/2022.1/hpc-stack-v1.2.0_6eb6/intel-2022.1/mpt-2.25/nemsio/2.5.4/lib64/cmake/nemsio/nemsio-config.cmake:41 (find_dependency)
  CMakeLists.txt:56 (find_package)


-- Configuring incomplete, errors occurred!
```

I'm going to retry in a fresh login shell to rule out any funny business with previous things I was doing.

christopherwharrop-noaa (approved):
My initial failed test was due to a corrupted environment from previous work in the same shell.  A retest of Intel build worked fine, and the gnu build also worked.

I am curious why the GNU build for the release is using version 11.2 while develop is using 10.1.  I would normally expect new versions to be used on develop first, before being applied to a release, but I'm assuming there is some good explanation for that, and I'm curious to know what it is.

mark-a-potts:
> I am curious why the GNU build for the release is using version 11.2 while develop is using 10.1. I would normally expect new versions to be used on develop first, before being applied to a release, but I'm assuming there is some good explanation for that, and I'm curious to know what it is.

The explanation is probably not as good as would be desired, but it came down to the fact that we had gotten module files updated for 11.2 before the modules were updated for 10.1, so I made the switch. While testing with 10.1 later, I also ran into some problems during the forecast, which I haven't had time to figure out. Fortunately, no problems with 11.2.

PR #313, opened 2022-06-22
[release/public-v2] added a module build_linux_gnu, updated build_macos_gnu

natalie-perlin:
## ISSUE : 
Added a modulefile for Linux build_linux_gnu, 
Updated a modulelefile build_macos_gnu

## CONTRIBUTORS (optional): 
@gspetro-NOAA


modulefiles/build_linux_gnu
```diff
@@ -0,0 +1,87 @@
...
+proc ModulesHelp { } {
+   puts stderr "Load the environmental modules, modules for the compilers if present,"
+   puts stderr "specify compilers, path for HPC-stack and SRW directory on Linux systems" 
+   puts stderr "
+}
+
+module-whatis "Loads libraries needed for building SRW on Linux"
+
+if { [module-info mode load] } {
+   puts "ulimit -S -s unlimited;"
```
dshawul: I found in the past, on a linux machine like hera, `system` has the desired effect instead of `puts`.
Could you check if you get unlimited stack after module loading this file on your system.
Same should go for puts statement in `wflow_linux`

mark-a-potts: I can confirm that the puts works as expected (ulimit -s is unlimited after module load) on my Linux box. Since these are supposed to be templates, I think it might be a good idea to comment out almost everything and put in a stderr message that tells users to modify the file(s) to suit their system.

danielabdi-noaa: I think it is the Lmod version that is the problem. On an ubuntu 18.04 with Lmod 6.6, puts simply prints the command to the screen whether you puts to stdout/stderr, while system has the required effect. It was the opposite for macos though. If the install instructions for Lmod on linux systems lead to the desired behaviour, I guess there will be no problem so I will approve.

danielabdi-noaa (approved):

mark-a-potts (approved):

natalie-perlin:
Corrected build_macos_gnu, build_linux_compiler (a template); 
Corrected wflow_macos, added wflow_linux

Issue #316, opened 2022-06-23
Modify directory layout for operational RRFS needs


MatthewPyle-NOAA:
## Description

Currently, the regional workflow is cloned into a subdirectory under the primary SRW app directory, which is inconsistent with the directory organization needed for NWS operations such as the RRFS implementation.   

## Solution

For operational implementation needs, the workflow contents (directories such as jobs, scripts, ush, ...) need to be in the primary SRW app directory.   Also, the "src" directory would ideally be renamed "sorc".    Much of this could be achieved by checking out the regional_workflow external into "./" instead of "./regional_workflow", but there are some common files/directories in both repositories that complicate this type of solution.

## Alternatives (optional)

Would need to create an add-on script to rearrange things into an operationally appropriate structure, which could work but is less than ideal.

PR #345, opened 2022-09-08
Add back the dependency on GSI for rrfs_utl.

christinaholtNOAA:
## DESCRIPTION OF CHANGES: 
The dependency on GSI for RRFS_utl was inadvertently removed in a recent PR. This will add it back, and should fix the build script attempting to build all the components together.

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## TESTS CONDUCTED: 

- [ ] hera.intel
- [ ] orion.intel
- [ ] cheyenne.intel
- [ ] cheyenne.gnu
- [ ] gaea.intel
- [ ] jet.intel
- [ ] wcoss2.intel
- [ ] NOAA Cloud (indicate which platform)
- [ ] Jenkins
- [ ] fundamental test suite
- [ ] comprehensive tests (specify *which* if a subset was used)

## DEPENDENCIES:
None.

## CHECKLIST

- [ ] My code follows the style guidelines in the Contributor's Guide
- [x] I have performed a self-review of my own code using the Code Reviewer's Guide
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes need updates to the documentation. I have made corresponding changes to the documentation
- [x] My changes do not require updates to the documentation (explain).
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] Any dependent changes have been merged and published

## LABELS (optional): 

A Code Manager needs to add the following labels to this PR: 
- [ ] Work In Progress
- [ ] bug
- [ ] enhancement
- [ ] documentation
- [ ] release
- [ ] high priority
- [ ] run_ci
- [ ] run_we2e_fundamental_tests
- [ ] run_we2e_comprehensive_tests
- [ ] Needs Cheyenne test 
- [ ] Needs Jet test 
- [ ] Needs Hera test 
- [ ] Needs Orion test 
- [ ] help wanted

## CONTRIBUTORS (optional): 
@danielabdi-noaa

PR #348, opened 2022-09-12
Changes towards making SRW app NCO compliant

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 
Changes to make SRW app NCO complaint following the [standard](https://www.nco.ncep.noaa.gov/idsb/implementation_standards/ImplementationStandards.v11.0.0.pdf) and [review](https://docs.google.com/document/d/1qEK5Rs24An3x7iwpOEXaNoo4D32BqhhNN8p8mRtAKcg/edit).

Here are some of the changes in this PR:
- Follow new NCO directory structure and naming convention of output files, log files, temporary directory etc
- Make any WE2E test case to be run in either `community` or `nco` mode
- Also make verification tasks sort of nco complaint but this is not run in production
- Use standard production environment variables listed in Table 1 of standard
- Symlink fix files by default -- optionally can be turned off
- Build to "exec" directory activated
- Activate debugging mode by optionally setting `set -eux`
- Make j-job and ex-script names consistent
- Add versions of libraries specifically for wcoss2
- Add a unique workflow ID for resolving clashes between different files, while keeping relation between different tasks in same workflow run
- Adds unittest for end-to-end workflow generation using `config.community.yaml` and `config.nco.yaml`

## TESTS CONDUCTED: 

- Run fundamental tests (9 test cases) on **Hera**, **Jet** and **Orion** manually using both `community` and `nco` mode.

- Run fundamental tests on **Gaea** through [Jenkins](https://jenkins-epic.woc.noaa.gov/blue/organizations/jenkins/ufs-srweather-app%2Fpipeline/detail/PR-376/5/pipeline/230)

- Run the comprehensive test (82 test cases) on Hera using **NCO** mode. All of the test cases are successful.
  
  Some unexpected failures.
    - ~~`get_from_HPSS*` tasks that fail when run in tandem with other tests, but not if run individually. *Cause:* Probably HPSS access issue with multiple processes?~~ After a fix for conflicting ics/lbcs temporay working directories in NCO mode all test cases run to completion in one attempt. I suspect that I was very lucky to get this result since develop branch can sometimes fail all of them.
    - ~~ensemble runs going into second day forecast not being run. *Cause:* make_grid_complete.txt is stored in first day's directory. Not relevant for operations since make_grid is not run, but necessary to run every test case in NCO mode. Find a better way to communicate grid generation completion than using dummy file?~~ 
    Solved by moving grid/orog completion files to a shared space for all cycle dates in `EXPTDIR/grid(orog)(sfc_climo)`
    - ~~`community_ensemble_2mems_stoch` fails. *Cause:* This test case generates new `input.nml` per ensemble member.~~ Solved by outputting `input.nml` to whatever is current working directory.

  The test directory and rocot stats files:
   ```
   /scratch1/BMC/zrtrr/Daniel.Abdi/rrfs/benchmark-nco
   /scratch1/BMC/zrtrr/Daniel.Abdi/rrfs/stat-nco.txt
   ```
  The NCO operations directory:
  ```
  /scratch1/BMC/zrtrr/Daniel.Abdi/rrfs/OPSROOT
  ```
  The subdirectories `com`, `output` and `tmp` are the ones with content.

- Run the comprehensive test (82 test cases) on Hera using **COMMUNITY** mode. All of the test cases are successful.

  The test directory and rocot stats files:
  ```
  /scratch1/BMC/zrtrr/Daniel.Abdi/rrfs/benchmark-com
  /scratch1/BMC/zrtrr/Daniel.Abdi/rrfs/stat-com.txt
  ```
  
## DEPENDENCIES:
None

## DOCUMENTATION:
None but would need documentation updates

## ISSUE (optional): 
#316 

## CONTRIBUTORS (optional): 
@MatthewPyle-NOAA @christinaholtNOAA


chan-hoo:
@danielabdi-noaa, thank you so much for this work!!! In this PR, you newly define some directories such as HOMErrfs, PARMrrfs, and USHrrfs. This kind of naming only works for RRFS. For example, Online-CMAQ (AQMv7) is an extension of the UFS SRW App. For this implementation, we should name them as HOMEaqm, PARMaqm, and USHaqm. The model name is defined with the "NET" variable in the app. For example, NET=rrfs for RRFS, and NET=aqm for Online-CMAQ. Could you change the way to define these directory names like "HOME{NET}", "PARM{NET}", and "USH{NET}"? Is this possible?

danielabdi-noaa:
@chan-hoo Thanks! We can probably re-name HOMErrfs, USHrrfs etc to be HOME{NET}, USH{NET} in the ex-scripts and j-jobs, since all we have to do is modify what gets written to the `var_defns.sh` file accordingly. However, for the python workflow generation code itself, we would have to stick with one name. HOMErrfs would be Ok for that I guess. There would be a disparity between the workflow generation code and ex-scripts/j-job names if we decide to go this route. Let me know what you think.

chan-hoo:
@danielabdi-noaa, I agree with you! This requirement is only for the "J-job" (and its ex-script).

chan-hoo:
@danielabdi-noaa, what about naming it with "model" such as "HOMEmodel", "USHmodel", and "PARMmodel" in the python script?

MatthewPyle-NOAA:
There probably isn't a great solution for this issue (with the SRW App intended for multiple operational systems), but NCO will really want a $HOMErrfs for RRFS and a $HOMEaqm for AQM.  Maybe a generic placeholder like $HOMEmodel could be modified ahead of final code delivery to what it needs to be for the specific application.

chan-hoo:
@danielabdi-noaa @MatthewPyle-NOAA, If so, the "rrfs" suffix would be fine for the python script. I'll modify it in my forked repo for Online-CMAQ when AQMv7 is delivered to NCO.

MatthewPyle-NOAA:
@danielabdi-noaa I noticed a small issue with the build for WCOSS2.  The version defining variable for wrf_io is inconsistently named between versions/build.ver.wcoss2 (export wrf_io_ver=1.2.0) and  modulefiles/build_wcoss2_intel ( module load wrf_io/$::env(wrfio_ver) )  Could you modify build_wcoss2_intel to pull in wrf_io_ver?

It built for me on WCOSS2 after making this change.

danielabdi-noaa:
@MatthewPyle-NOAA Thanks, just fixed it in the repo.

danielabdi-noaa:
@chan-hoo After starting to work on the change, I am realizing it may not be the best idea. 
```Python
+        f"HOME{NET}": HOMErrfs,
+        f"USH{NET}": USHrrfs,
+        f"SCRIPTS{NET}": SCRIPTSrrfs,
+        f"JOBS{NET}": JOBSrrfs,
+        f"SORC{NET}": SORCrrfs,
+        f"PARM{NET}": PARMrrfs,
+        f"MODULES{NET}": MODULESrrfs,
+        f"EXEC{NET}": EXECrrfs,
+        f"FIX{NET}": FIXrrfs,
```
In the ex-scripts/j-job scripts we would have to parametrize the variable name itself. If we had an m4 macro processor run over the scripts, maybe it would have been easier, but for now replacing HOMErrfs etc with the appropriate name before delivery sounds better.

chan-hoo:
@danielabdi-noaa, I got it. I'll replace them manually for AQM. Thank you!

MatthewPyle-NOAA:
@danielabdi-noaa One more thing jumped out at me when doing some testing.  Would it be possible to make DATAROOT a distinctly specifiable variable, rather than having it specified as OPSROOT/tmp/?  While production does run that way, development work on WCOSS generally has DATAROOT and COMROOT on different temporary disks that have different retention periods.

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/chgres_cube does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/emcsfc_ice_blend does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/emcsfc_snow2mdl does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/filter_topo does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/fregrid does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/fvcom_to_FV3 does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/global_cycle does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/global_equiv_resol does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/inland does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/lakefrac does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/make_hgrid does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/make_solo_mosaic does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/upp.x does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/orog does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/orog_gsl does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/regional_esg_grid does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/sfc_climo_gen does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/shave does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/ufs_model does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/vcoord_gen does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/enkf.x does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/gsi.x does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/nc_diag_cat.x does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/ncdiag_cat_serial.x does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/test_nc_unlimdims.x does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/adjust_soiltq.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/check_imssnow_fv3lam.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/fv3lam_nonvarcldana.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/gen_annual_maxmin_GVF.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/gen_cs.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/gen_ensmean_recenter.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/lakesurgery.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/process_imssnow_fv3lam.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/process_larccld.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/process_Lightning.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/process_metarcld.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/process_NSSL_mosaic.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/process_updatesst.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/ref2tten.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/update_bc.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/update_GVF.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/update_ice.exe does NOT exist'
+ echo 'FAIL: intel executable file /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914185011/ufs-srweather-app/test/../bin_intel/bin/use_raphrrr_sfc.exe does NOT exist'
+ echo 'BUILD(S) FAILED'
+ msg=FAIL
+ echo FAIL
Build failed
ci-hera-intel-WE

danielabdi-noaa:
@MatthewPyle-NOAA It should be possible but I've had doubts on how NCO variables are specified. Currently the variables are in the `var_defns.sh` file under the nco section that the user can edit to his liking.
```Bash
# [nco]
envir='para'
NET='rrfs'
RUN='nco_ensemble'
model_ver='we2e'
OPSROOT='/scratch1/BMC/zrtrr/Daniel.Abdi/rrfs/OPSROOT'
COMIN_BASEDIR='/scratch1/BMC/zrtrr/Daniel.Abdi/rrfs/OPSROOT/com/rrfs/we2e'
COMOUT_BASEDIR='/scratch1/BMC/zrtrr/Daniel.Abdi/rrfs/OPSROOT/com/rrfs/we2e'
COMROOT='/scratch1/BMC/zrtrr/Daniel.Abdi/rrfs/OPSROOT/com'
PACKAGEROOT='/scratch1/BMC/zrtrr/Daniel.Abdi/rrfs/OPSROOT/packages'
DATAROOT='/scratch1/BMC/zrtrr/Daniel.Abdi/rrfs/OPSROOT/tmp'
DCOMROOT='/scratch1/BMC/zrtrr/Daniel.Abdi/rrfs/OPSROOT/dcom'
DBNROOT=''
SENDECF='FALSE'
SENDDBN='FALSE'
SENDDBN_NTC='FALSE'
SENDCOM='FALSE'
SENDWEB='FALSE'
KEEPDATA='TRUE'
MAILTO=''
MAILCC=''

```
If some of these variables are specified by an export from the environment, currently it won't take effect because both j-jobs and ex-scripts source the `var_defns.sh` file again and overwrite the specified values. So how are these variables specified in operations? We may need to completely remove them from `var_defns.sh` if they are exported from the environment.

venitahagerty:
Machine: jet
Compiler: intel
Job: WE
Repo location: /lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/chgres_cube does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/emcsfc_ice_blend does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/emcsfc_snow2mdl does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/filter_topo does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/fregrid does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/fvcom_to_FV3 does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/global_cycle does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/global_equiv_resol does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/inland does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/lakefrac does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/make_hgrid does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/make_solo_mosaic does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/upp.x does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/orog does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/orog_gsl does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/regional_esg_grid does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/sfc_climo_gen does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/shave does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/ufs_model does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/vcoord_gen does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/enkf.x does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/gsi.x does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/nc_diag_cat.x does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/ncdiag_cat_serial.x does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/test_nc_unlimdims.x does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/adjust_soiltq.exe does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/check_imssnow_fv3lam.exe does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/fv3lam_nonvarcldana.exe does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/gen_annual_maxmin_GVF.exe does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/gen_cs.exe does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/gen_ensmean_recenter.exe does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/lakesurgery.exe does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/process_imssnow_fv3lam.exe does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/process_larccld.exe does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/process_Lightning.exe does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/process_metarcld.exe does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/process_NSSL_mosaic.exe does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/process_updatesst.exe does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/ref2tten.exe does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/update_bc.exe does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/update_GVF.exe does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/update_ice.exe does NOT exist'
+ echo 'FAIL: intel executable file /mnt/lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914185012/ufs-srweather-app/test/../bin_intel/bin/use_raphrrr_sfc.exe does NOT exist'
+ echo 'BUILD(S) FAILED'
+ msg=FAIL
+ echo FAIL
Build failed
ci-jet-intel-WE

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914192016/ufs-srweather-app
Build was Successful
Rocoto jobs started
Long term tracking will be done on 9 experiments
ci-hera-intel-WE

MatthewPyle-NOAA:
@danielabdi-noaa Sorry about that - I saw something in setup.py, not thinking that I could fix it in var_defns.sh.   

The highest level items like OPSROOT and DATAROOT  (items listed as "job card" in Table 1 of the implementation standards) are specified by the ecflow jobs in production.  For rocoto-driven jobs, they would ideally be defined somewhere at the rocoto or rocoto-driving script level.  At the J-job and below, the expectation is that they would be available as environmental variables.

dshawul:
@MatthewPyle-NOAA I forgot to mention that you can actually export OPSROOT, DATAROOT etc before workflow generation and it will take those values, instead of the deffaults in `setup.py` so you don't really need to edit `var_defns.sh`. I was thinking about the situation after the workflow is generated, in which case the only way to specifiy it by editing `var_defns.sh`. If `SENDDBN` etc is something that is decided before workflow is generated and does not change afterwards, we are all good.

venitahagerty:
Machine: jet
Compiler: intel
Job: WE
Repo location: /lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220914192010/ufs-srweather-app
Build was Successful
Rocoto jobs started
Long term tracking will be done on 9 experiments
ci-jet-intel-WE

danielabdi-noaa:
@MatthewPyle-NOAA The code currently assumes that if COMROOT is defined in the environment everything else is. You may have tried to set only DATAROOT, which is probably why it didn't work for you.
```Python
        try:
            COMROOT
        except NameError:
            OPSROOT = OPSROOT or EXPTDIR
            COMROOT = os.path.join(OPSROOT, "com")
            PACKAGEROOT = os.path.join(OPSROOT, "packages")
            DATAROOT = os.path.join(OPSROOT, "tmp")
            DCOMROOT = os.path.join(OPSROOT, "dcom")
```
If these variables can be set individually (not in bulk), I can fix the code so that it checks if the environment defined these variables individually.

MatthewPyle-NOAA:
@danielabdi-noaa I think adding separate checks and definitions for these variables would be safest.   Thanks!

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220914223511/ufs-srweather-app
Build was Successful
Rocoto jobs started
Long term tracking will be done on 9 experiments
ci-hera-intel-WE

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220915033512/ufs-srweather-app
Build was Successful
Rocoto jobs started
Long term tracking will be done on 9 experiments
ci-hera-intel-WE
Experiment Succeeded on hera: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_RRFS_v1beta
Experiment Succeeded on hera: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_HRRR

venitahagerty:
Machine: jet
Compiler: intel
Job: WE
Repo location: /lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1053825830/20220915033513/ufs-srweather-app
Build was Successful
Rocoto jobs started
Long term tracking will be done on 9 experiments
ci-jet-intel-WE
Experiment Failed on jet: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_HRRR
2022-09-15 04:20:10 +0000 :: fe2 :: Task get_extrn_lbcs, jobid=11477743, in state DEAD (OUT_OF_MEMORY), ran for 182.0 seconds, exit status=253, try=1 (of 1)
Experiment Succeeded on jet: grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v15p2
Experiment Succeeded on jet: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
Experiment Succeeded on jet: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
Experiment Succeeded on jet: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_RAP_suite_HRRR
Experiment Succeeded on jet: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR
Experiment Succeeded on jet: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
Experiment Succeeded on jet: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_RRFS_v1beta

MatthewPyle-NOAA:
@danielabdi-noaa I was playing around with the individual component builds.  Does it make sense that rrfs_utils can't be built by itself?  Doing so gives me this cmake error.  Seemingly any combination of components without GSI gives this failure.

CMake Error at /apps/cmake/3.20.1/share/cmake-3.20/Modules/ExternalProject.cmake:3111 (get_property):
  get_property could not find TARGET GSI.  Perhaps it has not yet been
  created.

danielabdi-noaa:
@MatthewPyle-NOAA Ah we went back and forth with that issue. rrfs_utls depends on GSI so technically it can not be built if GSI is not already built, so the only way right now is to build them together.
```
./devbuild.sh -p=hera gsi rrfs_utils
```
I removed the dependency in order to meet the NCO requirements but it caused an issue in the automated testing because without the DEPENDS clause, it tires to build rrfs_utls before GSI. CMakeLists does not seem to respect the order in CMakeLists.txt, so the change got reverted just recently in PR #345 .

.github/workflows/build.yml
```diff
@@ -12,10 +12,5 @@ jobs:
       - run: |
              cd ufs-srweather-app
              ./manage_externals/checkout_externals
-             module use modulefiles
-             module load build_cheyenne_intel
-             mdkir build
-             cd build
-             cmake -DCMAKE_INSTALL_PREFIX=.. ..
-             make -j 4
+             ./devbuild.sh -p=cheynne -c=intel
```
christinaholtNOAA: cheynne -> cheyenne

chan-hoo:
@danielabdi-noaa, I got an error when I ran a WE2E test with your branch: "End get_WE2Etest_names_subdirs_descs.sh at Tue Sep 20 18:31:45 UTC 2022 with error code 1 (time elapsed: 00:00:02)". Should I run it differently? What I did: "./run_WE2E_tests.sh tests_file="my_expts.txt" machine=orion account="fv3-cam" use_cron_to_relaunch=TRUE cron_relaunch_intvl_mnts=03 >& we2e.log &".

danielabdi-noaa:
@chan-hoo That looks correct to me. Maybe it is an issue with Orion because, on hera a similar script seems to be fine.
If it fails during workflow generation, maybe there is some bad code that `set -e` captures now, which is activated only in this PR. I will try to do a test on Orion later -- this PR has only been tested on Hera and Jet so far.
```Bash
#!/bin/bash

DIR="/scratch1/BMC/zrtrr/Daniel.Abdi/rrfs"

./run_WE2E_tests.sh \
    tests_file=test.txt \
    machine=hera \
    account="zrtrr" \
    expt_basedir="$DIR/benchmark-test" \
    verbose="TRUE" \
    opsroot="${DIR}/OPSROOT-test" \
    run_envir="nco"
```

chan-hoo:
@danielabdi-noaa, ok. I'll test it on Hera. Hera is quite slow these days.

danielabdi-noaa:
@chan-hoo I think it is like I suspected. `set -e` has uncovered faulty code in `get_WE2Etest_names_subdirs_descs.sh`. Removing that, it generates and runs the workflow properly. Since it is good to have `set -e` so we will have to fix the issue script. It was failing around these lines:
```Bash
+ (( k<=10 ))
+ var_name=INCR_CYCL_FREQ
++ grep '^[ ]*INCR_CYCL_FREQ='
+ cmd=
+ postamble get_WE2Etest_names_subdirs_descs.sh 1663704350 1
+ set +x
```
Another issue on Orion seems to be there is no system `python3` so you would need to load the modulefiles and activate conda before you run a test case.

danielabdi-noaa:
@chan-hoo I have made the fix and run fundamental tests on Orion now. Let me know if you face further issues.

natalie-perlin (approved):
Looks like a huge work and notable enhancement of the SRW code!

chan-hoo (approved):

.github/workflows/python_unittests.yaml
```diff
@@ -15,13 +15,17 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install python3 python3-pip netcdf-bin
-          sudo pip3 install pyyaml jinja2 f90nml
+          sudo pip3 install pyyaml jinja2==2.11 f90nml
           sudo pip3 install numpy matplotlib basemap
+          sudo ln -sf /bin/bash /bin/sh
```
christinaholtNOAA: Do we have scripts that are not explicitly /bin/bash? We should def make sure we fix that! 

Does it make sense to set the default shell environment here?

```
defaults:
  run:
    shell: bash
```

.github/workflows/python_unittests.yaml
```diff
@@ -15,13 +15,17 @@ jobs:
...
 
       # Run python unittests
       - name: Run python unittests
         run: |
+          ./manage_externals/checkout_externals ufs-weather-model
           cd ush
           python3 -m unittest -b python_utils/test_python_utils.py
-          python3 -m unittest -b *.py
+          files=$(find . -maxdepth 1 -name '*.py' -a ! -name 'test_retrieve_data.py' -exec basename {} \;)
```
christinaholtNOAA, line 31: Should we leave a comment here that retrieve_data is tested in the functional tests workflow? I def spent a minute trying to understand why we'd want to exclude that test file.

Externals.cfg
```diff
@@ -3,7 +3,7 @@ protocol = git
 repo_url = https://github.com/ufs-community/UFS_UTILS
```
christinaholtNOAA: We don't want to update hashes here, do we?

docs/UsersGuide/source/Graphics.rst
```diff
@@ -196,36 +196,36 @@ At a minimum, the account should be set appropriately prior to job submission:
...
+   export HOMEdir=/path-to/ufs-srweather-app/regional_workflow
    export EXPTDIR1=/path-to/experiment/directory1
    export EXPTDIR2=/path-to/experiment/directory2
 
 If the user's login shell is csh/tcsh, they can be set as follows:
 
 .. code-block:: console
 
-   setenv HOMErrfs /path-to/ufs-srweather-app/regional_workflow
+   setenv HOMEdir /path-to/ufs-srweather-app/regional_workflow
```
christinaholtNOAA, line 228: Thanks for this @danielabdi-noaa!

modulefiles/srw_common
```diff
@@ -17,7 +17,7 @@ module load g2/3.4.5
 module load g2tmpl/1.10.0
 module load ip/3.3.3
 module load sp/2.3.3
-module load w3nco/2.4.1
```
christinaholtNOAA: I'm not sure we want to change any of the module versions here.

parm/FV3LAM_wflow.xml
```diff
@@ -229,28 +259,32 @@ MODULES_RUN_TASK_FP script.
...
   {%- endif %}
     <jobname>&MAKE_SFC_CLIMO_TN;</jobname>
-    <join><cyclestr>&LOGDIR;/&MAKE_SFC_CLIMO_TN;.log</cyclestr></join>
+    <join><cyclestr>&LOGDIR;/&MAKE_SFC_CLIMO_TN;&LOGEXT;</cyclestr></join>
 
     <envar><name>GLOBAL_VAR_DEFNS_FP</name><value>&GLOBAL_VAR_DEFNS_FP;</value></envar>
+    <envar><name>PDY</name><value><cyclestr>@Y@m@d</cyclestr></value></envar>
+    <envar><name>cyc</name><value><cyclestr>@H</cyclestr></value></envar>
+    <envar><name>subcyc</name><value><cyclestr>@M</cyclestr></value></envar>
+    <envar><name>LOGDIR</name><value><cyclestr>&LOGDIR;</cyclestr></value></envar>
```
christinaholtNOAA, line 276: This one doesn't have a real date associated with it, so shouldn't need the cycle information.

parm/FV3LAM_wflow.xml
```diff
@@ -266,20 +300,22 @@ MODULES_RUN_TASK_FP script.
...
     <jobname>&GET_EXTRN_ICS_TN;</jobname>
-    <join><cyclestr>&LOGDIR;/&GET_EXTRN_ICS_TN;_@Y@m@d@H.log</cyclestr></join>
+    <join><cyclestr>&LOGDIR;/&GET_EXTRN_ICS_TN;_@Y@m@d@H&LOGEXT;</cyclestr></join>
 
     <envar><name>GLOBAL_VAR_DEFNS_FP</name><value>&GLOBAL_VAR_DEFNS_FP;</value></envar>
     <envar><name>PDY</name><value><cyclestr>@Y@m@d</cyclestr></value></envar>
+    <envar><name>cyc</name><value><cyclestr>@H</cyclestr></value></envar>
+    <envar><name>subcyc</name><value><cyclestr>@M</cyclestr></value></envar>
+    <envar><name>LOGDIR</name><value><cyclestr>&LOGDIR;</cyclestr></value></envar>
     <envar><name>CDATE</name><value><cyclestr>@Y@m@d@H</cyclestr></value></envar>
```
christinaholtNOAA: Is CDATE an NCO-compliant variable? Can we remove it here and in the other tasks if not so that we aren't duplicating that information?

rename_model.sh
```diff
@@ -0,0 +1,25 @@
...
+    echo "Usage: $0 s_suffix t_suffix"
+    echo " Example: $0 rrfs aqm"
+    exit 0
+fi
+
+# directories to look for model variables to rename
+mdirs=( "docs" "jobs" "parm" "scripts" "test" "tests" "ush" )
+
+# model variables to rename
+mvars=( "home" "parm" "ush" "sorc" "modules" "exec" "scripts" "jobs" "fix" "tests" "we2e" )
```
christinaholtNOAA, line 13: Did we leave this many `rrfs` variables in there?

scripts/exregional_make_ics.sh
```diff
@@ -638,20 +621,23 @@ Please check the following user defined variables:
...
-  cp_vrfy ${fvcom_data_fp} ${ics_dir}/fvcom.nc
-  cd_vrfy ${ics_dir}
-  ${RUN_CMD_UTILS} ${fvcom_exec_fn} sfc_data.tile${TILE_RGNL}.halo${NH0}.nc fvcom.nc ${FVCOM_WCSTART} ${fvcom_time}|| \
-  print_err_msg_exit "\
+  cp_vrfy ${fvcom_data_fp} ${INPUT_DATA}/fvcom.nc
+  cd_vrfy ${INPUT_DATA}
+  PREP_STEP
+  eval ${RUN_CMD_UTILS} ${fvcom_exec_fn} \
+       ${NET}.${cycle}${dot_ensmem}.sfc_data.tile${TILE_RGNL}.halo${NH0}.nc fvcom.nc ${FVCOM_WCSTART} ${fvcom_time} \
+       ${REDIRECT_OUT_ERR} || print_err_msg_exit "\
```
christinaholtNOAA, line 629: Why do we need eval now?

scripts/exregional_run_fcst.sh
```diff
@@ -505,20 +511,17 @@ code."
 #
 if [ ${WRITE_DOPOST} = "TRUE" ]; then
 
-  yyyymmdd=${cdate:0:8}
-  hh=${cdate:8:2}
+  yyyymmdd=${CDATE:0:8}
+  hh=${CDATE:8:2}
   cyc=$hh
   fmn="00"
```
christinaholtNOAA, line 526: All date references should be in terms of the NCO variables set at the jobs level, I think. It's probably find to have these date definitions here for use later in the file, but perhaps something like:
```
yyyymmdd=${PDY}
hh=${cyc}
```
@MatthewPyle-NOAA Is this allowed?

scripts/exregional_run_vx_ensgrid.sh
```diff
@@ -106,35 +82,45 @@ NUM_PAD=${NDIGITS_ENSMEM_NAMES}
...
+    export INPUT_BASE=$COMIN
+    export OUTPUT_BASE=$COMOUT/metout
+    export MEM_BASE=$OUTPUT_BASE
+    export LOG_DIR=$LOGDIR
 
-if [ ${VAR} == "APCP" ]; then
-  LOG_SUFFIX=ensgrid_${CDATE}_${VAR}_${ACCUM}h
+    export POSTPRD=
+    export MEM_STAR=
+    export MEM_CUSTOM=
```
christinaholtNOAA, line 96: It might be better to use `unset` here (and in other similar instances) so that if these are used later, it throws an error and we know we need more logic for NCO vs community mode.

tests/WE2E/machine_suites/jet.txt
```diff
@@ -1,9 +0,0 @@
-grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
```
christinaholtNOAA, line 1: Why is this one removed?

tests/WE2E/setup_WE2E_tests.sh
```diff
@@ -35,7 +35,7 @@ function usage {
 
 }
 
-machines=( hera jet )
+machines=( hera jet cheyenne orion wcoss2 gaea odin singularity macos noaacloud )
```
christinaholtNOAA, line 38: We do only want these for hera and jet. A different mechanism tests the WE2E on the other platforms.

tests/WE2E/setup_WE2E_tests.sh
```diff
@@ -62,13 +62,13 @@ EXPTS_DIR=${TOP_DIR}/expt_dirs
 # Set the path to the machine-specific test suite file.
 #-----------------------------------------------------------------------
 
-auto_file=${scrfunc_dir}/machine_suites/${machine}.txt
+auto_file=${scrfunc_dir}/machine_suites/fundamental.txt
```
christinaholtNOAA: We do want this to still be a test suite for the machine.

tests/WE2E/setup_WE2E_tests.sh
```diff
@@ -62,13 +62,13 @@ EXPTS_DIR=${TOP_DIR}/expt_dirs
...
 
 #----------------------------------------------------------------------
 # Use exec_subdir consistent with the automated build.
 #----------------------------------------------------------------------
 
-exec_subdir='bin_intel/bin'
+exec_subdir='bin_intel/exec'
```
christinaholtNOAA: Should this be `exec_intel/exec`?

ush/set_ozone_param.py
```diff
@@ -187,10 +187,11 @@ def set_ozone_param(ccpp_phys_suite_fp):
 
 class Testing(unittest.TestCase):
     def test_set_ozone_param(self):
+        USHdir = os.path.dirname(os.path.abspath(__file__))
         self.assertEqual(
             "ozphys_2015",
             set_ozone_param(
-                ccpp_phys_suite_fp=f"test_data{os.sep}suite_FV3_GSD_SAR.xml"
+                ccpp_phys_suite_fp=f"{USHdir}{os.sep}test_data{os.sep}suite_FV3_GSD_SAR.xml"
```
christinaholtNOAA, line 194: I didn't know about os.sep!

ush/setup.py
```diff
@@ -1270,250 +1306,80 @@ def setup():
...
     #
-    global RUN_TASK_MAKE_GRID, RUN_TASK_MAKE_OROG, RUN_TASK_MAKE_SFC_CLIMO
     global GRID_DIR, OROG_DIR, SFC_CLIMO_DIR
-    global RUN_TASK_VX_GRIDSTAT, RUN_TASK_VX_POINTSTAT, RUN_TASK_VX_ENSGRID
 
     #
-    # -----------------------------------------------------------------------
-    #
-    # Make sure that DO_ENSEMBLE is set to TRUE when running ensemble vx.
+    # If RUN_TASK_MAKE_GRID is set to False, the workflow will look for
```
christinaholtNOAA, line 1306: We probably want to set these tasks to not run for any of the NCO configurations that currently exist in the tests suite, or in the ush/config.nco.yaml files.

christinaholtNOAA (commented):
Phew! @danielabdi-noaa that was a LOT! Nice work. Thanks!

I left some comments/questions/concerns below. They aren't all necessarily for you, specifically.

.github/workflows/build.yml
danielabdi-noaa: yes, later commit have fixed this issue.

Externals.cfg
danielabdi-noaa: That came from Chan-hoo's commit. It should not have shown up on my PR though, maybe I should do a `git rebase` instead of `git merge`.

ush/setup.py
MatthewPyle-NOAA: @christinaholtNOAA I would agree that the config.nco.yaml should have a workflow_switches section to make it easy to exclude or include tasks in the setup.

.github/workflows/python_unittests.yaml
danielabdi-noaa: It is the sourcing bash config file from python that is problematic
```Python
    code = dedent(
        f"""      #!/bin/bash
      t1="./t1.{pid}"
      t2="./t2.{pid}"
      (set -o posix; set) > $t1
      {{ . {config_file}; set +x; }} &>/dev/null
      (set -o posix; set) > $t2
      diff $t1 $t2 | grep "> " | cut -c 3-
      rm -rf $t1 $t2
    """
    )   
    (_, config_str, _) = run_command(code)
```
I will specify the default shell instead of symlinking.

scripts/exregional_make_ics.sh
danielabdi-noaa: `REDIRECT_OUT_ERR` needs it. Before it was only `RUN_CMD_UTILS` for either `srun/mpirun` that needed it and was done elswhere. Since we need both to be evaled, I though it is easier to do them together.

.github/workflows/python_unittests.yaml
danielabdi-noaa: Done

Externals.cfg
danielabdi-noaa: I have rebased my PR to develop and these do not show up anymore.

modulefiles/srw_common
danielabdi-noaa: This also came from merge issue now resolved.

parm/FV3LAM_wflow.xml
danielabdi-noaa: I know this and in other places in the xml file some variables may not be needed or be empty, but in NCO standards Table 1, a variable `cycle` need to be defined which i do in `ush/job_preamble.sh` as either `t{hh}z` or `t{hh}{mm}z`, although I've not tested the latter with a subhourly test case (existing one is broken). I think it is good to be consistent with defining these basic variables in all tasks even though some may not need it.

parm/FV3LAM_wflow.xml
danielabdi-noaa: It is not NCO required and was already there. Some of the j-job/ex-scripts use it, and some use lowercase `cdate`. I did not want to tinker through how to remove it, but it definitely looks removable.

rename_model.sh
danielabdi-noaa: Most of those were in the format `EXECDIR` etc, and I thought that every subdirectory that is under the root directory of UFS SW app need to be named consistently so that renaming the app would be easier. From todays meeting, i understood that it is only `HOMErrfs` that NCO requires the rest can be `PARM`, `USH` etc

tests/WE2E/setup_WE2E_tests.sh
danielabdi-noaa: I needed the script to do the same test as `hera` on `orion` even though we do not have CI enabled there. I recall having to do something similar in other systems to run the same set of fundamental tests manually.

tests/WE2E/setup_WE2E_tests.sh
danielabdi-noaa: The setup script uses `bin_intel/bin` before and i needed to change only the exec subdirectory for NCO. I could try and change both but seems to work fine with it except for the issue with NCO test case not being reported.

tests/WE2E/setup_WE2E_tests.sh
danielabdi-noaa: I thought it would be easier to have one file for fundamental tests that should be same on all machines. (`hera.txt` and `jet.txt` were identical when i removed them). Otherwise I would have to create `orion.txt`, `gaea.txt` etc.

ush/setup.py
danielabdi-noaa: The NCO configuration is no more associated with pre-generated grids though. So if I tell it to set `RUN_TASK_MAKE_GRID=FALSE` in NCO mode, I am sure it will cause problems when trying to run a `community` test case in NCO mode. However, I could do something like that specifically for `machine=WCOSS2` assuming anything run on wcoss2 is in NCO mode -- actually did something similar in the rocoto xml before but got rid of it later.
@MatthewPyle-NOAA Indeed the default values of those are TRUE so definately need to be set to false. I guess i've never run that test case before.

danielabdi-noaa: @christinaholtNOAA  @MatthewPyle-NOAA Good catch! Not only is the `config.nco.yaml` that needs those but also 4 or so other tests cases in WE2E that run in NCO mode needed it. They must have run in commmunity mode in the regression test...

parm/FV3LAM_wflow.xml
danielabdi-noaa: If I recall correctly, code actually breaks with `set -u` turned on if I don't define them.

parm/FV3LAM_wflow.xml
danielabdi-noaa: I just tried to see if I can remove it, and it looks like it is used in many places in ex-scripts and METplus config files. Also there is not an equivalent variable at the moment because PDY does not have HH, so everytime it is needed i would have to do `$PDY$cyc`.

scripts/exregional_run_vx_ensgrid.sh
danielabdi-noaa: This is the hack I mentioned about in getting MET plus conf files to work with both NCO and COMMUNITY mode. It would need to be defined to an "empty string" since it is actually used the conf file, so I don't think `unset` will work. POSTPRD is not used in NCO mode but it is still there in the conf file.

scripts/exregional_run_fcst.sh
danielabdi-noaa: I have made the change according to your suggestion in the ex-scripts and j-jobs. In the process, I noticed something odd in `exregional_get_obs_ccpa.sh` and other similar scripts where a variable
```
init=${CDATE}${hh}
```
I thought CDATE already has the HH.

parm/FV3LAM_wflow.xml
danielabdi-noaa: I have removed CDATE from the xml file and defined it in job_preamble as `$PDY$cyc`. It doesn't completely remove it because of reasons I mentioned earlier, but thought it maybe useful to have as less variables defined in the xml file as possible for the ecFlow transition?

tests/WE2E/setup_WE2E_tests.sh
danielabdi-noaa: I think i found a good compromise for this. If a machine specific list of test cases is available such as `hera.txt` or `jet.txt`, that will be used, otherwise we fall back on to `fundamental.txt`. That way we can remove/add a test case for specific machines.

tests/WE2E/setup_WE2E_tests.sh
danielabdi-noaa: I changed this to `install_intel/exec` since the directory `install_intel` is the installation directory and has `include`, `lib` and '`share` besides `exec`.

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220923173517/ufs-srweather-app
Build was Successful
Rocoto jobs started
Long term tracking will be done on 9 experiments
ci-hera-intel-WE
Experiment Succeeded on hera: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_RRFS_v1beta
Experiment Succeeded on hera: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_HRRR

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1053825830/20220923193517/ufs-srweather-app
Build was Successful
Rocoto jobs started
Long term tracking will be done on 9 experiments
ci-hera-intel-WE
Experiment Succeeded on hera: nco_grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR
Experiment Succeeded on hera: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_RRFS_v1beta
Experiment Succeeded on hera: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_HRRR
All experiments completed

danielabdi-noaa:
I've tested the fundamental tests on Gaea through Jenkins since last time.  I am fairly confident it works as intended so going to merge now.

christinaholtNOAA:
@danielabdi-noaa I missed that you'd merged already. Thank you for being so quick and thorough on your responses to reviews! Great job!

PR #414, opened 2022-10-13
[develop] Improve error messaging and logging for generate_FV3LAM_wflow.py

### LINKED ISSUE #385: Improve error trapping and messaging to the end user
## Description
With the switch to a python-based workflow comes a whole new paradigm of error messaging. Users used to bash will not be familiar with python traceback errors, and may be overwhelmed by unnecessary information. This is especially the case where errors are not trapped properly, and so the final error message does not actually indicate the problem that caused the error condition. Furthermore, bash scripts that *call* python scripts do not properly pass error information to the user, leaving the actual problem buried far up in the output.

## Solution
 - Trap errors at a higher level, and hide full error trace from the user (unless DEBUG is active)
 - Ensure that there is only a single traceback with a final error message, rather than multiple exceptions, or exceptions unrelated to the actual error

## Requirements**
Specific items that need improvement:
### Case 1
 - [ ] This is a carry-over from the original bash workflow, but the specific error message should appear at the *end* of the error trace for clarity. As an example, attempting to run WE2E tests before building code gives this traceback:
```
========================================================================
Starting function setup() in "setup.py"...
========================================================================
  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/ush/generate_FV3LAM_wflow.py", line 1115, in <module>
    p.start()
  File "/contrib/miniconda3/4.5.12/envs/regional_workflow/lib/python3.8/multiprocessing/process.py", line 121, in start
    self._popen = self._Popen(self)
  File "/contrib/miniconda3/4.5.12/envs/regional_workflow/lib/python3.8/multiprocessing/context.py", line 224, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File "/contrib/miniconda3/4.5.12/envs/regional_workflow/lib/python3.8/multiprocessing/context.py", line 276, in _Popen
    return Popen(process_obj)
  File "/contrib/miniconda3/4.5.12/envs/regional_workflow/lib/python3.8/multiprocessing/popen_fork.py", line 19, in __init__
    self._launch(process_obj)
  File "/contrib/miniconda3/4.5.12/envs/regional_workflow/lib/python3.8/multiprocessing/popen_fork.py", line 75, in _launch
    code = process_obj._bootstrap(parent_sentinel=child_r)
  File "/contrib/miniconda3/4.5.12/envs/regional_workflow/lib/python3.8/multiprocessing/process.py", line 313, in _bootstrap
    self.run()
  File "/contrib/miniconda3/4.5.12/envs/regional_workflow/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/ush/generate_FV3LAM_wflow.py", line 1103, in workflow_func
    generate_FV3LAM_wflow()
  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/ush/generate_FV3LAM_wflow.py", line 112, in generate_FV3LAM_wflow
    setup()
  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/ush/setup.py", line 296, in setup
    print_err_msg_exit(
  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/ush/python_utils/print_msg.py", line 19, in print_err_msg_exit
    traceback.print_stack(file=sys.stderr)

The base directory in which the FV3 source code should be located
(UFS_WTHR_MDL_DIR) does not exist:
  UFS_WTHR_MDL_DIR = "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/sorc/ufs-weather-model"
Please clone the external repository containing the code in this directory,
build the executable, and then rerun the workflow.
Exiting with nonzero status.
rm: cannot remove ‘/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/ush/tmp’: No such file or directory
  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/ush/generate_FV3LAM_wflow.py", line 1132, in <module>
    rm_vrfy(tmp_fp)
  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/ush/python_utils/filesys_cmds_vrfy.py", line 33, in rm_vrfy
    return cmd_vrfy("rm", *args)
  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/ush/python_utils/filesys_cmds_vrfy.py", line 20, in cmd_vrfy
    print_err_msg_exit(f'System call "{cmd}" failed.')
  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/ush/python_utils/print_msg.py", line 19, in print_err_msg_exit
    traceback.print_stack(file=sys.stderr)
System call "rm /scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/ush/tmp" failed.
Exiting with nonzero status.

ERROR:
  From script:  "run_WE2E_tests.sh"
  Full path to script:  "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/tests/WE2E/test_configs/wflow_features/run_WE2E_tests.sh"
Could not generate an experiment for the test specified by test_name:
  test_name = "grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_RAP_suite_HRRR"
Exiting with nonzero status.
```
The end message is generic, and the actual error (the code hasn't been built yet) is hidden deep in the above trace.

### Case 2
 - [ ] Users who accidentally include an incorrect entry in their config.yaml file are met with this intimidating wall of traceback text:
```
========================================================================
Starting function setup() in "setup.py"...
========================================================================
Process Process-1:
  File "/contrib/miniconda3/4.5.12/envs/regional_workflow/lib/python3.8/multiprocessing/process.py", line 313, in _bootstrap
    self.run()
  File "/contrib/miniconda3/4.5.12/envs/regional_workflow/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "./generate_FV3LAM_wflow.py", line 1103, in workflow_func
    generate_FV3LAM_wflow()
  File "./generate_FV3LAM_wflow.py", line 112, in generate_FV3LAM_wflow
    setup()
  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/test_not_building_execs/ufs-srweather-app/ush/setup.py", line 130, in setup
    set_predef_grid_params()
  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/test_not_building_execs/ufs-srweather-app/ush/set_predef_grid_params.py", line 41, in set_predef_grid_params
    params_dict = params_dict[PREDEF_GRID_NAME]
KeyError: 'RRFS_CONUS_25km_SH'
rm: cannot remove ‘/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/test_not_building_execs/ufs-srweather-app/ush/tmp’: No such file or directory
  File "./generate_FV3LAM_wflow.py", line 1132, in <module>
    rm_vrfy(tmp_fp)
  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/test_not_building_execs/ufs-srweather-app/ush/python_utils/filesys_cmds_vrfy.py", line 33, in rm_vrfy
    return cmd_vrfy("rm", *args)
  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/test_not_building_execs/ufs-srweather-app/ush/python_utils/filesys_cmds_vrfy.py", line 20, in cmd_vrfy
    print_err_msg_exit(f'System call "{cmd}" failed.')
  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/test_not_building_execs/ufs-srweather-app/ush/python_utils/print_msg.py", line 19, in print_err_msg_exit
    traceback.print_stack(file=sys.stderr)
System call "rm /scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/test_not_building_execs/ufs-srweather-app/ush/tmp" failed.
Exiting with nonzero status.
```

### Case 3
 - [ ] Attempting to run `run_WE2E_tests.sh` without setting the proper python environment first leads to a jumbled mess of python and bash error messages:
```
Gathering test descriptions and experiment variable values from the 
configuration files of the primary WE2E tests...

  Reading in the test description for primary WE2E test:  "grid_CONUS_25km_GFDLgrid_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16"
  In category (subdirectory):  "wflow_features"

  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/ush/python_utils/config_parser.py", line 50, in load_yaml_config
    cfg = yaml.safe_load(f)
NameError: name 'yaml' is not defined

During handling of the above exception, another exception occurred:

  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/ush/config_utils.py", line 14, in <module>
    cfg_main()
  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/ush/python_utils/config_parser.py", line 483, in cfg_main
    cfg = load_config_file(args.cfg, 2)
  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/ush/python_utils/config_parser.py", line 425, in load_config_file
    return load_yaml_config(file_name)
  File "/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/ush/python_utils/config_parser.py", line 51, in load_yaml_config
    except yaml.YAMLError as e:
NameError: name 'yaml' is not defined
/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/tests/WE2E/get_WE2Etest_names_subdirs_descs.sh: line 986: ${#test_desc}-1: substring expression < 0
/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/tests/WE2E/get_WE2Etest_names_subdirs_descs.sh: line 517: avail_WE2E_test_names: unbound variable

Performing sanity checks on user-specified list of WE2E tests to run...
/scratch2/BMC/gsd-fv3-dev/kavulich/UFS/tmp/ufs-srweather-app/tests/WE2E/get_WE2Etest_names_subdirs_descs.sh: line 549: avail_WE2E_test_names[@]: unbound variable
```

 - End users should not see long error traces into unrelated packages, but should be met with error messages specific to the context of the end-to-end workflow. Limited error traces with an appropriately obvious error are acceptable, but the current state of error messages (as demonstrated in above examples) will be very intimidating even to people with casual python experience.
 - Workflows that combine bash and python should ensure that error conditions in one are appropriately handled and passed to the other.

mkavulich:
This PR is now open for review

## DESCRIPTION OF CHANGES: 
The error messaging in the generate_FV3LAM_wflow.py has been overhauled to be more pythonic *and* more user-friendly. The old bash-like `print_err_msg_exit()` calls are replaced with proper exception handling, and `print_info_msg()` calls are replaced with the use of Python's built-in logging module. The latter has the added benefit of giving the same "tee"-like functionality--where messages are printed to screen and written to a log file simultaneously--without the need for messy subprocesses and the confusing error messages that come with them. 

Errors are now handled in a consistent way, with the errors clearly highlighted both on screen and in the log file, as well as an easy-to-follow exception traceback for those wanting that information. In addition, several error conditions that were previously **not** caught at the generate step now cause the script to fail with helpful error messages.

From the user perspective, functionality will not change, aside from differences when error conditions are present. Successful runs will remain the same, except that some log messaging has been clarified, and the log *file* contains additional information for each message, as well as some indentation changes for ease of reading.

### Example output
In this example, I specified an invalid variable "INVALID_VAR" in config.yaml. In the current code, the script *succeeds*, which is a dangerous condition that can produce unexpected results:
**Updated code**:
```
./generate_FV3LAM_wflow.py 

  ========================================================================
  Starting experiment generation...
  ========================================================================

  ========================================================================
  Starting function setup() in "setup.py"...
  ========================================================================

*********************************************************************
Experiment generation failed. See the error message(s) printed below.
For more detailed information, check the log file from the workflow
generation script: /mnt/lfs4/BMC/gsd-fv3-dev/kavulich/workdir/PR_414/ufs-srweather-app/ush/log.generate_FV3LAM_wflow
*********************************************************************

  File "./generate_FV3LAM_wflow.py", line 1037, in <module>
    generate_FV3LAM_wflow(USHdir, logfile)
  File "./generate_FV3LAM_wflow.py", line 103, in generate_FV3LAM_wflow
    setup()
  File "/mnt/lfs4/BMC/gsd-fv3-dev/kavulich/workdir/PR_414/ufs-srweather-app/ush/setup.py", line 107, in setup
    raise Exception(dedent(f'''
Exception: 
User-specified variable "INVALID_VAR" in config.yaml is not valid
Check config_defaults.yaml for allowed user-specified variables

```
For this same case, the log file shows the following text:
```
root                   DEBUG    Finished setting up debug file logging in /mnt/lfs4/BMC/gsd-fv3-dev/kavulich/workdir/PR_414/ufs-srweather-app/ush/log.generate_FV3LAM_wflow
root                   DEBUG    Logging set up successfully
generate_FV3LAM_wflow  INFO
  ========================================================================
  Starting experiment generation...
  ========================================================================
setup                  INFO
  ========================================================================
  Starting function setup() in "setup.py"...
  ========================================================================
root                   ERROR
*********************************************************************
Experiment generation failed. See the error message(s) printed below.
For more detailed information, check the log file from the workflow
generation script: /mnt/lfs4/BMC/gsd-fv3-dev/kavulich/workdir/PR_414/ufs-srweather-app/ush/log.generate_FV3LAM_wflow
*********************************************************************

  File "./generate_FV3LAM_wflow.py", line 1037, in <module>
    generate_FV3LAM_wflow(USHdir, logfile)
  File "./generate_FV3LAM_wflow.py", line 103, in generate_FV3LAM_wflow
    setup()
  File "/mnt/lfs4/BMC/gsd-fv3-dev/kavulich/workdir/PR_414/ufs-srweather-app/ush/setup.py", line 107, in setup
    raise Exception(dedent(f'''
Exception:
User-specified variable "INVALID_VAR" in config.yaml is not valid
Check config_defaults.yaml for allowed user-specified variables
```

In the log file you can see the above-mentioned additional information. Before each message, the subroutine writing the message is listed (a few top-level processes involving setting up logging and printing the error message are logged as "root"). Next on the line is a note of the "logging level"; see [Logging HOWTO](https://docs.python.org/3/howto/logging.html) for more details, but basically this denotes the "urgency" of the message, with "debug" being the lowest and "error" being the highest used here, indicating the script has failed in some way. While not implemented here, this will allow us to properly quarantine lower-priority printouts to the log file, or only print them to screen if DEBUG=TRUE, by making use of lower-priority levels such as logging.debug. And finally, log messages that extend to multiple lines are indented by two spaces, to make individual messages easier to parse visually.

### Type of change

## TESTS CONDUCTED: 
- [x] hera.intel
  - Attempted to run as many error condition checks as possible to ensure smooth error handling/messaging
    - [x] Ran ./generate_FV3LAM_wflow.py with a number of different failure conditions
      - [x] Prior to checking out code
      - [x] Prior to building code
        - Note: currently this does *not* exit with an error (in develop or with these new changes), as it would break existing tests. Will need to re-visit this in the future.
      - [x] Prior to loading regional_workflow conda environment
        - Note: On some platforms this error will be impossible to catch gracefully, due to not even being able to start running the script
      - [x] Removed Externals.cfg
      - [x] malformed Externals.cfg
      - [x] With no config.yaml
      - [x] With malformed config.yaml
      - [x] PREEXISTING_DIR_METHOD = quit
      - [ ] Incorrect config.yaml settings
        - [x] Did not include `MACHINE`
        - [x] Did not include `ACCOUNT` with `WORKFLOW_MANAGER=rocoto`
        - [x] Included invalid variable
        - [x] Mismatch between SPP array sizes
        - [x] Mismatch between LSM_SPP array sizes
        - [x] Incorrect PREDEF_GRID_NAME
        - [x] RESTART_INTERVAL set to non-int value
        - [x] DATE_FIRST_CYCL or DATE_LAST_CYCL set to non-date value
        - [x] USE_CUSTOM_POST_CONFIG_FILE true but CUSTOM_POST_CONFIG_FP does not exist
        - [x] USE_CRTM true but CRTM_DIR does not exist
        - [x] FCST_LEN_HRS > 999
        - [x] FCST_LEN_HRS not evenly divisible by LBC_SPEC_INTVL_HRS
        - [x] Mandatory variables (DT_ATMOS, LAYOUT_X, LAYOUT_Y, BLOCKSIZE) not set
        - [x] CCPP_PHYS_SUITE not valid
        - [x] EXTRN_MDL_SOURCE_BASEDIR_ICS, EXTRN_MDL_SOURCE_BASEDIR_LBCS set to non-existent location while USE_USER_STAGED_EXTRN_FILES is True
        - [x] RUN_TASK_VX_ENSGRID or RUN_TASK_VX_ENSPOINT set True while DO_ENSEMBLE is False
        - [x] With RUN_TASK_MAKE_GRID = False
          - [x] Set GRID_DIR to non-existent location
          - [x] Set GRID_DIR to ""
        - [x] With RUN_TASK_MAKE_OROG = False
          - [x] Set OROG_DIR to non-existent location
          - [x] Set OROG_DIR to ""
        - [x] With RUN_TASK_MAKE_SFC_CLIMO = False
          - [x] Set SFC_CLIMO_DIR to non-existent location
          - [x] Set SFC_CLIMO_DIR to ""
        - [x] Set PREEXISTING_DIR_METHOD to ""
      - [x] Incorrect machine file settings
        - [x] "MACHINE" does not correspond to a machine file
        - [x] machine file is malformed
    - [x] Ran ./run_WE2E_tests.sh with a number of different failure conditions
      - [x] Prior to checking out code
      - [x] Prior to loading regional_workflow conda environment
    - [x] Ran ALL WE2E tests, compared output to current develop
      - Develop: /scratch2/BMC/fv3lam/kavulich/UFS/workdir/issue_385/before_and_after_testing/develop/expt_dirs
      - This branch: /scratch2/BMC/fv3lam/kavulich/UFS/workdir/issue_385/before_and_after_testing/after/expt_dirs/
      - [x] No unexpected differences
- [ ] cheyenne.intel
    - [ ] Ran fundamental suite of WE2E tests
- [ ] cheyenne.gnu
    - [ ] Ran fundamental suite of WE2E tests
- [x] Jenkins testing
- [x] All CI checks passed

## DEPENDENCIES:
None

## DOCUMENTATION:
None, though the new, clearer error messaging and removal of extraneous comments could be considered a type of documentation change. 

Check the comments on #385 for more examples of the differences in error messages.

## ISSUE: 
Fixes #385 

## CHECKLIST

### DIFF
--- a/tests/WE2E/run_WE2E_tests.sh
+++ b/tests/WE2E/run_WE2E_tests.sh
@@ -65,6 +65,63 @@ WE2Edir="$TESTSdir/WE2E"
 #
 #-----------------------------------------------------------------------
 #
+# Run python checks
+#
+#-----------------------------------------------------------------------
+#
+
+# This line will return two numbers: the python major and minor versions
+pyversion=($(/usr/bin/env python3 -c 'import platform; major, minor, patch = platform.python_version_tuple(); print(major); print(minor)'))
+
+#Now, set an error check variable so that we can print all python errors rather than just the first
+pyerrors=0
+
+# Check that the call to python3 returned no errors, then check if the 
+# python3 minor version is 6 or higher
+if [[ -z "$pyversion" ]];then
+  print_info_msg "\
+
+  Error: python3 not found"
+  pyerrors=$((pyerrors+1))
+else
+  if [[ ${#pyversion[@]} -lt 2 ]]; then
+    print_info_msg "\
+
+  Error retrieving python3 version"
+    pyerrors=$((pyerrors+1))
+  elif [[ ${pyversion[1]} -lt 6 ]]; then
+    print_info_msg "\
+
+  Error: python version must be 3.6 or higher
+  python version: ${pyversion[*]}"
+    pyerrors=$((pyerrors+1))
+  fi
+fi
+
+#Next, check for the non-standard python packages: jinja2, yaml, and f90nml
+pkgs=(jinja2 yaml f90nml)
+for pkg in ${pkgs[@]}  ; do
+  if ! /usr/bin/env python3 -c "import ${pkg}" &> /dev/null; then
+  print_info_msg "\
+
+  Error: python module ${pkg} not available"
+  pyerrors=$((pyerrors+1))
+  fi
+done
+
+#Finally, check if the number of errors is >0, and if so exit with helpful message
+if [ $pyerrors -gt 0 ];then
+  print_err_msg_exit "\
+  Errors found: check your python environment
+
+  Instructions for setting up python environments can be found on the web:
+  https://github.com/ufs-community/ufs-srweather-app/wiki/Getting-Started
+
+"
+fi
+#
+#-----------------------------------------------------------------------
+#
 # Save current shell options (in a global array).  Then set new options
 # for this script or function.
 #
--- a/ush/generate_FV3LAM_wflow.py
+++ b/ush/generate_FV3LAM_wflow.py
@@ -101,7 +101,7 @@ def generate_FV3LAM_wflow(USHdir, logfile: str = 'log.generate_FV3LAM_wflow') ->
 
     # The setup function reads the user configuration file and fills in 
     # non-user-specified values from config_defaults.yaml
-    setup()
+    setup(logging.getLogger('root.setup'))
 
     # import all environment variables
     import_vars()
--- a/ush/generate_FV3LAM_wflow.py
+++ b/ush/generate_FV3LAM_wflow.py
@@ -101,7 +101,7 @@ def generate_FV3LAM_wflow(USHdir, logfile: str = 'log.generate_FV3LAM_wflow') ->
 
     # The setup function reads the user configuration file and fills in 
     # non-user-specified values from config_defaults.yaml
-    setup(logging.getLogger('root.setup'))
+    setup()
 
     # import all environment variables
     import_vars()
--- a/ush/generate_FV3LAM_wflow.py
+++ b/ush/generate_FV3LAM_wflow.py
@@ -1065,12 +1065,15 @@ if __name__ == "__main__":
     except:
         # If the call to the generate_FV3LAM_wflow function above was not successful, 
         # print out an error message and exit with a nonzero return code.
-        logging.exception(
+        logging.exception(dedent(
             f"""
+            *********************************************************************
             Experiment generation failed. See the error message(s) printed below.
             For more detailed information, check the log file from the workflow
-            generation script: {logfile}"""
-        )
+            generation script: {logfile}
+            *********************************************************************\n
+            """
+        ))
 
 class Testing(unittest.TestCase):
     def test_generate_FV3LAM_wflow(self):
--- a/ush/generate_FV3LAM_wflow.py
+++ b/ush/generate_FV3LAM_wflow.py
@@ -431,8 +431,10 @@ def generate_FV3LAM_wflow(USHdir, logfile: str = 'log.generate_FV3LAM_wflow') ->
                 The variable \"settings\" specifying values of the rococo XML variables
                 has been set as follows:
                 #-----------------------------------------------------------------------
-                settings =\n\n""")
-        log_info(settings_str)
+                settings =\n\n""",
+                verbose=VERBOSE,
+            )
+        log_info(settings_str, verbose=VERBOSE)
 
         #
         # Call the python script to generate the experiment's actual XML file
@@ -472,6 +474,7 @@ def generate_FV3LAM_wflow(USHdir, logfile: str = 'log.generate_FV3LAM_wflow') ->
         workflow launch script (WFLOW_LAUNCH_SCRIPT_FP):
           EXPTDIR = \"{EXPTDIR}\"
           WFLOW_LAUNCH_SCRIPT_FP = \"{WFLOW_LAUNCH_SCRIPT_FP}\"''',
+        verbose=VERBOSE,
     )
 
     create_symlink_to_file(
@@ -499,6 +502,7 @@ def generate_FV3LAM_wflow(USHdir, logfile: str = 'log.generate_FV3LAM_wflow') ->
             Symlinking fixed files from system directory (FIXgsm) to a subdirectory (FIXam):
               FIXgsm = \"{FIXgsm}\"
               FIXam = \"{FIXam}\"''',
+            verbose=VERBOSE,
         )
 
         ln_vrfy(f'''-fsn "{FIXgsm}" "{FIXam}"''')
@@ -509,6 +513,7 @@ def generate_FV3LAM_wflow(USHdir, logfile: str = 'log.generate_FV3LAM_wflow') ->
             Copying fixed files from system directory (FIXgsm) to a subdirectory (FIXam):
               FIXgsm = \"{FIXgsm}\"
               FIXam = \"{FIXam}\"''',
+            verbose=VERBOSE,
         )
 
         check_for_preexist_dir_file(FIXam, "delete")
@@ -534,6 +539,7 @@ def generate_FV3LAM_wflow(USHdir, logfile: str = 'log.generate_FV3LAM_wflow') ->
               FIXaer = \"{FIXaer}\"
               FIXlut = \"{FIXlut}\"
               FIXclim = \"{FIXclim}\"''',
+            verbose=VERBOSE,
         )
 
         check_for_preexist_dir_file(FIXclim, "delete")
@@ -555,23 +561,27 @@ def generate_FV3LAM_wflow(USHdir, logfile: str = 'log.generate_FV3LAM_wflow') ->
     log_info(
         f"""
         Copying templates of various input files to the experiment directory...""",
+        verbose=VERBOSE,
     )
 
     log_info(
         f"""
         Copying the template data table file to the experiment directory...""",
+        verbose=VERBOSE,
     )
     cp_vrfy(DATA_TABLE_TMPL_FP, DATA_TABLE_FP)
 
     log_info(
         f"""
         Copying the template field table file to the experiment directory...""",
+        verbose=VERBOSE,
     )
     cp_vrfy(FIELD_TABLE_TMPL_FP, FIELD_TABLE_FP)
 
     log_info(
         f"""
         Copying the template NEMS configuration file to the experiment directory...""",
+        verbose=VERBOSE,
     )
     cp_vrfy(NEMS_CONFIG_TMPL_FP, NEMS_CONFIG_FP)
     #
@@ -583,6 +593,7 @@ def generate_FV3LAM_wflow(USHdir, logfile: str = 'log.generate_FV3LAM_wflow') ->
         f"""
         Copying the CCPP physics suite definition XML file from its location in
         the forecast model directory sturcture to the experiment directory...""",
+        verbose=VERBOSE,
     )
     cp_vrfy(CCPP_PHYS_SUITE_IN_CCPP_FP, CCPP_PHYS_SUITE_FP)
     #
@@ -594,6 +605,7 @@ def generate_FV3LAM_wflow(USHdir, logfile: str = 'log.generate_FV3LAM_wflow') ->
         f"""
         Copying the field dictionary file from its location in the forecast
         model directory sturcture to the experiment directory...""",
+        verbose=VERBOSE,
     )
     cp_vrfy(FIELD_DICT_IN_UWM_FP, FIELD_DICT_FP)
     #
@@ -838,7 +850,9 @@ def generate_FV3LAM_wflow(USHdir, logfile: str = 'log.generate_FV3LAM_wflow') ->
     log_info(
             f"""
             The variable \"settings\" specifying values of the weather model's
-            namelist variables has been set as follows:\n""")
+            namelist variables has been set as follows:\n""",
+        verbose=VERBOSE,
+    )
     log_info("\nsettings =\n\n" + settings_str)
     #
     # -----------------------------------------------------------------------
--- a/ush/generate_FV3LAM_wflow.py
+++ b/ush/generate_FV3LAM_wflow.py
@@ -917,9 +917,11 @@ def generate_FV3LAM_wflow(USHdir, logfile: str = 'log.generate_FV3LAM_wflow') ->
 
         set_FV3nml_sfc_climo_filenames()
 
-    # Call script to get NOMADS data
+    # Call function to get NOMADS data
     if NOMADS:
-        get_nomads_data(NOMADS_file_type,EXPTDIR,USHdir,DATE_FIRST_CYCL,CYCL_HRS,FCST_LEN_HRS,LBC_SPEC_INTVL_HRS)
+        raise Exception("Nomads script does not work!")
+
+        # get_nomads_data(NOMADS_file_type,EXPTDIR,USHdir,DATE_FIRST_CYCL,CYCL_HRS,FCST_LEN_HRS,LBC_SPEC_INTVL_HRS)
 
     #
     # -----------------------------------------------------------------------
@@ -959,10 +961,9 @@ def generate_FV3LAM_wflow(USHdir, logfile: str = 'log.generate_FV3LAM_wflow') ->
         ========================================================================
         """
     )
-    #
     # -----------------------------------------------------------------------
     #
-    # If rocoto is required, print instructions on how to load and use it
+    # If rocoto is required, print instructions on how to use it
     #
     # -----------------------------------------------------------------------
     #
@@ -1007,9 +1008,8 @@ def get_nomads_data(NOMADS_file_type,EXPTDIR,USHdir,DATE_FIRST_CYCL,CYCL_HRS,FCS
     print(f"NOMADS_file_type= {NOMADS_file_type}")
     cd_vrfy(EXPTDIR)
     NOMADS_script = os.path.join(USHdir, "NOMADS_get_extrn_mdl_files.sh")
-    # run_command(f"""{NOMADS_script} {date_to_str(DATE_FIRST_CYCL,format="%Y%m%d")} \
-    #                 {date_to_str(DATE_FIRST_CYCL,format="%H")} {NOMADS_file_type} {FCST_LEN_HRS} {LBC_SPEC_INTVL_HRS}""")
-    raise Exception("Nomads script does not work")
+    run_command(f"""{NOMADS_script} {date_to_str(DATE_FIRST_CYCL,format="%Y%m%d")} \
+                    {date_to_str(DATE_FIRST_CYCL,format="%H")} {NOMADS_file_type} {FCST_LEN_HRS} {LBC_SPEC_INTVL_HRS}""")
 
 def setup_logging(logfile: str = 'log.generate_FV3LAM_wflow') -> None:
     """
@@ -1036,11 +1036,10 @@ if __name__ == "__main__":
     try:
         generate_FV3LAM_wflow(USHdir, logfile)
     except:
-        # If the call to the generate_FV3LAM_wflow function above was not successful, 
-        # print out an error message and exit with a nonzero return code.
         logging.exception(dedent(
             f"""
             *********************************************************************
+            FATAL ERROR:
             Experiment generation failed. See the error message(s) printed below.
             For more detailed information, check the log file from the workflow
             generation script: {logfile}
--- a/ush/generate_FV3LAM_wflow.py
+++ b/ush/generate_FV3LAM_wflow.py
@@ -1047,8 +1047,8 @@ class Testing(unittest.TestCase):
     def test_generate_FV3LAM_wflow(self):
 
         # run workflows in separate process to avoid conflict between community and nco settings
-        def run_workflow():
-            p = Process(target=generate_FV3LAM_wflow)
+        def run_workflow(USHdir,logfile):
+            p = Process(target=generate_FV3LAM_wflow,args=(USHdir,logfile))
             p.start()
             p.join()
             exit_code = p.exitcode
@@ -1062,13 +1062,13 @@ class Testing(unittest.TestCase):
         # community test case
         cp_vrfy(f"{USHdir}/config.community.yaml", f"{USHdir}/config.yaml")
         run_command(f"""{SED} -i 's/MACHINE: hera/MACHINE: linux/g' {USHdir}/config.yaml""")
-        generate_FV3LAM_wflow(USHdir, logfile)
+        run_workflow(USHdir, logfile)
 
         # nco test case
         set_env_var("OPSROOT", f"{USHdir}/../../nco_dirs")
         cp_vrfy(f"{USHdir}/config.nco.yaml", f"{USHdir}/config.yaml")
         run_command(f"""{SED} -i 's/MACHINE: hera/MACHINE: linux/g' {USHdir}/config.yaml""")
-        generate_FV3LAM_wflow(USHdir, logfile)
+        run_workflow(USHdir, logfile)
 
     def setUp(self):
         define_macos_utilities()
--- a/ush/generate_FV3LAM_wflow.py
+++ b/ush/generate_FV3LAM_wflow.py
@@ -5,6 +5,7 @@ import sys
 import platform
 import subprocess
 import unittest
+import logging
 from multiprocessing import Process
 from textwrap import dedent
 from datetime import datetime, timedelta
@@ -61,12 +62,12 @@ except ImportError as error:
     python_error_handler()
 
 
-def generate_FV3LAM_wflow():
+def generate_FV3LAM_wflow(logfile: str = 'log.generate_FV3LAM_wflow') -> None:
     """Function to setup a forecast experiment and create a workflow
-    (according to the parameters specified in the config file
+    (according to the parameters specified in the config file)
 
     Args:
-        None
+        logfile (str): The name of the file where logging is written
     Returns:
         None
     """
@@ -90,14 +91,15 @@ def generate_FV3LAM_wflow():
     # check python version
     major, minor, patch = platform.python_version_tuple()
     if int(major) < 3 or int(minor) < 6:
-        print_info_msg(
+        logging.error(
             f"""
 
             Error: python version must be 3.6 or higher
             python version: {major}.{minor}"""
         )
+        raise
 
-    # define macros
+    # define utilities 
     define_macos_utilities()
 
     #
@@ -143,7 +145,7 @@ def generate_FV3LAM_wflow():
 
         template_xml_fp = os.path.join(PARMdir, WFLOW_XML_FN)
 
-        print_info_msg(
+        logging.info(
             f'''
             Creating rocoto workflow XML file (WFLOW_XML_FP) from jinja template XML
             file (template_xml_fp):
@@ -438,7 +440,7 @@ def generate_FV3LAM_wflow():
         # End of "settings" variable.
         settings_str = cfg_to_yaml_str(settings)
 
-        print_info_msg(
+        logging.info(
             dedent(
                 f"""
                 The variable \"settings\" specifying values of the rococo XML variables
@@ -447,7 +449,6 @@ def generate_FV3LAM_wflow():
                 settings =\n\n"""
             )
             + settings_str,
-            verbose=VERBOSE,
         )
 
         #
@@ -459,7 +460,7 @@ def generate_FV3LAM_wflow():
                 ["-q", "-u", settings_str, "-t", template_xml_fp, "-o", WFLOW_XML_FP]
             )
         except:
-            print_err_msg_exit(
+            logging.exception(
                 dedent(
                     f"""
                 Call to python script fill_jinja_template.py to create a rocoto workflow
@@ -482,7 +483,7 @@ def generate_FV3LAM_wflow():
     #
     # -----------------------------------------------------------------------
     #
-    print_info_msg(
+    logging.info(
         f'''
         Creating symlink in the experiment directory (EXPTDIR) that points to the
         workflow launch script (WFLOW_LAUNCH_SCRIPT_FP):
@@ -521,12 +522,11 @@ def generate_FV3LAM_wflow():
     #
     if SYMLINK_FIX_FILES:
 
-        print_info_msg(
+        logging.info(
             f'''
             Symlinking fixed files from system directory (FIXgsm) to a subdirectory (FIXam):
               FIXgsm = \"{FIXgsm}\"
               FIXam = \"{FIXam}\"''',
-            verbose=VERBOSE,
         )
 
         ln_vrfy(f'''-fsn "{FIXgsm}" "{FIXam}"''')
@@ -535,12 +535,11 @@ def generate_FV3LAM_wflow():
     #
     else:
 
-        print_info_msg(
+        logging.info(
             f'''
             Copying fixed files from system directory (FIXgsm) to a subdirectory (FIXam):
               FIXgsm = \"{FIXgsm}\"
               FIXam = \"{FIXam}\"''',
-            verbose=VERBOSE,
         )
 
         check_for_preexist_dir_file(FIXam, "delete")
@@ -559,14 +558,13 @@ def generate_FV3LAM_wflow():
     # -----------------------------------------------------------------------
     #
     if USE_MERRA_CLIMO:
-        print_info_msg(
+        logging.info(
             f'''
             Copying MERRA2 aerosol climatology data files from system directory
             (FIXaer/FIXlut) to a subdirectory (FIXclim) in the experiment directory:
               FIXaer = \"{FIXaer}\"
               FIXlut = \"{FIXlut}\"
               FIXclim = \"{FIXclim}\"''',
-            verbose=VERBOSE,
         )
 
         check_for_preexist_dir_file(FIXclim, "delete")
@@ -585,30 +583,26 @@ def generate_FV3LAM_wflow():
     #
     # -----------------------------------------------------------------------
     #
-    print_info_msg(
+    logging.info(
         f"""
         Copying templates of various input files to the experiment directory...""",
-        verbose=VERBOSE,
     )
 
-    print_info_msg(
+    logging.info(
         f"""
         Copying the template data table file to the experiment directory...""",
-        verbose=VERBOSE,
     )
     cp_vrfy(DATA_TABLE_TMPL_FP, DATA_TABLE_FP)
 
-    print_info_msg(
+    logging.info(
         f"""
         Copying the template field table file to the experiment directory...""",
-        verbose=VERBOSE,
     )
     cp_vrfy(FIELD_TABLE_TMPL_FP, FIELD_TABLE_FP)
 
-    print_info_msg(
+    logging.info(
         f"""
         Copying the template NEMS configuration file to the experiment directory...""",
-        verbose=VERBOSE,
     )
     cp_vrfy(NEMS_CONFIG_TMPL_FP, NEMS_CONFIG_FP)
     #
@@ -616,11 +610,10 @@ def generate_FV3LAM_wflow():
     # clone of the FV3 code repository to the experiment directory (EXPT-
     # DIR).
     #
-    print_info_msg(
+    logging.info(
         f"""
         Copying the CCPP physics suite definition XML file from its location in
         the forecast model directory sturcture to the experiment directory...""",
-        verbose=VERBOSE,
     )
     cp_vrfy(CCPP_PHYS_SUITE_IN_CCPP_FP, CCPP_PHYS_SUITE_FP)
     #
@@ -628,11 +621,10 @@ def generate_FV3LAM_wflow():
     # clone of the FV3 code repository to the experiment directory (EXPT-
     # DIR).
     #
-    print_info_msg(
+    logging.info(
         f"""
         Copying the field dictionary file from its location in the forecast
         model directory sturcture to the experiment directory...""",
-        verbose=VERBOSE,
     )
     cp_vrfy(FIELD_DICT_IN_UWM_FP, FIELD_DICT_FP)
     #
@@ -642,7 +634,7 @@ def generate_FV3LAM_wflow():
     #
     # -----------------------------------------------------------------------
     #
-    print_info_msg(
+    logging.info(
         f'''
         Setting parameters in weather model's namelist file (FV3_NML_FP):
         FV3_NML_FP = \"{FV3_NML_FP}\"'''
@@ -877,7 +869,7 @@ def generate_FV3LAM_wflow():
 
     settings_str = cfg_to_yaml_str(settings)
 
-    print_info_msg(
+    logging.info(
         dedent(
             f"""
             The variable \"settings\" specifying values of the weather model's
@@ -886,7 +878,6 @@ def generate_FV3LAM_wflow():
             settings =\n\n"""
         )
         + settings_str,
-        verbose=VERBOSE,
     )
     #
     # -----------------------------------------------------------------------
@@ -917,7 +908,7 @@ def generate_FV3LAM_wflow():
             ]
         )
     except:
-        print_err_msg_exit(
+        logging.exception(
             dedent(
                 f"""
             Call to python script set_namelist.py to generate an FV3 namelist file
@@ -975,7 +966,7 @@ def generate_FV3LAM_wflow():
         rocotorun_cmd = f"rocotorun -w {WFLOW_XML_FN} -d {wflow_db_fn} -v 10"
         rocotostat_cmd = f"rocotostat -w {WFLOW_XML_FN} -d {wflow_db_fn} -v 10"
 
-    print_info_msg(
+    logging.info(
         f"""
         ========================================================================
         ========================================================================
@@ -997,7 +988,7 @@ def generate_FV3LAM_wflow():
     #
     if WORKFLOW_MANAGER == "rocoto":
 
-        print_info_msg(
+        logging.info(
             f"""
             To launch the workflow, first ensure that you have a compatible version
             of rocoto available. For most pre-configured platforms, rocoto can be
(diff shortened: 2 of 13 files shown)


danielabdi-noaa:
@mkavulich The unittest is failing probably due to a mixup of the community and nco test cases settings. I recall that when i first called `generate_workflow` directly,  it remembered the `RUN_ENVIR` setting from the first test and used it for the latter. So I was forced to run them in separate processes.

mkavulich:
@danielabdi-noaa I restored the multiprocessing to the unit tests in generate_FV3LAM_wflow.py, but I am still seeing errors that I don't understand (I can't figure out how to see the full logs). I am on PTO Monday and Tuesday but I may ask for a quick meeting after that to try to figure out what I did wrong.

danielabdi-noaa:
@mkavulich You can run the unittest locally 
```
python3 -m unittest generate_FV3LAM_wflow.py
```
It looks like you are just missing a `dedent` import in `get_crontab_contents.py`
```Process Process-2:
  File "/contrib/miniconda3/4.5.12/envs/regional_workflow/lib/python3.8/multiprocessing/process.py", line 313, in _bootstrap
    self.run()
  File "/contrib/miniconda3/4.5.12/envs/regional_workflow/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/scratch2/BMC/gsd-hpcs/Daniel.Abdi/ufs-srweather-app/ush/generate_FV3LAM_wflow.py", line 494, in generate_FV3LAM_wflow
    add_crontab_line()
  File "/scratch2/BMC/gsd-hpcs/Daniel.Abdi/ufs-srweather-app/ush/get_crontab_contents.py", line 104, in add_crontab_line
    logger.info(dedent(
NameError: name 'dedent' is not defined
```

danielabdi-noaa:
By the way, is it possible to override how the logger prints messages so that it automatically dedents them just like `print_info_msg` does? That way we do not have to dedent every message going to the log file or the screen. There is a case where you have to dedent the message beforehand, e.g. when concatenating two strings, but in most cases it is convenient to just send the string as is.

mkavulich:
@danielabdi-noaa Thanks for the info, the tests seem to be working now!

> By the way, is it possible to override how the logger prints messages so that it automatically dedents them just like `print_info_msg` does? That way we do not have to dedent every message going to the log file or the screen. There is a case where you have to dedent the message beforehand, e.g. when concatenating two strings, but in most cases it is convenient to just send the string as is.

There's no built-in way to do this as far as I can tell, but I could add a new function that does this. I didn't want to change the existing `print_info_msg()` or `print_err_msg_exit()` because I wanted to keep these initial improvements confined to the context of workflow generation and the existing functions are used throughout the system in several different contexts. I will put that in and then open this PR for review.

danielabdi-noaa:
@mkavulich Are you thinking of modifying the logger.info/debug behaviour to include dedentation automatically or a separate function? As you pointed out, there is no built-in way but there seem to be workarounds like [this](https://stackoverflow.com/a/66638269/1382302) that will automatically dedent both console and logfile output. I tried it quickly on your PR with this patch, and it looks like it does the job.

mkavulich:
@danielabdi-noaa We could discuss a better strategy in the future but I don't want to "force" dedenting on all log messages, so I have implemented it as a separate function similar to `print_info_msg()` for this initial effort.

Thanks for all your comments here so far, I am opening this up for wider review but let me know if you have additional comments/suggestions.

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1085525584/20221020000517/ufs-srweather-app
Build was Successful
Rocoto jobs started
Long term tracking will be done on 9 experiments
ci-hera-intel-WE
Experiment Failed on hera: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
2022-10-20 00:36:11 +0000 :: hfe05 :: Task make_orog, jobid=36829776, in state DEAD (FAILED), ran for 39.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR
2022-10-20 00:36:13 +0000 :: hfe05 :: Task make_orog, jobid=36829774, in state DEAD (FAILED), ran for 51.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_RAP_suite_HRRR
2022-10-20 00:36:10 +0000 :: hfe11 :: Task make_orog, jobid=36829772, in state DEAD (FAILED), ran for 55.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_RRFS_v1beta
2022-10-20 00:36:06 +0000 :: hfe05 :: Task make_orog, jobid=36829779, in state DEAD (FAILED), ran for 38.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
2022-10-20 00:36:14 +0000 :: hfe12 :: Task make_orog, jobid=36829780, in state DEAD (FAILED), ran for 41.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_HRRR
2022-10-20 00:36:10 +0000 :: hfe08 :: Task make_orog, jobid=36829770, in state DEAD (FAILED), ran for 53.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v15p2
2022-10-20 00:36:12 +0000 :: hfe08 :: Task make_orog, jobid=36829778, in state DEAD (FAILED), ran for 40.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
2022-10-20 00:36:09 +0000 :: hfe07 :: Task make_orog, jobid=36829775, in state DEAD (FAILED), ran for 43.0 seconds, exit status=256, try=2 (of 2)
Experiment Succeeded on hera: nco_grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR
All experiments completed

venitahagerty:
Machine: jet
Compiler: intel
Job: WE
Repo location: /lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1085525584/20221020000511/ufs-srweather-app
Build was Successful
Rocoto jobs started
Long term tracking will be done on 9 experiments
ci-jet-intel-WE
Experiment Succeeded on jet: nco_grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR
Experiment Succeeded on jet: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
Experiment Succeeded on jet: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR
Experiment Succeeded on jet: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_RAP_suite_HRRR
Experiment Succeeded on jet: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
Experiment Succeeded on jet: grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v15p2
Experiment Succeeded on jet: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
Experiment Succeeded on jet: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_HRRR
Experiment Succeeded on jet: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_RRFS_v1beta

venitahagerty:
Machine: jet
Compiler: intel
Job: WE
Repo location: /lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1085525584/20221020022018/ufs-srweather-app
Build was Successful
Rocoto jobs started
Long term tracking will be done on 9 experiments
ci-jet-intel-WE
Experiment Succeeded on jet: nco_grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR
Experiment Succeeded on jet: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_RAP_suite_HRRR
Experiment Succeeded on jet: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
Experiment Succeeded on jet: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR
Experiment Succeeded on jet: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
Experiment Succeeded on jet: grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v15p2
Experiment Succeeded on jet: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
Experiment Succeeded on jet: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_HRRR
Experiment Succeeded on jet: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_RRFS_v1beta

ush/generate_FV3LAM_wflow.py
```diff
@@ -950,6 +916,11 @@ def generate_FV3LAM_wflow():
     if not RUN_TASK_MAKE_GRID:
 
         set_FV3nml_sfc_climo_filenames()
+
+    # Call script to get NOMADS data
+    if NOMADS:
+        get_nomads_data(NOMADS_file_type,EXPTDIR,USHdir,DATE_FIRST_CYCL,CYCL_HRS,FCST_LEN_HRS,LBC_SPEC_INTVL_HRS)
```
christinaholtNOAA: Consider putting each of these arguments on separate lines, also consider making them named arguments. This is quite a few arguments to manage as a single-line list.

ush/generate_FV3LAM_wflow.py
```diff
@@ -950,6 +916,11 @@ def generate_FV3LAM_wflow():
     if not RUN_TASK_MAKE_GRID:
 
         set_FV3nml_sfc_climo_filenames()
+
+    # Call script to get NOMADS data
```
christinaholtNOAA: script -> function

ush/generate_FV3LAM_wflow.py
```diff
@@ -1043,146 +998,82 @@ def generate_FV3LAM_wflow():
...
-        run_command(
-            f"""{NOMADS_script} {date_to_str(DATE_FIRST_CYCL,format="%Y%m%d")} \
-                      {date_to_str(DATE_FIRST_CYCL,format="%H")} {NOMADS_file_type} {FCST_LEN_HRS} {LBC_SPEC_INTVL_HRS}"""
-        )
 
+    # If we got to this point everything was successful: move the log file to the experiment directory.
+    mv_vrfy(logfile, EXPTDIR)
+
+def get_nomads_data(NOMADS_file_type,EXPTDIR,USHdir,DATE_FIRST_CYCL,CYCL_HRS,FCST_LEN_HRS,LBC_SPEC_INTVL_HRS):
+    print("Getting NOMADS online data")
```
christinaholtNOAA, line 1007: A doc string would be useful here.

ush/generate_FV3LAM_wflow.py
```diff
@@ -1043,146 +998,82 @@ def generate_FV3LAM_wflow():
...
+    print(f"NOMADS_file_type= {NOMADS_file_type}")
+    cd_vrfy(EXPTDIR)
+    NOMADS_script = os.path.join(USHdir, "NOMADS_get_extrn_mdl_files.sh")
+    # run_command(f"""{NOMADS_script} {date_to_str(DATE_FIRST_CYCL,format="%Y%m%d")} \
+    #                 {date_to_str(DATE_FIRST_CYCL,format="%H")} {NOMADS_file_type} {FCST_LEN_HRS} {LBC_SPEC_INTVL_HRS}""")
+    raise Exception("Nomads script does not work")
```
christinaholtNOAA: Hahahaha! Maybe giving some feedback earlier than this would be useful!

ush/generate_FV3LAM_wflow.py
```diff
@@ -1043,146 +998,82 @@ def generate_FV3LAM_wflow():
...
+    Sets up logging, printing high-priority (INFO and higher) messages to screen, and printing all
+    messages with detailed timing and routine info in the specified text file.
+    """
+    logging.basicConfig(level=logging.DEBUG,
+                        format='%(name)-22s %(levelname)-8s %(message)s',
+                        filename=logfile,
+                        filemode='w')
+    logging.debug(f'Finished setting up debug file logging in {logfile}')
+    console = logging.StreamHandler()
+    console.setLevel(logging.INFO)
```
christinaholtNOAA, line 1025: Does this mean we won't see errors on the screen?

ush/generate_FV3LAM_wflow.py
```diff
@@ -1043,146 +998,82 @@ def generate_FV3LAM_wflow():
...
-    # code.
-    #
-    else:
-        print_err_msg_exit(
+    # experiment/workflow.
+    try:
+        generate_FV3LAM_wflow(USHdir, logfile)
+    except:
+        # If the call to the generate_FV3LAM_wflow function above was not successful, 
+        # print out an error message and exit with a nonzero return code.
```
christinaholtNOAA: Is this comment necessary? Seems duplicative and obvious given the code.

ush/python_utils/print_msg.py
```diff
@@ -39,3 +40,25 @@ def print_info_msg(info_msg, verbose=True):
         print(dedent(info_msg))
         return True
     return False
+
+def log_info(info_msg, verbose=True, ddent=True):
```
christinaholtNOAA: You might consider `dedent_` as this variable name. It's a fairly common approach to dealing with namespace issues that doesn't make you wonder if it was misspelled, and is clear that you are avoiding a namespace collision.

ush/python_utils/print_msg.py
```diff
@@ -39,3 +40,25 @@ def print_info_msg(info_msg, verbose=True):
...
+
+    Args:
+        info_msg : info message to print
+        verbose : set to False to silence printing
+        ddent : set to False to disable "dedenting" (print string as-is)
+    Returns:
+        None
+    """
+
+    # "sys._getframe().f_back.f_code.co_name" returns the name of the calling function
```
christinaholtNOAA, line 56: Consider removing commented lines of code.

ush/set_ozone_param.py
```diff
@@ -92,12 +90,11 @@ def set_ozone_param(ccpp_phys_suite_fp):
         fixgsm_ozone_fn = "global_o3prdlos.f77"
         ozone_param = "ozphys"
     else:
-        print_err_msg_exit(
+        raise Exception(
```
christinaholtNOAA: Does KeyError make sense here?

ush/setup.py
```diff
@@ -84,30 +85,38 @@ def setup():
...
-    # specified configuration file is not tracked by the repository, whereas
-    # the default configuration file is tracked.
-    #
-    # -----------------------------------------------------------------------
-    #
-    if os.path.exists(EXPT_CONFIG_FN):
+
+    # Load the user config file, containing contains user-specified values 
+    # variables that override their default values. Ensure all user-specified
+    # variables correspond to a default value.
```
christinaholtNOAA, line 93: This comment could use some work.

ush/setup.py
```diff
@@ -84,30 +85,38 @@ def setup():
...
-            f'''
-            User config file not found
-              EXPT_CONFIG_FN = \"{EXPT_CONFIG_FN}\"'''
-        )
+    except:
+        errmsg = dedent(f'''\n
+                        Could not load YAML config file:  {EXPT_CONFIG_FN}
+                        The file may be formatted incorrectly; reference the Users Guide for more info 
+                        ''')
+        raise Exception(errmsg) from None
```
christinaholtNOAA: This is going to cause some problems for debugging if you can't see the original exception raised when loading the config file fails. It would be useful to catch specific failures here and raise them. 

Perhaps the file is formatted correctly but the user doesn't have appropriate permissions? Or a whole host of other things?

ush/setup.py
```diff
@@ -84,30 +85,38 @@ def setup():
...
+
+    cfg_u = flatten_dict(cfg_u)
+    for key in cfg_u:
+        if key not in flatten_dict(cfg_d):
+            raise Exception(dedent(f'''
+                            User-specified variable "{key}" in {EXPT_CONFIG_FN} is not valid
```
christinaholtNOAA: Suggest adding a period at the end of the line.

ush/setup.py
```diff
@@ -84,30 +85,38 @@ def setup():
...
+
+    # Mandatory variables *must* be set in the user's config; the default value is invalid
+    mandatory = ['MACHINE']
+    for val in mandatory:
+        if val not in cfg_u:
+            raise Exception(f'Mandatory variable "{val}" not found in user config file {EXPT_CONFIG_FN}')
+
+    import_vars(dictionary=cfg_u, env_vars=["MACHINE",
+                   "EXTRN_MDL_NAME_ICS", "EXTRN_MDL_NAME_LBCS",
+                   "FV3GFS_FILE_FMT_ICS", "FV3GFS_FILE_FMT_LBCS"])
```
christinaholtNOAA, line 121: This list would be helpful with one item per line.

ush/setup.py
```diff
@@ -402,26 +425,17 @@ def get_location(xcs,fmt):
...
-              FIXgsm = \"{FIXgsm or ""}
-              FIXaer = \"{FIXaer or ""}
-              FIXlut = \"{FIXlut or ""}
-              TOPO_DIR = \"{TOPO_DIR or ""}
-              SFC_CLIMO_INPUT_DIR = \"{SFC_CLIMO_INPUT_DIR or ""}
-              DOMAIN_PREGEN_BASEDIR = \"{DOMAIN_PREGEN_BASEDIR or ""}
-            You can specify the missing location(s) in config.sh"""
-        )
+    # Mandatory variables *must* be set in the user's config or the machine file; the default value is invalid
+    mandatory = ['NCORES_PER_NODE', 'FIXgsm', 'FIXaer', 'FIXlut', 'TOPO_DIR', 'SFC_CLIMO_INPUT_DIR']
```
christinaholtNOAA, line 431: This is a long enough list that one item per line is probably easier to manage.

ush/setup.py
```diff
@@ -402,26 +425,17 @@ def get_location(xcs,fmt):
...
+    for val in mandatory:
+        # globals() returns dictionary of global variables
+        if not globals()[val]:
```
christinaholtNOAA: It is more efficient to call globals once and assign it to a variable outside this loop.

ush/setup.py
```diff
@@ -505,123 +512,62 @@ def get_location(xcs,fmt):
...
             The coupling flag CPL has not been specified for this value of FCST_MODEL:
               FCST_MODEL = \"{FCST_MODEL}\"'''
         )
-    #
-    # -----------------------------------------------------------------------
-    #
-    # Make sure RESTART_INTERVAL is set to an integer value if present
-    #
-    # -----------------------------------------------------------------------
-    #
```
christinaholtNOAA, line 517: Any chance you'd be willing to make this sort of change throughout this file? The 7 lines to 1 line is clearer.

ush/setup.py
```diff
@@ -505,123 +512,62 @@ def get_location(xcs,fmt):
...
     # If using a custom post configuration file, make sure that it exists.
-    #
-    # -----------------------------------------------------------------------
-    #
     if USE_CUSTOM_POST_CONFIG_FILE:
-        if not os.path.exists(CUSTOM_POST_CONFIG_FP):
-            print_err_msg_exit(
+        try:
+            if not os.path.exists(CUSTOM_POST_CONFIG_FP):
+                raise
```
christinaholtNOAA, line 547: What are you raising here? Why does it need to be in a try block?

christinaholtNOAA (commented):
Nice!

tests/WE2E/run_WE2E_tests.sh
```diff
@@ -65,6 +65,63 @@ WE2Edir="$TESTSdir/WE2E"
...
+  Errors found: check your python environment
+
+  Instructions for setting up python environments can be found on the web:
+  https://github.com/ufs-community/ufs-srweather-app/wiki/Getting-Started
+
+"
+fi
+#
+#-----------------------------------------------------------------------
+#
```
danielabdi-noaa, line 124: `generate_workflow.py` does these checks so I wonder if we can somehow do without this piece of code to avoid duplication? However, I do understand the idea is to catch errors as early as possible! It just looks odd to me that we are checking python packages from bash, so maybe putting them in a separate python script and calling that will work?

ush/generate_FV3LAM_wflow.py
```diff
@@ -459,7 +445,7 @@ def generate_FV3LAM_wflow():
                 ["-q", "-u", settings_str, "-t", template_xml_fp, "-o", WFLOW_XML_FP]
             )
         except:
-            print_err_msg_exit(
+            logging.exception(
```
danielabdi-noaa, line 448: There is this NCO requirement that "ERROR:" be printed so just want to make sure the exception mechanism do that. The error messages look good to me but you may want to check with NCO folks. @MatthewPyle-NOAA

ush/generate_FV3LAM_wflow.py
```diff
@@ -1043,146 +998,82 @@ def generate_FV3LAM_wflow():
...
-    #
-    # Read in experiment/workflow variables needed later below from the tem-
-    # porary file created in the subshell above containing the call to the
-    # generate_FV3LAM_wflow function.  These variables are not directly
-    # available here because the call to generate_FV3LAM_wflow above takes
-    # place in a subshell (due to the fact that we are then piping its out-
-    # put to the "tee" command).  Then remove the temporary file.
-    #
-    (_, exptdir, _) = run_command(f'''sed "1q;d" "{tmp_fp}"''')
-    (_, retval, _) = run_command(f''' sed "2q;d" "{tmp_fp}"''')
```
danielabdi-noaa, line 1131: I am glad this most un-pythonic piece of code is gone! There are still a couple of `run_commands` left in the crontab management code, which hopefully can be replaced with a python crontab management package instead.

ush/python_utils/print_msg.py
```diff
@@ -39,3 +40,25 @@ def print_info_msg(info_msg, verbose=True):
...
+    logger=getLogger(sys._getframe().f_back.f_code.co_name)
+
+    if verbose:
+        if ddent:
+            logger.info(indent(dedent(info_msg), '  '))
+        else:
+            logger.info(info_msg)
+
```
danielabdi-noaa, line 64: Will there be a log_debug counterpart for logging.debug?

ush/setup.py
danielabdi-noaa: This is my fault. I often let the linter/formatter take care of it and it would be good to have that or something similar as a precommit script.

danielabdi-noaa (commented):
@mkavulich Thanks for your great work! This PR is a nice demonstration of the power of python workflow compared to shell workflow. I have left a few minor suggestions and will approve once I am done with testing the error reporting cases you listed.

ush/generate_FV3LAM_wflow.py
MatthewPyle-NOAA: @danielabdi-noaa  From the implementation standards document:

> The context of the error must be communicated as descriptively as possible and prefaced with “WARNING:” or “FATAL ERROR:”


But honestly for this workflow generating functionality that won't run routinely in operations, it probably isn't a huge concern

ush/python_utils/print_msg.py
mkavulich: Yes, I plan on introducing "debug" functionality later; right now it's in a weird "in-between" where VERBOSE and DEBUG are specified in the config file, but I'd like to do away with that and make it a command-line argument. But that will require a lot more discussion so I didn't want to make that part of this PR (especially since it's a "minor release" that shouldn't make major changes to how the workflow is set up and run). Now that you mention it, I will open up an issue for that so it gets tracked.

tests/WE2E/run_WE2E_tests.sh
mkavulich: @danielabdi-noaa This script calls get_WE2Etest_names_subdirs_descs.sh which calls $USHdir/calculate_cost.py, which at the very least needs pyyaml (and maybe more, I haven't checked further than that). So we need python checks at the beginning of this script to check for a graceful exit.

Hopefully soon this will be python-based end-to-end so these checks can be neater.

ush/generate_FV3LAM_wflow.py
mkavulich: I think this is a good idea actually, even if it is not technically needed. I will add a "FATAL ERROR:" line to the beginning of the catch-all error message, for extra "oomph" to catch the user's attention.

ush/python_utils/print_msg.py
mkavulich: This isn't code, it's a comment clarifying the use of obscure python functionality

ush/setup.py
mkavulich: I have returned the stack trace for this particular error. It is very ugly however, and I would like to find a cleaner solution in the future

ush/setup.py
mkavulich: That seems like a way to blow up a little code into a lot of lines. Is this recommended by a standard or just personal preference?

ush/setup.py
mkavulich: I feel that's outside the scope of this PR, the whole workflow needs a comment trim

ush/setup.py
mkavulich: Because I'd like to check for the existence of this file and error out appropriately here in this script, even though we don't need to take any action on this file until later in the workflow. Do you know of a more elegant way to do this?

ush/python_utils/print_msg.py
christinaholtNOAA: Oops!

ush/generate_FV3LAM_wflow.py
mkavulich: No, all messages of "INFO" level or higher are printed to screen; this includes ERROR and WARNING messages as well.  https://docs.python.org/3/library/logging.html#levels

I hope to later do a sanitation of the workflow to demote some messages to "DEBUG" level, and have those only print to the log file unless the user specifies otherwise. That is for future work though.

christinaholtNOAA: Thanks for the clarification!

mkavulich:
@christinaholtNOAA  I believe I have addressed most of your concerns and followed up on the rest with a few comments/questions. Let me know what you think.

ush/setup.py
christinaholtNOAA: It's not just a personal preference. It's easier to maintain the list over time as we can clearly see when the list is changed in a variety of different `diff` tools. I'm cool with you not making the change here, but I will likely address it in an upcoming PR.

ush/setup.py
christinaholtNOAA: That's why I asked.

ush/setup.py
christinaholtNOAA: My first thought is to raise the exception inside the if block if the file doesn't exist. I know you'd mentioned the logger.exception needs to be used in a try block though. You don't seem to be using that here. Just curious.

venitahagerty:
Machine: jet
Compiler: intel
Job: WE
Repo location: /lfs1/BMC/nrtrr/rrfs_ci/autoci/pr/1085525584/20221021213518/ufs-srweather-app
ci-jet-intel-WE

ush/setup.py
```diff
@@ -400,26 +427,18 @@ def get_location(xcs,fmt):
...
-              TOPO_DIR = \"{TOPO_DIR or ""}
-              SFC_CLIMO_INPUT_DIR = \"{SFC_CLIMO_INPUT_DIR or ""}
-              DOMAIN_PREGEN_BASEDIR = \"{DOMAIN_PREGEN_BASEDIR or ""}
-            You can specify the missing location(s) in config.sh"""
-        )
+    # Mandatory variables *must* be set in the user's config or the machine file; the default value is invalid
+    mandatory = ['NCORES_PER_NODE', 'FIXgsm', 'FIXaer', 'FIXlut', 'TOPO_DIR', 'SFC_CLIMO_INPUT_DIR']
+    globalvars = globals()
+    for val in mandatory:
+        # globals() returns dictionary of global variables
```
christinaholtNOAA, line 434: Maybe move this comment up and out of the loop with the call to `globals()`?

christinaholtNOAA (approved):
Everything looks good to me. The last two items (a comment below and an unresolved previous comment) are purely cosmetic.

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1085525584/20221021213515/ufs-srweather-app
Build was Successful
Rocoto jobs started
Long term tracking will be done on 9 experiments
ci-hera-intel-WE
Experiment Succeeded on hera: nco_grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR

ush/setup.py
mkavulich: Ah, I see the issue. os.path.exists returns an exception if passed "None", so this structure allows us to catch all errors in one. I added a clarifying comment for future developers.

mkavulich:
@danielabdi-noaa I think you still have one un-resolved comment as well, let me know if everything looks good!

danielabdi-noaa:
I've now run almost all error testing cases listed in the PR description and they work as expected so approving. 

Only the following case looks like it is impossible to catch on Hera atleast. If we run a WE2E test case though without conda activation, it would have been probably caught!
- Prior to loading regional_workflow conda environment

This one is disabled to get unittests working and i think for now it should be that way in my opinion. I've had to disable a couple of hard exits to get unittests working before, but if the benefits outweight the cons we could get rid of the unittests.
- Prior to building code

danielabdi-noaa (approved):

mkavulich:
@danielabdi-noaa Agreed on both counts. It would be nice to implement a check that the code is built, but I agree we will need to think of a better way to sync that with automated tests.

PR #419, opened 2022-10-14
[develop] Updated modulefiles for Cheyenne, Hera, Jet, Gaea, Orion 

natalie-perlin:
## DESCRIPTION OF CHANGES: 

Updated modulefiles for the following HPC systems: **Cheyenne**, **Gaea**, **Hera**, **Orion**, **Jet**, in particular, for the SRW 2.1 release. The changes include new paths for HPC-stack and miniconda3 installations in EPIC-managed common space. Met and Metplus verification packages were installed as part of the hpc-stack. Rocoto has also been installed on Cheyenne in  common EPIC-manages space. Met and Metplus modules are not loaded explicitly as modules, but the installation paths are specified in ./ush/machine/<_platform_>.yaml files. The packages needed for plotting routines are included in the **regional_workflow** conda virtual environment.

The following types of modulefiles were updated:

./modulefiles/build_<_platform_>_<compiler>
./modulefiles/wflow_<_platform_>
./modulefiles/tasks/<_platform_>/miniconda_regional_workflow
./modulefiles/tasks/<_platform_>/make_grid.local, make_ics.local, make_lbcs.local, make_orog.local, get_extrn_lbcs.local, get_extrn_ics.local, run_fcst.local, run_vx.local
./ush/machine/<_platform_>.yaml

where <_platform_> is **Cheyenne**, **Jet**, **Hera**, **Orion**, **Gaea**, 
and compiler either <_intel_> or <_gnu_>; Cheyenne has both compilers

A list of the recent hpc-stack builds updated recently for the current PR (plus additional compiler-mpi combinations for Hera) could be found in the following open issue in the [UFS-WM/Issues-1465:](https://github.com/ufs-community/ufs-weather-model/issues/1465).

### Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## TESTS CONDUCTED: 

- [x] hera.intel - building SRW, running tests via Jenkins, basic plots via Jenkins
- [x] orion.intel - building SRW, running tests via Jenkins, basic plots via Jenkins
- [x] cheyenne.intel - building SRW, running tests via Jenkins
- [x] cheyenne.gnu - building SRW on
- [x] gaea.intel  - building SRW, running tests via Jenkins, basic plots via Jenkins
- [x] jet.intel -  building SRW, running tests via Jenkins, basic plots via Jenkins
- [ ] wcoss2.intel
- [ ] NOAA Cloud (indicate which platform)
- [x] Jenkins
[...]

PR #404, opened 2022-10-07
Remove ENV_INIT_SCRIPT and avoid dependency on system python

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 

This PR addresses issues #401 and #402. The latter is addressed through the alternative solution described there.

Detailed changes:
- removes `ENV_INIT_SCRIPT_FP` and apply the sourcing of `/etc/profile` directly in `etc/lmod-setup.sh`. There will only be one way to setup Lmod going forward.
- Adds a central place to load worflow modules `ush/load_modules_wflow.sh` that is sourced by CIs run scripts
- Activate conda on all systems besides Cheyenne so that we can use the python version in conda for crontab deletion.

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## TESTS CONDUCTED: 
Run on Hera and Jet a test case successfully, but the most critical test is Gaea that does additional Lmod sourcing through scripts passed by `ENV_INIT_SCRIPT_FP`

- [X] hera.intel
- [ ] orion.intel
- [ ] cheyenne.intel
- [ ] cheyenne.gnu
- [ ] gaea.intel
- [X] jet.intel
- [ ] wcoss2.intel
- [ ] NOAA Cloud (indicate which platform)
- [ ] Jenkins
- [X] fundamental test suite
- [ ] comprehensive tests (specify *which* if a subset was used)

## DOCUMENTATION:

None

## ISSUE: 

- #401 
- #402 

## CHECKLIST

- [X] My code follows the style guidelines in the Contributor's Guide
- [X] I have performed a self-review of my own code using the Code Reviewer's Guide
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes need updates to the documentation. I have made corresponding changes to the documentation
- [ ] My changes do not require updates to the documentation (explain).
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] Any dependent changes have been merged and published

## LABELS (optional): 

A Code Manager needs to add the following labels to this PR: 
- [X] Work In Progress
- [ ] bug
- [ ] enhancement
- [ ] documentation
- [ ] release
- [ ] high priority
- [ ] run_ci
- [ ] run_we2e_fundamental_tests
- [ ] run_we2e_comprehensive_tests
[...]

PR #428, opened 2022-10-25
fix for conda on gaea

mark-a-potts:
## DESCRIPTION OF CHANGES: 
This fixes a glitch with conda on gaea. In order to get the correct path set for python after activating the conda environment, "conda deactivate" must be run first. 

### Type of change

## TESTS CONDUCTED: 

- [ ] hera.intel
- [ ] orion.intel
- [ ] cheyenne.intel
- [ ] cheyenne.gnu
- [x] gaea.intel
- [ ] jet.intel
- [ ] wcoss2.intel
- [ ] Jenkins
- [ ] fundamental test suite

## CHECKLIST

- [x] My changes need updates to the documentation. I have made corresponding changes to the documentation
- [x] Any dependent changes have been merged and published

## LABELS (optional): 

- [x] bug

### DIFF
--- a/ush/load_modules_run_task.sh
+++ b/ush/load_modules_run_task.sh
@@ -183,9 +183,10 @@ module list
 
 if [ -n "${SRW_ENV:-}" ] ; then
   set +u
-  conda activate ${SRW_ENV}
   if [ $machine = "gaea" ]; then
      conda deactivate
+     conda activate ${SRW_ENV}
+  else
      conda activate ${SRW_ENV}
   fi
   set -u
--- a/ush/load_modules_run_task.sh
+++ b/ush/load_modules_run_task.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -l
 
 #
 #-----------------------------------------------------------------------
@@ -183,12 +183,7 @@ module list
 
 if [ -n "${SRW_ENV:-}" ] ; then
   set +u
-  if [ $machine = "gaea" ]; then
-     conda deactivate
-     conda activate ${SRW_ENV}
-  else
-     conda activate ${SRW_ENV}
-  fi
+  conda activate ${SRW_ENV}
   set -u
 fi
 
--- a/modulefiles/tasks/gaea/get_extrn_ics.local.lua
+++ b/modulefiles/tasks/gaea/get_extrn_ics.local.lua
@@ -1,1 +1,0 @@
-load("miniconda_regional_workflow")
--- a/modulefiles/tasks/gaea/get_extrn_lbcs.local.lua
+++ b/modulefiles/tasks/gaea/get_extrn_lbcs.local.lua
@@ -1,1 +1,0 @@
-load("miniconda_regional_workflow")
--- a/modulefiles/tasks/gaea/make_grid.local.lua
+++ b/modulefiles/tasks/gaea/make_grid.local.lua
@@ -1,1 +1,0 @@
-load("miniconda_regional_workflow")
--- a/modulefiles/tasks/gaea/make_ics.local.lua
+++ b/modulefiles/tasks/gaea/make_ics.local.lua
@@ -1,1 +1,0 @@
-load("miniconda_regional_workflow")
--- a/modulefiles/tasks/gaea/make_lbcs.local.lua
+++ b/modulefiles/tasks/gaea/make_lbcs.local.lua
@@ -1,1 +1,0 @@
-load("miniconda_regional_workflow")
--- a/modulefiles/tasks/gaea/make_orog.local.lua
+++ b/modulefiles/tasks/gaea/make_orog.local.lua
@@ -1,1 +1,0 @@
-load("miniconda_regional_workflow")
--- a/modulefiles/tasks/gaea/miniconda_regional_workflow.lua
+++ b/modulefiles/tasks/gaea/miniconda_regional_workflow.lua
@@ -1,4 +1,0 @@
-prepend_path("MODULEPATH", "/lustre/f2/dev/role.epic/contrib/modulefiles")
-load(pathJoin("miniconda3", os.getenv("miniconda3_ver") or "4.12.0"))
-
-setenv("SRW_ENV", "regional_workflow")
--- a/modulefiles/tasks/gaea/run_fcst.local.lua
+++ b/modulefiles/tasks/gaea/run_fcst.local.lua
@@ -1,1 +1,0 @@
-load("miniconda_regional_workflow")
--- a/modulefiles/tasks/gaea/run_vx.local.lua
+++ b/modulefiles/tasks/gaea/run_vx.local.lua
@@ -1,2 +1,2 @@
-load("miniconda_regional_workflow")
+--load("miniconda_regional_workflow")
 prepend_path("LD_LIBRARY_PATH", "/ncrc/sw/gaea-cle7/python/3.9/anaconda-base/envs/noaa_py3.9/lib")


willmayfield (approved):

panll (approved):

ush/load_modules_run_task.sh
```diff
@@ -183,10 +183,11 @@ module list
 
 if [ -n "${SRW_ENV:-}" ] ; then
   set +u
-  conda activate ${SRW_ENV}
   if [ $machine = "gaea" ]; then
      conda deactivate
      conda activate ${SRW_ENV}
+  else
+     conda activate ${SRW_ENV}
```
danielabdi-noaa: We had a problem on Gaea and Jenkins, which is why this `activate->deactivate->activate` route is used specifically for Gaea. Please take a look at this discussion:
https://github.com/ufs-community/ufs-srweather-app/pull/404#issuecomment-1276609408
Maybe something has changed since moving to Lua, but pretty sure it was working with Jenkins as it was a couple of days ago.

mark-a-potts: Thanks, Daniel. The deactivate worked for me when I tried it without first activating an environment, but I am looking into this further now. I think that we may not need the deactivate at all, actually. The problem I was trying to fix on my end was because I had an old conda init stanza in my .bash_profile that was confusing the new conda. I will see if something similar is messing up the Jenkins job.

danielabdi-noaa: @mark-a-potts Maybe there is a way to make gaea consistent with other systems (one conda activate) once EPIC installs are used in PR #419 ? I also don't like the hack either.

MichaelLueken: Interestingly, the Jenkins CI [pipeline](https://jenkins-epic.woc.noaa.gov/blue/organizations/jenkins/ufs-srweather-app%2Fpipeline/detail/PR-428/1/pipeline/157/) for Gaea successfully ran through to completion for this PR.  If it was possible to remove the `conda deactivate` entirely, that would be great, but it might need to wait until after the updated miniconda PRs come in.

danielabdi-noaa: @MichaelLueken Indeed. I was under the impression that the `conda deactivate` will be a no-op before `conda activate $SRW` is executed, but it looks like it does deactivate the conda from the parent shell. I guess this is one less command for gaea but would be desirable to get rid of the `conda deactivate` as well.

mark-a-potts: I have tested a little more this afternoon, and I can't quite figure out what is going on. When I load wflow_gaea by hand, the python path is right and conda is known to be a function, but when I run a step with rocoto, the load_modules_and_run.sh script does not know that conda is a function and so the conda activate line does not work without a conda deactivate first. 

You are probably right @MichaelLueken, that we should test again once the miniconda PRs come through.

mark-a-potts: I did some more testing today and think I have found the problem, though I am not entirely sure *why* it is a problem. It seems that the extra module load for miniconda3 that happens in the modulefiles/tasks/gaea/*.lua files adds an extra level of conda stuff that messes things up. I removed all but the vx modulefile under gaea and everything seems to be working correctly without a need to deactivate the conda environment. I don't quite understand why the extra conda modules seem to be okay on other platforms, but not gaea, but I think this is a bit better of a solution than the deactivate workaround. I will re-open this and push the latest updates I have made.

danielabdi-noaa: It looks like this is happening on all systems now after transition to EPIC modules with @natalie-perlin 's PR?
https://github.com/ufs-community/ufs-srweather-app/pull/431#discussion_r1005025325

ush/load_modules_run_task.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -l
```
MichaelLueken, line 1: Does the bash script need to be login enabled in order for this work to function properly?  According to the code manager's guide, all bash scripts must explicitly be `#!/bin/bash` scripts (and not login enabled).

MichaelLueken (commented):
@mark-a-potts I have a question about the addition of the login enabling for `ush/load_modules_run_task.sh`.  Is making this script login enabled required?  This appears to go against the code manager's [guide](https://github.com/ufs-community/ufs-srweather-app/wiki/Code-Manager's-Guide#code) for shell scripting in the SRW.

ush/load_modules_run_task.sh
mark-a-potts: Oh, good catch! I don't think that is necessary. I added it to see if I could figure out what was going on. I will remove it on my side and run a quick test to be sure.

ush/load_modules_run_task.sh
danielabdi-noaa: Note that it is currently not activating conda in `load_modules_run_task.sh` because `SRW_ENV` is not set for it. So gaea is still an outlier compared to other Tier-1 systems. I wish there was a better solution that switches the python paths to `*/envs/regional_workflow/bin` instead of `*/bin` which is the root cause of the problem.

MichaelLueken: Since it appears as though @natalie-perlin's PR #431 is having the same issue for all machines, merging this PR would likely break what Natalie has done.  Would it be best to open a PR in Natalie's fork to include these changes in PR #431 and then ask her to make similar modifications to the local modulefiles for machines that aren't Gaea (since Mark has already done this for Gaea)?

MichaelLueken:
@mark-a-potts All of the Jenkins tests successfully ran and completed on all machines.  Now, we just need to figure out how to best get both these changes and @natalie-perlin's PR #431 in without adversely affecting each.

ulmononian:
@mark-a-potts @danielabdi-noaa @MichaelLueken : @EdwardSnyder-NOAA recently identified this same conda activate/deactivate issue with the workflow when we were testing things for an update to all the noaacloud module files (PR forthcoming from https://github.com/ulmononian/ufs-srweather-app/tree/feature/update_cloud_mods). we locally added an OR statement to include the noaacloud platform in this line https://github.com/ulmononian/ufs-srweather-app/blob/2f27a09f3a99cba6c582ebfd2607cf581f487847/ush/load_modules_run_task.sh#L187, but given the solution in this PR, i am wondering if we need to address the cloud module files in a similar way?

mark-a-potts:
Sorry, I have been stuck in meetings all afternoon. I will touch base with @natalie-perlin tomorrow and see if we can come up with a single solution.

danielabdi-noaa:
@ulmononian Good to know this behavior is present in noaacloud! I don't think removing the miniconda3 modulefile loading from the `modulefiles/tasks/machine` like this PR does is a good idea. I believe the reason it is working here is because it is relying on the `conda activate` from the parent shell. Also both CIs (jenkins & github actions) do the same  before launching tasks so the problem is masked away. I can reproduce this on Hera after removing miniconda3 modulefile loading from `modulefiles/tasks/hera/miniconda_regional_worfklow.lua` and running a test case
- with conda activated from command linek
```
(regional_workflow) Daniel.Abdi@hfe02
```
All tasks finish even though miniconda3 module is not loaded by any of them.
- conda not activated (prompt without regional_workflow)
```
Daniel.Abdi@hfe02
```
I encounter `ModuleImportError` on yaml simlar to what we saw on `gaea` before the fix.

So I don't think it is a good idea to remove miniconda3 from the task modulefiles. The philsophy behind task specific modulefiles is probably that each task loads the modules and packages it needs itself, not by the parent shell.
If we do want to rely on this behaviour, it is best to make the change to all task modulefiles -- it does simplify the code afterall
but we want to be sure we are not breaking something else.

MichaelLueken:
@mark-a-potts Since PR #444 has been merged, is it safe to close this PR now?

MichaelLueken:
Closing PR #428 since the necessary changes have already been merged in via PR #444.

Issue #462, opened 2022-11-10
Running comprehensive suite of WE2E tests overwhelms some platforms


mkavulich:
## Expected behavior
All comprehensive tests should run 
## Current behavior
The large number of cron tasks needing to be executed simultaneously (57 for the current comprehensive suite) as well as the large number of submitted jobs (several hundred) results in problems on certain platforms. The known issues are detailed in the next section.

### Cheyenne
The large number of cron tasks needing to be executed simultaneously (57 for the current comprehensive suite) strains the capability of the node(s) handling cron jobs on the Cheyenne machine. CISL has asked us specifically to limit our use of cron in this way.
### Hera
The large number of individual tests means that on platforms with a by-user job submission limit (300 on Hera) can be surpassed. While this does not create errors, attempting to submit additional jobs results in the new jobs being rejected, which rocoto counts as a failure and does not handle well.

## Steps To Reproduce
Run the comprehensive suite of WE2E tests on the platforms described above.

## Detailed Description of Fix
There are a few potential fixes for these issues individually, but the best fix would be to create a python script that will set up each test in the comprehensive suite, generate all of them, and track and manage the active jobs using rocoto rather than creating a crontab entry for each test. This python script would be intended to run continuously until all WE2E tasks are complete, and could be run via the command line or submitted to a compute node. By functioning this way, we will bypass the cron functionality to avoid those issues, and it will be relatively easy to implement any necessary throttling to avoid issues related to the number of submitted jobs.

## Possible Implementation
My proposal for this implementation would be to create a script that initially has just the basic functionality described above, which could be created quickly and used to run the comprehensive test suite in the short term. Once implemented, the script could be fleshed out to include all the necessary functionality of run_WE2E_tests.sh, eventually superseding it completely.

Bonuses of a re-write in python include 
 - Working nicely with the existing efforts to python-ize the app
 - Take advantage of existing python tools easily
 - Would make adding additional useful functionality, such as a test summary file detailing the status of each test, much easier. 

### Specific requirements:
 - Should be able to run from login node or batch queue
 - Must be able to stop specific tests without killing all tests
 - *desired but not required* Should be able to resume interrupted test suite

mkavulich:
#466 Represents a good interim solution for the Cheyenne crontab issue. I will be working on a more complete re-write of run_WE2E_tests.sh in python in the coming days for a more complete solution to these issues.

Issue #369, opened 2022-09-22
Workflow end-to-end script should follow standard syntax for arguments, documentation needs updating


mkavulich:
Original issue opened by @climbfuji migrated from regional_workflow repository: https://github.com/ufs-community/regional_workflow/issues/607

## Description
Currently the workflow end-to-end script uses a non-standard format for arguments passed to the script, for example:
```
./run_WE2E_tests.sh \
    tests_file="explist.txt" \
    machine="HERA" \
    account="gsd-hpcs"
```
Standard syntax would be
```
./run_WE2E_tests.sh \
    --tests_file="explist.txt" \
    --machine="HERA" \
    --account="gsd-hpcs"
```
or
```
./run_WE2E_tests.sh \
    -t "explist.txt" \
    -m "HERA" \
    -a "gsd-hpcs"
```
or a mixture of these. This should be changed, because it causes unnecessary confusion.

Also, the documentation in https://ufs-srweather-app.readthedocs.io/en/latest/WE2Etests.html is outdated.

## Solution
Adopt standard syntax and update documentation.

## Alternative Solutions
Do nothing and make life for users harder than necessary.

mkavulich:
Resolved by #586

PR #558, opened 2023-01-24
[develop] First implementation of run_WE2E_tests.py

### LINKED ISSUE #539: Python unittest fails in github actions
## Expected behavior

Current python_unittests github action workflow fails.

## Current behavior

It fails while generating workflow for the community test case.
However, the unittest passes just fine on linux/hera.

## Steps To Reproduce

Open a PR on develop and run a unittest.

mkavulich:
## DESCRIPTION OF CHANGES: 
This PR introduces two new scripts to the repository: `run_WE2E_tests.py` and `monitor_jobs.py`. The purpose of these scripts is to eventually provide a pythonic replacement for the current workflow end-to-end test submission script. Additionally, the monitor_jobs function gives the capability to monitor and submit jobs automatically via the command line or a batch job, rather than relying on crontab entries.

### run_WE2E_tests.py

This new script is roughly analogous to the legacy bash version, `run_WE2E_tests.sh`. This script will set up and run a set of workflow end-to-end tests as specified by the user. 

This script is similar in behavior to the original script, but introduces several improvements and/or simplifications:

 - The argument list is made simpler and more pythonic (which will solve problems listed in #369)
 - The three various test specification arguments (`tests_file`, `test_type`, and `test_name`) are collapsed into a single argument, `--tests`. 
 - The large bit of logic assessing available tests, linking test names, IDs, and subdirectories is omitted; replaced with simple checks based on file names. 
 - "Verbose" and "Debug" options as applied to individual tests are replaced with "verbose_tests" and "debug_tests" respectively, and a "--debug" flag for script verbosity is added.
 - A new "--quiet" flag is introduced, which prevents output (except for warnings and errors) from generate_FV3LAM_workflow printing to screen (they will still be printed to the log file, `log.run_WE2E_tests`). This can prevent important messages from being drowned out by the flood of information printed to screen.
 - Because we are not intermingling all variables from test config, defaults, and machine defaults together, we can better control the resulting test files that are submitted to the generate function
 - Omits checks that are redundant to checks in generate_FV3LAM_workflow.py (for example, ensuring the forecast length is evenly divisible by the LBC interval)
 - Unless `use_cron_to_relaunch` is set to true, the experiment data will be automatically fed into the `monitor_jobs()` function, which will launch and track experiments until all are complete. If the script is interrupted for some reason, it can be re-run and experiments continued (see below)

#### Example usage:
 - Running a list of tests in a file named "testlist" on Hera
`./run_WE2E_tests.py -m=hera -a=fv3lam --tests=testlist`
 - Running the single test "nco_inline_post" on orion, while not printing output from generate_FV3LAM_workflow to screen
`./run_WE2E_tests.py -t=nco_inline_post -m=orion -a=gsd-fv3-test -q`
 - Running the "comprehensive" set of tests on Jet, in debug mode (lots more output print to screen)
`./run_WE2E_tests.py -m=jet -a=gsd-fv3-dev --tests=comprehensive -d`
 - Running the "fundamental" set of tests on Cheyenne, with the gnu compiler, with all experiment directories under the directory "/glade/scratch/kavulich"
`./run_WE2E_tests.py -t=fundamental -c=gnu -m=cheyenne -a=P48500053 --expt_basedir=/glade/scratch/kavulich`

#### Example output:
The `--help` flag gives some usage information:

```
$ ./run_WE2E_tests.py -h
usage: run_WE2E_tests.py [-h] -m MACHINE -a ACCOUNT -t [TESTS ...] [-c COMPILER] [-d] [-q] [--modulefile MODULEFILE]
                         [--run_envir RUN_ENVIR] [--expt_basedir EXPT_BASEDIR] [--exec_subdir EXEC_SUBDIR]
                         [--use_cron_to_relaunch] [--cron_relaunch_intvl_mnts CRON_RELAUNCH_INTVL_MNTS] [--debug_tests]
                         [--verbose_tests]

required arguments:
  -m MACHINE, --machine MACHINE
                        Machine name; see ush/machine/ for valid values
  -a ACCOUNT, --account ACCOUNT
                        Account name for running submitted jobs
  -t [TESTS ...], --tests [TESTS ...]
                        Can be one of three options (in order of priority): 1. A test name or list of test names. 2. A
                        test suite name ("fundamental", "comprehensive", or "all") 3. The name of a file (full or relative
                        path) containing a list of test names.

optional arguments:
  -h, --help            show this help message and exit
  -c COMPILER, --compiler COMPILER
                        Compiler used for building the app
  -d, --debug           Script will be run in debug mode with more verbose output
  -q, --quiet           Suppress console output from workflow generation; this will helpkeep the screen uncluttered
  --modulefile MODULEFILE
                        Modulefile used for building the app
  --run_envir RUN_ENVIR
                        Overrides RUN_ENVIR variable to a new value ( "nco" or "community" ) for all experiments
  --expt_basedir EXPT_BASEDIR
                        Explicitly set EXPT_BASEDIR for all experiments
  --exec_subdir EXEC_SUBDIR
                        Explicitly set EXEC_SUBDIR for all experiments
  --use_cron_to_relaunch
                        Explicitly set USE_CRON_TO_RELAUNCH for all experiments; this option disables the "monitor" script
                        functionality
  --cron_relaunch_intvl_mnts CRON_RELAUNCH_INTVL_MNTS
                        Overrides CRON_RELAUNCH_INTVL_MNTS for all experiments
  --debug_tests         Explicitly set DEBUG=TRUE for all experiments
  --verbose_tests       Explicitly set VERBOSE=TRUE for all experiments

For more information about config arguments (denoted in CAPS), see ush/config_defaults.yaml
```

For a real example of its usage, here I use the `-q` flag to suppress output from `generate_FV3LAM_workflow()`, showing just the output being made by this script:

```
./run_WE2E_tests.py -t=fundamental -m=orion -a=gsd-fv3-test -q 
Checking that all tests are valid
Will run 5 tests:
/work/noaa/gsd-fv3-test/kavulich/UFS/issue_462/ufs-srweather-app/tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta.yaml
/work/noaa/gsd-fv3-test/kavulich/UFS/issue_462/ufs-srweather-app/tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_SUBCONUS_Ind_3km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta.yaml
/work/noaa/gsd-fv3-test/kavulich/UFS/issue_462/ufs-srweather-app/tests/WE2E/test_configs/wflow_features/config.community_ensemble_008mems.yaml
/work/noaa/gsd-fv3-test/kavulich/UFS/issue_462/ufs-srweather-app/tests/WE2E/test_configs/wflow_features/config.deactivate_tasks.yaml
/work/noaa/gsd-fv3-test/kavulich/UFS/issue_462/ufs-srweather-app/tests/WE2E/test_configs/wflow_features/config.inline_post.yaml

Inline post is turned on, deactivating post-processing tasks:
RUN_TASK_RUN_POST = False

calling function that monitors jobs, prints summary
Writing information for all experiments to monitor_jobs_20230124154853.yaml
Checking tests available for monitoring...
Starting experiment grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta running
Starting experiment grid_SUBCONUS_Ind_3km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta running
Starting experiment community_ensemble_008mems running
Starting experiment deactivate_tasks running
Starting experiment inline_post running
Setup complete; monitoring 5 experiments
Experiment deactivate_tasks is COMPLETE; will no longer monitor.
Experiment grid_SUBCONUS_Ind_3km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta is COMPLETE; will no longer monitor.
Experiment grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta is COMPLETE; will no longer monitor.
Experiment inline_post is COMPLETE; will no longer monitor.
Experiment community_ensemble_008mems is COMPLETE; will no longer monitor.
All 5 experiments finished in 0:46:43.495644
All experiments are complete
Summary of results available in monitor_jobs_20230124154853.yaml
```

Note that the message about Inline post is a "warning"-level message from generate_FV3LAM_workflow, and so is still printed to screen.
### monitor_jobs.py

This new script, designed to be called automatically by run_WE2E_tests.py or run stand-alone, will read a dictionary (either provided directly to the function or read from a YAML file) that specifies the location of a number of experiments that need to be monitored. The main function, monitor_jobs(), will keep track of these experiments, advance the workflow with calls to `rocotorun` in each experiment directory, and monitor successes and errors as they occur, reporting a summary at the end.

In addition, while these jobs are being monitored and run, a YAML file tracking the status of all jobs will be written to disk. This file can be read directly to see the details of how each job is coming along, but most importantly, this job file can be read back into monitor_jobs.py as a command line argument. Therefore if the `./run_WE2E_tests.py` script fails or is quit at any point after the experiments have been generated, the script can be re-started and continue to monitor jobs where it left off.

#### Example usage:
In this case, I ran `./run_WE2E_tests.py` but killed it before all experiments completed. Here I just look in my test directory for the latest "monitor_jobs" yaml file, and feed that back to the script:
```
$ ls -l
total 488
-rw-r--r-- 1 Michael.Kavulich fv3lam   5827 Jan 17 15:59 create_WE2E_resource_summary.py
-rwxr-xr-x 1 Michael.Kavulich fv3lam  15347 Jan 17 15:59 get_expts_status.sh
-rwxr-xr-x 1 Michael.Kavulich fv3lam  64794 Jan 17 15:59 get_WE2Etest_names_subdirs_descs.sh
-rw-r--r-- 1 Michael.Kavulich fv3lam  12423 Jan 23 21:33 log.monitor_jobs
-rw-r--r-- 1 Michael.Kavulich fv3lam 182013 Jan 24 21:55 log.run_WE2E_tests
drwxr-sr-x 2 Michael.Kavulich fv3lam   4096 Jan 23 22:19 machine_suites
-rw-r--r-- 1 Michael.Kavulich fv3lam    915 Jan 24 21:55 monitor_jobs_20230124215513.yaml
-rwxr-xr-x 1 Michael.Kavulich fv3lam  12508 Jan 24 00:26 monitor_jobs.py
drwxr-sr-x 2 Michael.Kavulich fv3lam   4096 Jan 24 18:47 __pycache__
drwxr-sr-x 9 Michael.Kavulich fv3lam   4096 Jan 20 04:51 rocoto
-rwxr-xr-x 1 Michael.Kavulich fv3lam  22882 Jan 24 21:50 run_WE2E_tests.py
-rwxr-xr-x 1 Michael.Kavulich fv3lam  50269 Jan 17 16:00 run_WE2E_tests.sh
-rwxr-xr-x 1 Michael.Kavulich fv3lam   2844 Jan 17 15:59 setup_WE2E_tests.sh
drwxr-sr-x 6 Michael.Kavulich fv3lam   4096 Jan 23 16:58 test_configs
$ cat monitor_jobs_20230124222802.yaml 
### WARNING ###
### THIS FILE IS AUTO_GENERATED AND REGULARLY OVER-WRITTEN BY monitor_jobs.py
### EDITS MAY RESULT IN MISBEHAVIOR OF EXPERIMENTS RUNNING
deactivate_tasks:
  expt_dir: /scratch2/BMC/fv3lam/kavulich/UFS/workdir/issue_462_new_WE2E_script/expt_dirs/deactivate_tasks
  status: CREATED
nco_grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR:
  expt_dir: /scratch2/BMC/fv3lam/kavulich/UFS/workdir/issue_462_new_WE2E_script/expt_dirs/nco_grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR
  status: CREATED
custom_ESGgrid:
  expt_dir: /scratch2/BMC/fv3lam/kavulich/UFS/workdir/issue_462_new_WE2E_script/expt_dirs/custom_ESGgrid
  status: CREATED
$ ./monitor_jobs.py -y=monitor_jobs_20230124215513.yaml 
Checking tests available for monitoring...
Starting experiment deactivate_tasks running
Starting experiment nco_grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR running
Starting experiment custom_ESGgrid running
Setup complete; monitoring 3 experiments
Experiment deactivate_tasks is COMPLETE; will no longer monitor.
Experiment nco_grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR is COMPLETE; will no longer monitor.
Experiment custom_ESGgrid is COMPLETE; will no longer monitor.
All 3 experiments finished in 0:15:59.928853
```

And querying the monitor file afterwards shows that all experiments are indeed complete:
```
$ cat monitor_jobs_20230124215513.yaml 
### WARNING ###
### THIS FILE IS AUTO_GENERATED AND REGULARLY OVER-WRITTEN BY monitor_jobs.py
### EDITS MAY RESULT IN MISBEHAVIOR OF EXPERIMENTS RUNNING
deactivate_tasks:
  expt_dir: /scratch2/BMC/fv3lam/kavulich/UFS/workdir/issue_462_new_WE2E_script/expt_dirs/deactivate_tasks
  status: COMPLETE
  make_grid_201907010000: SUCCEEDED
  make_orog_201907010000: SUCCEEDED
  make_sfc_climo_201907010000: SUCCEEDED
nco_grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR:
  expt_dir: /scratch2/BMC/fv3lam/kavulich/UFS/workdir/issue_462_new_WE2E_script/expt_dirs/nco_grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR
  status: COMPLETE
  get_extrn_ics_202008100000: SUCCEEDED
  get_extrn_lbcs_202008100000: SUCCEEDED
  make_ics_202008100000: SUCCEEDED
  make_lbcs_202008100000: SUCCEEDED
  run_fcst_202008100000: SUCCEEDED
custom_ESGgrid:
  expt_dir: /scratch2/BMC/fv3lam/kavulich/UFS/workdir/issue_462_new_WE2E_script/expt_dirs/custom_ESGgrid
  status: COMPLETE
  make_grid_201907010000: SUCCEEDED
  get_extrn_ics_201907010000: SUCCEEDED
  get_extrn_lbcs_201907010000: SUCCEEDED
  make_orog_201907010000: SUCCEEDED
  make_sfc_climo_201907010000: SUCCEEDED
  make_ics_201907010000: SUCCEEDED
  make_lbcs_201907010000: SUCCEEDED
  run_fcst_201907010000: SUCCEEDED
  run_post_f000_201907010000: SUCCEEDED
  run_post_f001_201907010000: SUCCEEDED
  run_post_f002_201907010000: SUCCEEDED
  run_post_f003_201907010000: SUCCEEDED
  run_post_f004_201907010000: SUCCEEDED
  run_post_f005_201907010000: SUCCEEDED
  run_post_f006_201907010000: SUCCEEDED
```

### Additional changes

In addition to the new scripts, a few changes have been made to the rest of the workflow that should have no significant impact on existing tests:

 - generate_FV3LAM_wflow.py
    - function generate_FV3LAM_wflow() now takes an optional "debug" argument that, when true, will provide more verbose output from this and other functions that receive the argument.
    - function generate_FV3LAM_wflow() now returns EXPTDIR, the string indicating the full path of the created experiment. This is useful for scripts/functions that call this function directly rather than via command-line
    - Move the "Experiment generation completed" message to the *very* end of the script, so that any failure will pre-empt it
    - Generalize logging setup so that it will work properly even if called from another function.

 - setup.py
   - function setup() now takes an optional "debug" argument that, when true, will provide more verbose output
   - Add several debug logging messages, remove an old "print" statement
   - Avoid using "cd" and other full-path changes so that results will be the same regardless of the directory this function is called from

 - tests/WE2E/test_configs/wflow_features/config.deactivate_tasks.yaml
   - Remove sections for tasks that are not used (LBCS and ICS are not staged since those tasks are deactivated

### Missing features vs. run_WE2E_tests.sh

This initial version is incomplete, and a few more capabilities must be added before a wholesale replacement of run_WE2E_tests.sh

 - The command-line options `cron_relaunch_intvl_mnts`, `generate_csv_file`, and `opsroot` are not yet implemented
 - Currently, the default behavior is to *not* submit jobs to cron, but rather to use monitor_jobs.py to monitor and run the experiments
 - The ability to include calls to UNIX `date` in DATE_FIRST_CYCL and DATE_LAST_CYCL is not yet implemented
 - No checks that all MET executables/directories are set up properly for the given machine...should these be implemented in setup.py instead?
 - No hard-coded "maxtries" based on platform yet
 - Potentially other edge use cases I have missed?

### Type of change

## TESTS CONDUCTED: 

- [x] hera.intel
- [x] orion.intel
- [x] cheyenne.intel
- [x] cheyenne.gnu
- [ ] gaea.intel
- [x] jet.intel
- [ ] wcoss2.intel
- [ ] Jenkins
- [x] fundamental test suite
- [x] comprehensive tests (specify *which* if a subset was used)
    - [x] Jet

## DEPENDENCIES:
None

## DOCUMENTATION:
Will be contributed with later PR deprecating old system. This is a preliminary implementation for wider exposure and feedback.

## ISSUE: 
 - This is partially in response to #462. Still needs some more work before it can be closed.
 - When this script supersedes the legacy bash script, it will also solve #369
 - Some changes I had to make here (related to calling the generate function from another directory) also fix #539

## CHECKLIST


MichaelLueken (approved):
@mkavulich Thank you very much for creating the PR to bring in the initial implementation of the pythonized run_WE2E_tests.sh script!  I was able to run a manual test to ensure that Jenkins tests would still work, then submitted the fundamental tests using the new run_WE2E_tests.py script.  Everything worked like a charm!  Approving these changes.

ush/generate_FV3LAM_wflow.py
```diff
@@ -744,21 +731,36 @@ def generate_FV3LAM_wflow(ushdir, logfile: str = "log.generate_FV3LAM_wflow") ->
...
+    fh = logging.FileHandler(logfile, mode='w')
+    fh.setLevel(logging.DEBUG)
+    fh.setFormatter(formatter)
+    logging.getLogger().addHandler(fh)
     logging.debug(f"Finished setting up debug file logging in {logfile}")
+
+    # If there are already multiple handlers, that means generate_FV3LAM_workflow was called from another function.
+    # In that case, do not change the console (print-to-screen) logging.
+    if len(logging.getLogger().handlers) > 1:
+        return
```
danielabdi-noaa, line 757: I am curious, was this what solved the unittest failures?

MichaelLueken:
@mkavulich The Jenkins test, `specify_template_filenames`, on Cheyenne Intel, is failing with the following in the run_fcst step:

```
NOTE from PE     0: diag_manager_mod::diag_manager_init: diag_manager is using fms2_io

FATAL from PE     0: get_ascii_file_num_lines: File diag_table does not exist.


FATAL from PE     0: get_ascii_file_num_lines: File diag_table does not exist.

Image              PC                Routine            Line        Source
ufs_model          000000000432EFB7  Unknown               Unknown  Unknown
ufs_model          000000000274B6E9  mpp_mod_mp_mpp_er          70  mpp_util_mpi.inc
ufs_model          0000000002742A81  mpp_mod_mp_get_as        1476  mpp_util.inc
ufs_model          0000000002612E75  fms_io_utils_mod_         507  fms_io_utils.F90
```
Please see `/glade/scratch/epicufsrt/jenkins/workspace/fs-srweather-app_pipeline_PR-558__2/expt_dirs/specify_template_filenames` for more details.

gsketefian:
@mkavulich One thing I noticed with the current (bash) version of `run_WE2E_tests.sh` is that if there is an entry in the test config yaml file (`config.<test_name>.yaml`) that is not defined in `config_defaults.yaml`, it is not flagged by the script but simply ignored.  Does the new python version `run_WE2E_tests.py` in this PR flag such entries (haven't gotten to trying it myself yet)?  Thanks.

tests/WE2E/test_configs/wflow_features/config.specify_template_filenames.yaml
```diff
@@ -5,11 +5,11 @@ metadata:
...
   RUN_ENVIR: community
 workflow:
-  DATA_TABLE_TMPL_FN: data_table
-  DIAG_TABLE_TMPL_FN: diag_table.FV3_GFS_v15p2
-  FIELD_TABLE_TMPL_FN: field_table.FV3_GFS_v15p2
-  MODEL_CONFIG_TMPL_FN: model_configure
-  NEMS_CONFIG_TMPL_FN: nems.configure
+  DATA_TABLE_FN: data_table
+  DIAG_TABLE_FN: diag_table.FV3_GFS_v15p2
+  FIELD_TABLE_FN: field_table.FV3_GFS_v15p2
```
MichaelLueken: Since these two table indices include .{{ CCPP_PHYS_SUITE }}, shouldn't they be DIAG_TABLE_TMPL_FN and FIELD_TABLE_TMPL_FN?  Making this change on Cheyenne and manually rerunning the specify_template_filenames test allowed the test to run successfully.

MichaelLueken (changes requested):
@mkavulich After a little digging, I think I discovered the issue that is causing the failure in the specify_template_filenames test for Cheyenne Intel.  From the ush/config_defaults.yaml file, if either the diag_table or field_table include .{{ CCPP_PHYS_SUITE }}, then it should be named either DIAG_TABLE_TMPL_FN or FIELD_TABLE_TMPL_FN, respectively.  Making this change and manually rerunning the specify_template_filenames test showed that it ran successfully.

tests/WE2E/test_configs/wflow_features/config.specify_template_filenames.yaml
mkavulich: @MichaelLueken This was related to the problem I mentioned yesterday: these variables are not named correctly in develop, and that was causing the new script to fail. However, this fix needs to be applied more carefully than this, so I have reverted these changes for now in order to avoid the failure you pointed out. I will open an issue later with more details to hopefully fix this.

MichaelLueken: @mkavulich Thank you very much for the explanation and reverting this change for now.  I certainly agree that the entries for `DATA_TABLE_TMPL_FN`, `MODEL_CONFIG_TMPL_FN`, and `NEMS_CONFIG_TMPL_FN` appear to be set incorrectly with respect to the `specify_template_filenames` test.  As a matter of fact, there are no entries for either `MODEL_CONFIG_TMPL_FN` or `NEMS_CONFIG_TMPL_FN` in `config_defaults.yaml` and `data_table` would need to be changed to `data_table.FV3_GFS_v15p2` in order to properly be set to `DATA_TABLE_TMPL_FN`, so I can agree that these would likely cause the new script to fail.  I'll resolve this conversion for now, reapprove, and then rerun the Cheyenne tests.

MichaelLueken (approved):

ush/generate_FV3LAM_wflow.py
mkavulich: These changes in `setup_logging()` were needed to get log files (and messages) to behave correctly when called from another function with logging, so if that's how the unit tests are being called then I bet these changes as a whole are what fixed them. Specifically, the use of `logging.basicConfig` was the issue, as it has no effect when there already exists a logging handler. This could be solved with some settings that are only available in Python 3.8+, but this new implementation is cleaner anyway IMO.

tests/WE2E/monitor_jobs.yaml
```diff
@@ -0,0 +1,54 @@
+# This is an example yaml file showing the various entries that can be created for tracking jobs by monitor_jobs.py
```
christinaholtNOAA, line 1: How should this file be updated when changes are made to monitor_jobs.py such that this file is modified?

Can that be a GitHub action?

tests/WE2E/monitor_jobs.py
```diff
@@ -0,0 +1,280 @@
...
+    write_monitor_file(monitor_file,expt_dict)
+
+    # Perform initial setup for each experiment
+    logging.info("Checking tests available for monitoring...")
+    num_expts = 0
+    for expt in expt_dict:
+        logging.info(f"Starting experiment {expt} running")
+        num_expts += 1
+        rocoto_db = f"{expt_dict[expt]['expt_dir']}/FV3LAM_wflow.db"
+        subprocess.run(["rocotorun", f"-w {expt_dict[expt]['expt_dir']}/FV3LAM_wflow.xml", f"-d {rocoto_db}"])
```
christinaholtNOAA: Why not similarly define `rocoto_xml` for consistency?

tests/WE2E/monitor_jobs.py
```diff
@@ -0,0 +1,280 @@
...
+        except:
+            logging.warning(f"Unable to read database {rocoto_db}\nWill not track experiment {expt}")
+            expt_dict[expt]["status"] = "ERROR"
+            num_expts -= 1
+            continue
+        for task in db:
+            # For each entry from rocoto database, store that under a dictionary key named TASKNAME_CYCLE
+            # Cycle comes from the database in Unix Time (seconds), so convert to human-readable
+            cycle = datetime.utcfromtimestamp(task[1]).strftime('%Y%m%d%H%M')
+            expt_dict[expt][f"{task[0]}_{cycle}"] = task[2]
```
christinaholtNOAA: `task[2]` here depends on your `SELECT` statement never changing. It might be helpful for the user if you made a named tuple to hold the information in the task so that you don't have to worry about order here and losing track.

tests/WE2E/monitor_jobs.py
```diff
@@ -0,0 +1,280 @@
...
+        expt_dict[expt] = update_expt_status(expt_dict[expt], expt)
+        #Run rocotorun again to get around rocotobqserver proliferation issue
+        subprocess.run(["rocotorun", f"-w {expt_dict[expt]['expt_dir']}/FV3LAM_wflow.xml", f"-d {rocoto_db}"])
```
christinaholtNOAA: Will the calls to subprocess like this wait until `rocotorun` completes before control is returned to the Python script? If not, you might be over-taxing the system here.

Also, is `rocotorun` also running in a cron job at this point?

tests/WE2E/monitor_jobs.py
```diff
@@ -0,0 +1,280 @@
...
+
+    write_monitor_file(monitor_file,expt_dict)
+
+    logging.info(f'Setup complete; monitoring {num_expts} experiments')
+
+    #Make a copy of experiment dictionary; will use this copy to monitor active experiments
+    running_expts = expt_dict.copy()
```
christinaholtNOAA, line 56: Do you need a deep copy here? Python docs say this is a shallow copy, so you might be updating `expt_dict` when you write to `running_expts`

tests/WE2E/monitor_jobs.py
```diff
@@ -0,0 +1,280 @@
...
+
+    i = 0
+    while running_expts:
+        i += 1
+        for expt in running_expts.copy():
+            logging.debug(f"Updating status of {expt}")
```
christinaholtNOAA: This seems awfully similar to the code above. Perhaps a function would be helpful to reduce duplication?

tests/WE2E/monitor_jobs.py
```diff
@@ -0,0 +1,280 @@
...
+    statuses = list()
+    for task in expt:
+        # Skip non-task entries
+        if task in ["expt_dir","status"]:
+            continue
+        statuses.append(expt[task])
+
+    if "DEAD" in statuses:
+        if ( "RUNNING" in statuses ) or ( "SUBMITTING" in statuses ) or ( "QUEUED" in statuses ):
+            expt["status"] = "DYING"
```
christinaholtNOAA, line 172: This is more pythonic, but just sharing...

```
checks = ["RUNNING", "SUBMITTING", "QUEUED"]
if any(status in checks for status in statuses):
    ....
```

tests/WE2E/monitor_jobs.py
```diff
@@ -0,0 +1,280 @@
...
+             SUCCEEDED.
+             This is a normal state; we will continue to monitor this experiment.
+    SUCCEEDED: All jobs are status SUCCEEDED; we will monitor for one more cycle in case there are
+             unsubmitted jobs remaining.
+    COMPLETE:All jobs are status SUCCEEDED, and we have monitored this job for an additional cycle
+             to ensure there are no un-submitted jobs. We will no longer monitor this experiment.
+    """
+
+    #If we are no longer tracking this experiment, return unchanged
+    if expt["status"] in ['DEAD','ERROR','COMPLETE']:
```
christinaholtNOAA, line 132: Is `expt` guaranteed to have a `status` entry here?

tests/WE2E/monitor_jobs.py
```diff
@@ -0,0 +1,280 @@
...
+        raise
+
+def sqlite_read(db: str, ex: str) -> list:
+#    # Create Named Tuple for better data organization
+#    Task = namedtuple("taskname","cycle","state")
+
+    with closing(sqlite3.connect(db)) as connection:
+        with closing(connection.cursor()) as cur:
+            db = cur.execute(ex).fetchall()
+    return db
```
christinaholtNOAA: Any chance this is a problem if a user calls this script for a real-time experiment whose first cycle is a LONG time ago and the cycles since then are all inactive?

ush/setup.py
```diff
@@ -152,9 +158,6 @@ def load_config_for_setup(ushdir, default_config, user_config):
             if not (v is None or v == ""):
                 cfg_d[sect][k] = str_to_list(v)
 
-    for k, v in cfg_d["task_run_fcst"].items():
-        print(f"*** {k}: {v}")
-
```
christinaholtNOAA: Eek! Sorry.

christinaholtNOAA (commented):
@mkavulich I reviewed all the files _except_ the main show (run_WE2E_tests.py)...I should be able to come back to that a little later. Sorry to break it up into sections.

tests/WE2E/run_WE2E_tests.py
```diff
@@ -0,0 +1,447 @@
...
+        args : The argparse.Namespace object containing command-line arguments
+    Returns:
+        None
+    """
+
+    # Set up logging to write to screen and logfile
+    setup_logging(debug=args.debug)
+
+    # Set some important directories
+    ushdir=homedir + '/ush'
```
christinaholtNOAA: `os.path.join` will help with platform portability issues.

tests/WE2E/run_WE2E_tests.py
```diff
@@ -0,0 +1,447 @@
...
+
+    testfiles = glob.glob('test_configs/**/config*.yaml', recursive=True)
+    tests_to_run=[]
+    for test in tests:
+        # Skip blank/empty testnames; this avoids failure if newlines or spaces are included
+        if test.isspace():
+            continue
+        match=False
+        # Search for exact config file name to avoid accidental partial matches
+        test_config='config.' + test.rstrip() + '.yaml'
```
christinaholtNOAA: Consider an f-string here. They're generally more readable.

Is there any chance that the `test` might have leading whitespace? Would `strip` be better?

tests/WE2E/run_WE2E_tests.py
```diff
@@ -0,0 +1,447 @@
...
+    for testfile in tests_to_run.copy():
+        if os.path.islink(testfile):
+            if os.path.realpath(testfile) in tests_to_run:
+                logging.warning(dedent(f"""WARNING: test file {testfile} is a symbolic link to a
+                                test file ({os.path.realpath(testfile)}) that is also included in the
+                                test list. Only the latter test will be run."""))
+                tests_to_run.remove(testfile)
+    if len(tests_to_run) != len(set(tests_to_run)):
+        logging.warning("\nWARNING: Duplicate test names were found in list. Removing duplicates and continuing.\n")
+        tests_to_run = list(set(tests_to_run))
```
christinaholtNOAA, line 247: These two checks are _real_ nice!

tests/WE2E/run_WE2E_tests.py
```diff
@@ -0,0 +1,447 @@
...
+                        run_envir = 'nco'
+                        logging.debug(f'{testfilename} exists for this platform and run_envir has not been specified'\
+                                       'Setting run_envir = {run_envir} for all tests')
+                logging.debug(f"Reading test file: {testfilename}")
+                tests_to_check = list(open(testfilename))
+                logging.debug(f"Will check {user_spec_tests[0]} tests:\n{tests_to_check}")
+            else:
+                # If we have gotten this far then the only option left for user_spec_tests is a file containing test names
+                if os.path.isfile(user_spec_tests[0]):
+                    tests_to_check = list(open(user_spec_tests[0]))
```
christinaholtNOAA: Where does this get closed?

tests/WE2E/run_WE2E_tests.py
```diff
@@ -0,0 +1,447 @@
...
+        logging.debug(f"User specified a list of tests:\n{tests_to_check}")
+    else:
+        #First see if args.tests is a valid test name
+        user_spec_tests = args.tests
+        try:
+            logging.debug(f'Checking if {user_spec_tests} is a valid test name')
+            _ = check_tests(user_spec_tests)
+            # If no error, user_spec_tests is the name of the test specified on the command line
+            tests_to_check = user_spec_tests
+        except:
```
christinaholtNOAA: From a more purist perspective, using a try/except block is really meant to catch and handle exceptional situations. In this case, it seems it's being used to control logic. While that's not completely un-allowed, a blind except here does not meet the PEP-8 standard for good reason.

What if a user misspells a test name, but had every intent of doing this option correctly? I believe we'd get no useful information from the try/except block, and it would go along attempting to process the input as an attempt at specifying a file name and fail.

It may be helpful to check to see if a provided a file name...does it contain ".yaml", for example? Is one of the key words? Then, is it a test name?

tests/WE2E/run_WE2E_tests.py
```diff
@@ -0,0 +1,447 @@
...
+            # If not a valid test name, check if it is a test suite
+            if user_spec_tests[0] == 'all':
+                alltests = glob.glob('test_configs/**/config*.yaml', recursive=True)
+                tests_to_check = []
+                for f in alltests:
+                    filename = os.path.basename(f)
+                    filename = filename[7:-5]
```
christinaholtNOAA: Why `[7:-5]`? I assume it's because `config*.yaml`, but maybe a comment would be helpful?

tests/WE2E/run_WE2E_tests.py
```diff
@@ -0,0 +1,447 @@
...
+                    tests_to_check.append(filename)
+                logging.debug(f"Will check all tests:\n{tests_to_check}")
+            elif user_spec_tests[0] == 'fundamental' or user_spec_tests[0] == 'comprehensive':
```
christinaholtNOAA: Shorter option:
```
elif user_spec_tests[0] in ['fundamental', 'comprehensive']:
```

tests/WE2E/run_WE2E_tests.py
```diff
@@ -0,0 +1,447 @@
...
+                # I am writing this section of code under protest; we should use args.run_envir to check for run_envir-specific files!
```
christinaholtNOAA, line 66: I understand your concern!

While it's hideous, you may consider reducing duplication by setting a variable early on to set the "base file path" and then use that to build up your many options of that file name. If the files were all moved to a different subdir, for example, that'd be a LOT of edits to make instead of one.

tests/WE2E/run_WE2E_tests.py
```diff
@@ -0,0 +1,447 @@
...
+
+    logging.info("Checking that all tests are valid")
+
+    tests_to_run=check_tests(tests_to_check)
+
+    pretty_list = "\n".join(str(x) for x in tests_to_run)
+    logging.info(f'Will run {len(tests_to_run)} tests:\n{pretty_list}')
+
+
+    config_default_file = ushdir + '/config_defaults.yaml'
```
christinaholtNOAA: os.path.join() here would be better. And `machine_file` below.

tests/WE2E/run_WE2E_tests.py
```diff
@@ -0,0 +1,447 @@
...
+        cfg_ics : Updated dictionary for task_get_extrn_ics section of test config
+    """
+
+    #Make our lives easier by shortening some dictionary calls
+    cfg_ics = cfg['task_get_extrn_ics']
+
+    # If RUN_TASK_GET_EXTRN_ICS is false, do nothing and return
+    if 'workflow_switches' in cfg:
+        if 'RUN_TASK_GET_EXTRN_ICS' in cfg['workflow_switches']:
+            if cfg['workflow_switches']['RUN_TASK_GET_EXTRN_ICS'] is False:
```
christinaholtNOAA, line 291: We don't often have a bit of logic that is necessarily this wordy. Perhaps these lines can be simplified?

```
if not cfg['workflow_switches'].get('RUN_TASK_GET_EXTRN_ICS', True):
    return cfg_ics
```

This logic says, if it's set and False, return. If it's not there, keep going. If it's there and set to True keep going.

tests/WE2E/run_WE2E_tests.py
```diff
@@ -0,0 +1,447 @@
...
+                return cfg_ics
+
+    # If USE_USER_STAGED_EXTRN_FILES not specified or false, do nothing and return
+    if 'USE_USER_STAGED_EXTRN_FILES' not in cfg_ics:
+        logging.debug(f'USE_USER_STAGED_EXTRN_FILES not specified in task_get_extrn_ics section of config')
+        return cfg_ics
+    if not cfg_ics['USE_USER_STAGED_EXTRN_FILES']:
+        logging.debug(f'USE_USER_STAGED_EXTRN_FILES is false for task_get_extrn_ics section of config')
+        return cfg_ics
```
christinaholtNOAA, line 297: Similar thing here...if it doesn't really matter to the user if it's not set or False:

```
if not cfg_ics.get('USE_USER_STAGED_EXTRN_FILES'):
    return cfg_ics
```

This ability to check if a key exists may also potentially be helpful in cleaning up the logic below surrounding `EXTRN_MDL_SYSBASEDIR_ICS`.

tests/WE2E/run_WE2E_tests.py
```diff
@@ -0,0 +1,447 @@
...
+        cfg_ics['EXTRN_MDL_SOURCE_BASEDIR_ICS'] = f"{mach['platform']['TEST_EXTRN_MDL_SOURCE_BASEDIR']}/"\
+                                                  f"{cfg_ics['EXTRN_MDL_NAME_ICS']}/{cfg_ics['FV3GFS_FILE_FMT_ICS']}/${{yyyymmddhh}}"
+    else:
+        cfg_ics['EXTRN_MDL_SOURCE_BASEDIR_ICS'] = f"{mach['platform']['TEST_EXTRN_MDL_SOURCE_BASEDIR']}/"\
+                                                  f"{cfg_ics['EXTRN_MDL_NAME_ICS']}/${{yyyymmddhh}}"
+
+    return cfg_ics
+
+def check_task_get_extrn_lbcs(cfg: dict, mach: dict, dflt: dict) -> dict:
+    """
```
christinaholtNOAA, line 332: This is a big copy with some changes from ICS to LBCS!

christinaholtNOAA (commented):
This looks good to me! Only a handful of comments about minor changes you may consider.

mkavulich:
@gsketefian to answer your question here, yes, this script does solve the issue of the current script omitting invalid options silently. This new script takes a different approach to constructing the config.yaml file from the provided test one, in that we only perform the necessary manipulations to test the features needed and do not attempt to rebuild config.yaml from scratch. So if there is an invalid option provided, that will be passed through to the `generate_FV3LAM_workflow()` function, where the error will be caught and fail.

As an aside, I have documented this issue in #568, so it can be properly addressed

tests/WE2E/run_WE2E_tests.py
```diff
@@ -0,0 +1,447 @@
...
+        test_config='config.' + test.rstrip() + '.yaml'
+        for testfile in testfiles:
+            if test_config in testfile:
+                logging.debug(f"found test {test}")
+                match=True
+                tests_to_run.append(os.path.abspath(testfile))
+        if not match:
+            raise Exception(f"Could not find test {test}")
+    # Because some test files are symlinks to other tests, check that we don't
+    # include the same test twice
```
gsketefian, line 237: @mkavulich I thought you were getting rid of the symlinks feature.  Seems like you're still checking for that here.

mkavulich: Sorry if I was confusing in our previous conversation; I had no plans to change the existence of symlinks; I only changed it so that this does not result in a failure, it simply omits duplicate tests. Whether or not that's the best approach, I'm all ears on opinions there.

tests/WE2E/run_WE2E_tests.py
```diff
@@ -0,0 +1,447 @@
...
+    #Get the "Home" directory, two levels above this one
+    homedir=os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+    logfile='log.run_WE2E_tests'
+
+    #Parse arguments
+    parser = argparse.ArgumentParser(epilog="For more information about config arguments (denoted in CAPS), see ush/config_defaults.yaml\n")
+    optional = parser._action_groups.pop() # Create a group for optional arguments so they can be listed after required args
+    required = parser.add_argument_group('required arguments')
+
+    required.add_argument('-m', '--machine', type=str, help='Machine name; see ush/machine/ for valid values', required=True)
```
gsketefian, line 428: @mkavulich The "machine" argument needs to be made lower-case before being used since this failed:
```
./run_WE2E_tests.py --machine Hera --account rtrr --tests MET_verification --expt_basedir set01
```
with
```
FileNotFoundError: [Errno 2] No such file or directory: '/scratch2/BMC/det/Gerard.Ketefian/UFS_CAM/TEST_mkavulich_new_run_WE2E_tests/ufs-srweather-app/ush/machine/Hera.yaml'
```
but changing "Hera" to "hera" worked.

gsketefian:
@mkavulich I tried testing what happens with duplicate file names:  I copied `wflow_features/config.MET_verification.yaml` to a new directory under `test_configs` named `verification`, i.e. I created the file `verification/config.MET_verification.yaml`.  The files are identical.  

When I launched `run_WE2E_tests.py`, it tried running both instances, so it would be good to have it check for duplicate tests under the `test_configs` directory.  In this case, one of the `MET_verification` experiment directories got renamed to `MET_verification_old_...` (probably when the 2nd one launched), and the script seemed to monitor one experiment thereafter (I guess the first test failed because of the directory rename, but I'm not sure).

mkavulich:
@gsketefian The case of duplicate file names is not one I had accounted for; I had only accounted for the user specifying duplicate tests. I just pushed a new commit with a check that there are no config.TESTNAME.yaml files with the same TESTNAME.

tests/WE2E/monitor_jobs.yaml
mkavulich: It is unclear to me how that would be handled in an automatic way. Is it worth the effort for something that will likely not change very often, if at all?

tests/WE2E/monitor_jobs.py
mkavulich: `subprocess.run` waits for the subprocess to complete before continuing ([python documentation here](https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module)). I decided not to try to capture the output from the subprocess because this seems to slow things down drastically (like, orders of magnitude slower).

The `monitor_jobs()` function is only called by `run_WE2E_tests.py` if the `--use_cron_to_relaunch` option is not specified, so this should never be run in conjunction with crontab running. But I can't control what users choose to do; theoretically this should work just fine if someone also had a crontab entry for this job and called this script manually.

tests/WE2E/monitor_jobs.py
mkavulich: From my reading of [the python docs](https://docs.python.org/3/library/copy.html) that should not be necessary. The logic in this loop does not change any of the individual entries to the list (edit: in a way that will not also be reflected in the list copy; it could be that the line `running_expts[expt] = expt_dict[expt]` is redundant), it merely pops each entry out of the copy of the list as the corresponding test completes. A shallow copy should be sufficient for this purpose.

tests/WE2E/monitor_jobs.py
mkavulich: It is unless the input dictionary is misformatted. When the dict is created (either after the input file is read or in `run_WE2E_tests()`) the status is always set.

tests/WE2E/monitor_jobs.py
mkavulich: This script is not designed with real-time experiments in mind, just the WE2E tests. If realtime functionality is desired and the current logic is insufficient then those considerations can be taken into account in future updates.

tests/WE2E/monitor_jobs.py
mkavulich: I did you one better and defined the whole command passed to `subprocess.run`, since that is always the same.

tests/WE2E/run_WE2E_tests.py
mkavulich: Allegedly (via the wisdom of stackoverflow) files are automatically closed when they go out of scope, and here the scope is only within the call to `list()`. However, I can't find any explicit reference to this in the official python documentation, so I have changed to a two-liner `with` statement.

tests/WE2E/run_WE2E_tests.py
mkavulich: Re-arranged to get rid of "impure" `try` block 😉

tests/WE2E/run_WE2E_tests.py
mkavulich: Fixed, this was needed a few other places too.

tests/WE2E/run_WE2E_tests.py
mkavulich: Consolidated more duplicate code into `update_expt_status()`

tests/WE2E/run_WE2E_tests.py
mkavulich: The example you give will fail if the key `workflow_switches` is not set, and it will return if RUN_TASK_GET_EXTRN_ICS is not set, neither of which is acceptable. Unfortunately because we can't guarantee that this key exists, *and* we need to check that *if* the key cfg['workflow_switches']['RUN_TASK_GET_EXTRN_LBCS'] is set it is *false* (if it is not set, it defaults to `True`), this logic necessarily needs to be a little convoluted.

tests/WE2E/run_WE2E_tests.py
mkavulich: Simplified logic using `.get()` method here and in lbcs function

tests/WE2E/monitor_jobs.py
mkavulich: I thought I commented here but I guess it got lost. I updated the `update_expt_status()` function to consolidate more duplicate code.

tests/WE2E/monitor_jobs.py
mkavulich: I do not see a way to convert the output from fetchall() to a list of namedtuples without significantly increasing the complexity of the code. I'd rather just handle this with a comment.

mkavulich:
@gsketefian @christinaholtNOAA Thank you for your reviews, I believe I have addressed all your concerns. Let me know if I missed something or you have more questions/comments.

gsketefian:
@mkavulich I was testing redundancy with symlinks (I had a symlink to a test, and I listed both in my list of tests), and the script caught the redundancy with an appropriate warning.  Just the output to screen is not well-formatted (seems to be a lot of extra spaces where there should be just one) and thus a bit hard to read.  Do you mind fixing if possible?  Here's what the output looked like:
```
$ ./run_WE2E_tests.py --machine hera --account rtrr --tests my_tests.txt --expt_basedir /scratch2/BMC/det/Gerard.Ketefian/UFS_CAM/TEST_mkavulich_new_run_WE2E_tests/expt_dirs/set01 -q
Checking that all tests are valid
WARNING: test file /scratch2/BMC/det/Gerard.Ketefian/UFS_CAM/TEST_mkavulich_new_run_WE2E_tests/ufs-srweather-app/tests/WE2E/test_configs/verification/config.MET_verification_link.yaml is a symbolic link to a
                                test file (/scratch2/BMC/det/Gerard.Ketefian/UFS_CAM/TEST_mkavulich_new_run_WE2E_tests/ufs-srweather-app/tests/WE2E/test_configs/wflow_features/config.MET_verification.yaml) that is also included in the
                                test list. Only the latter test will be run.
Will run 1 tests:
/scratch2/BMC/det/Gerard.Ketefian/UFS_CAM/TEST_mkavulich_new_run_WE2E_tests/ufs-srweather-app/tests/WE2E/test_configs/wflow_features/config.MET_verification.yaml
Workflow for test MET_verification successfully generated in
/scratch2/BMC/det/Gerard.Ketefian/UFS_CAM/TEST_mkavulich_new_run_WE2E_tests/expt_dirs/set01/MET_verification

calling function that monitors jobs, prints summary
Writing information for all experiments to monitor_jobs_20230202091032.yaml
Checking tests available for monitoring...
Starting experiment MET_verification running
Setup complete; monitoring 1 experiments
```

gsketefian:
@mkavulich Two questions came to mind as I was doing further tests (these are just for discussion; I'm approving the PR):
1) The current (shell) version searches only a specific set of subdirectories under `test_config` for tests.  I had done it that way because often during testing I may make a backup copy of a subdirectory before starting to modify the files in it.  The new script searches all subdirectories.  I'm ok with that (I can always copy backup directories to somewhere outside of `test_configs`), just wondering how you decided on that.  Just code simplicity?
2) Once the cron features are enabled in a future PR, will the monitoring script be able to monitor tasks running with cron as well?  Currently, there's a script (`get_expts_status.sh`) that does this.  It would be nice to have a proper replacement for it (or just leave it as is if not).

gsketefian (approved):

tests/WE2E/monitor_jobs.yaml
christinaholtNOAA: It could be nice just to drop in a comment on how this appeared here so that future you is not completely baffled. ;)

tests/WE2E/run_WE2E_tests.py
christinaholtNOAA: Oh. Sure. What about:

```
if cfg.get('workflow_switches', {}).get('RUN_TASK_GET_EXTRN_ICS', True) is False:
    return cfg_ics
```

The logic should return if `workflow_switches` is set and `RUN_TASK_GET_EXTRN_ICS` is set to something that evaluates to False.

If `workflow_switches` or `RUN_TASK_GET_EXTRN_ICS` isn't set, or `RUN_TASK_GET_EXTRN_ICS` evaluates to something not False, it does not return.

christinaholtNOAA (approved):
As I mentioned last week, I'm totally fine with the PR as-is, but left just a couple of follow up comments just to circle back to the review.

mkavulich:
Thanks @christinaholtNOAA; for expediency and to avoid unnecessary testing, I will roll these suggestions into my working branch for the next update to these scripts.

Issue #587, opened 2023-02-08
Overhaul and consolidate WE2E tests, identify needed additional tests


mkavulich:
## Description
It has been a while since we re-visited the suite of WE2E tests. In that time, a lot of development has happened, and the way various test suites are used has changed. We should assess the current state of the WE2E tests, and discuss ways in which they can be improved.

**This issue will be modified and expanded as the effort continues.**

#### Duplication
There are many examples of this, but just to pick out one, these nine tests only have a single difference among them: the specific output grid being used
 - config.grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16.yaml
 - config.grid_CONUS_3km_GFDLgrid_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16.yaml
 - config.grid_CONUS_25km_GFDLgrid_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16.yaml
 - config.grid_RRFS_CONUScompact_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16.yaml
 - config.grid_RRFS_CONUScompact_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16.yaml
 - config.grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16.yaml
 - config.grid_RRFS_CONUScompact_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16.yaml
 - config.grid_RRFS_SUBCONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16.yaml
 - config.grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16.yaml

These tests could be used to test other capabilities, or be rolled into other existing tests for other capabilities

#### Missing capabilities
There are important capabilities either already present or being introduced soon that are not currently covered by any WE2E test:
 - ~~Retrieving data from AWS~~ (added in #526)
 - ~~Using GDAS and GEFS input data~~ (added in #526)
 - New capabilities being introduced by the RRFS merger effort
 - Retrieving newer GFS data from HPSS (see #510)

Some others may or may not be appropriate for WE2E testing:
 - Plotting scripts
 - New CMAQ capabilities

#### Other changes
[...]

PR #624, opened 2023-02-21
[develop] Update default timesteps (DT_ATMOS) to less wasteful values

mkavulich:
## DESCRIPTION OF CHANGES: 
Based on discussion in last week's code management meeting, it was agreed that the default values for the atmospheric timestep `DT_ATMOS` are too low, and for most cases are very wasteful of core hours. This change increases the default timestep for the following pre-defined domains:

 - RRFS_CONUS_25km, RRFS_CONUScompact_25km: 40 --> 180
 - RRFS_CONUS_13km, RRFS_CONUScompact_13km: 45 --> 75
 - RRFS_NA_13km: 50 --> 75
 - RRFS_CONUScompact_3km, RRFS_SUBCONUS_3km, SUBCONUS_Ind_3km: 40 --> 36
 - CONUS_3km_GFDLgrid: 18 --> 36

There is a known limitation for certain CCPP suites designed for high-resolution cases (per @JeffBeck-NOAA, these are  `FV3_RRFS_v1beta`, `FV3_WoFS_v0`, and `FV3_HRRR`); these suites may see failures if DT_ATMOS is too high. For those cases, the default DT_ATMOS is set to 40, however users can still over-write this value in their config file.

**NOTE: This change *will* affect results for users who are using the affected domains and have not manually specified `DT_ATMOS`. This should be reflected in the release notes for the next release.**

### Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## TESTS CONDUCTED: 
Ran comprehensive set of tests on Hera. Observed some failures due to a known issue with `HRRR` and `RRFS_v1beta` suites (@JeffBeck-NOAA, is this documented somewhere??), so introduced the suite-based default requested by @JeffBeck-NOAA . All comprehensive tests now succeed on Hera as well as Cheyenne (Cheyenne had one known failure and hit a few wallclock limits, but no other failures)

Note: The cost of running the comprehensive test suite on Hera was reduced by ~10% with this change.

- [x] hera.intel
- [x] cheyenne.intel
- [x] comprehensive tests

## DEPENDENCIES:
None

## DOCUMENTATION:
Will update appropriate documentation as part of resolution of #587; since the suite of tests will be overhauled this documentation will be changing anyway so I won't bother wasting time updating it here.

## ISSUE: 
Part of ongoing efforts related to #587

## CHECKLIST

- [x] My changes do not require updates to the documentation (explained above).


natalie-perlin (approved):

ush/predef_grid_params.yaml
```diff
@@ -308,7 +308,7 @@
   ESGgrid_NY: 200
   ESGgrid_PAZI: 0.0
   ESGgrid_WIDE_HALO_WIDTH: 6
-  DT_ATMOS: 40
+  DT_ATMOS: 36
   LAYOUT_X: 5
```
JeffBeck-NOAA, line 312: What was the justification for decreasing the time step here and above?

ush/predef_grid_params.yaml
```diff
@@ -61,7 +61,7 @@
   ESGgrid_NY: 131
   ESGgrid_PAZI: 0.0
   ESGgrid_WIDE_HALO_WIDTH: 6
-  DT_ATMOS: 40
+  DT_ATMOS: 180
   LAYOUT_X: 5
```
JeffBeck-NOAA, line 65: I don't see specific modifications to use smaller time steps when running with RAP/HRRR physics-based SDFs.  If we are counting on users to manually modify the dt_atmos value in their config files when using those SDFs, some people won't understand the need to do that, and the default time step will be too large and fail.  This was the original reason we had time step defaults set to lower values.  I recall you mentioning that there is now a method through which we can automatically decrease the time step when certain SDFs are used?  Can this be implemented?

JeffBeck-NOAA (changes requested):

JeffBeck-NOAA:
In addition to my review, I'll add a cautionary comment that I forgot to mention at last week's code management meeting.  There are certain fast physics processes included within parameterizations (particularly those used in convection-allowing resolutions) which are designed for time steps smaller than ~40-50 seconds.  Certain spatiotemporal relationships in the code require that the physics be called with specific regularity to accurately represent some types of sub-grid-scale feature evolution.  Longer time steps will work, but may likely compromise the accuracy of the model.   Therefore, I would be very wary of using O(100)-second time steps for a 3-km domain.

JeffBeck-NOAA:
In addition to my review, I'll add a cautionary comment that I forgot to mention at last week's code management meeting. There are certain fast physics processes included within parameterizations (particularly those used with convection-allowing resolutions) which are designed for time steps smaller than ~40-50 seconds. Certain spatiotemporal relationships in the code require that the physics be called with specific regularity to accurately represent some types of sub-grid-scale feature evolution. Longer time steps will work, but may likely compromise the accuracy of the model. Therefore, I would be very wary of using O(100)-second time steps for a 3-km domain.

JeffBeck-NOAA:
In addition to my review, I'll add a cautionary comment that I forgot to mention at last week's code management meeting. There are certain fast physics processes included within parameterizations (particularly those used with convection-allowing resolutions) which are designed for time steps smaller than ~40-50 seconds. Certain spatiotemporal relationships in the code require that the physics be called with specific regularity to accurately represent some types of sub-grid-scale feature evolution. Longer time steps will work, but may likely compromise the accuracy of the model. Therefore, I would be very wary of using O(100)-second time steps with a SDF designed for a convection-allowing resolution, regardless of the resolution of the domain in question.

JeffBeck-NOAA:
In addition to my review, I'll add a cautionary comment that I forgot to mention at last week's code management meeting. There are certain fast physics processes included within parameterizations (particularly those used with convection-allowing resolutions) which are designed for time steps smaller than ~40-50 seconds. Certain spatiotemporal relationships in the code require that the physics be called with specific regularity to accurately represent some types of sub-grid-scale feature evolution. Longer time steps will work, but may likely compromise the accuracy of the model. Therefore, I would be very wary of using an O(100)-second time step with an SDF that contains parameterizations designed for a convection-allowing resolution, regardless of the resolution of the domain in question.

ush/predef_grid_params.yaml
mkavulich: This was just for consistency's sake. I probably didn't mention this in last week's meeting, but when I raised the other 3km domains to 40s it caused a segfault, so I figured the other direction was more prudent. We can keep them inconsistent if that's preferable.

mkavulich:
@JeffBeck-NOAA Thanks for your comments. I can implement a time-step limiter for the suites if you can send me a list of suites that are affected...I noticed that the HRRR and RRFS_v1beta suites were affected but are there others? I'd also appreciate documentation to point to (either an issue or elsewhere) if you know of one to point users to get more info on why these suites in particular are treated specially.

mkavulich:
As a follow-up, if this documentation on physics processes that should not use large time steps does not exist, we should create it! This is something that should be transparent to users, especially if it can over-write user settings.

JeffBeck-NOAA:
> As a follow-up, if this documentation on physics processes that should not use large time steps does not exist, we should create it! This is something that should be transparent to users, especially if it can over-write user settings.

Yeah, this is all word-of-mouth info I've gotten previously, and I agree that it should be documented.  The next time I have a chance to talk with some of the physics developers at GSL, I'll try to get some detailed information compiled and written down.

ush/predef_grid_params.yaml
JeffBeck-NOAA: Got it!  It's probably a good idea to keep them consistent.  We also may want to just adopt the final RRFS-based time step once all code has been transitioned to the SRW App.

JeffBeck-NOAA:
> @JeffBeck-NOAA Thanks for your comments. I can implement a time-step limiter for the suites if you can send me a list of suites that are affected...I noticed that the HRRR and RRFS_v1beta suites were affected but are there others? I'd also appreciate documentation to point to (either an issue or elsewhere) if you know of one to point users to get more info on why these suites in particular are treated specially.

The affected suites should be FV3_HRRR, FV3_RRFS_v1beta, FV3_WoFS_v0, and FV3_HRRR.  I thought FV3_RAP was added recently, but I'm not seeing it in valid_param_vals.yaml?  That SDF may also be affected, but we'd need to talk with someone like Joe Olson to confirm.

tests/WE2E/run_WE2E_tests.py
```diff
@@ -175,8 +175,8 @@ def run_we2e_tests(homedir, args) -> None:
 
         if 'verification' in test_cfg:
             logging.debug(test_cfg['verification'])
```
gsketefian: @mkavulich As things stand now, a "verification" section does not have to be included in the test configuration file in order to run a test with verification, but such a section has to exist in the final `config.yaml` file that the generate script reads in because it contains the variable `VX_FCST_INPUT_BASEDIR`, which must be specified for verification (and the call to `check_task_verification` sets this variable).  Your change will still work with all the current WE2E vx tests because they all contain a "verification" section, but if that section is removed from any of the tests because it contains only optional settings (e.g. in the test `MET_verification_only_vx` the section only sets `VX_FCST_MODEL_NAME`, which is optional), the test will fail (I just ran the test both ways myself).  That's why I originally didn't include the call to `check_task_verification` under the if-statement.  This situation may change with upcoming vx PRs, but for now it is safer to always call `check_task_verification`.

mkavulich: @gsketefian Thanks for the further info, but I am a bit confused at how this new capability is implemented. Is there no flag for running "verification only"? It appears as if this flag is needed: The reason I made this change is because on Cheyenne (and I presume other non-Hera machines) the `deactivate_tasks` test is now failing due to this path not being set for those platforms.

gsketefian: @mkavulich Currently, there is no single flag that turns on/off vx tasks.  Instead, there are the four flags `RUN_TASK_VX_GRIDSTAT`, `RUN_TASK_VX_POINTSTAT`, `RUN_TASK_VX_ENSGRID`, and  `RUN_TASK_VX_ENSPOINT`.  The ultimate goal is to have one flag for deterministic verification and another for ensemble verification.  But I'm a couple of PRs from getting to that point.

The staged forecasts for the vx tasks are only available on Hera.  If I understand you correctly, the `deactivate_tasks` test is failing because `TEST_VX_FCST_INPUT_BASEDIR` is not defined in the machine files of platforms other than Hera.  I think the easy solution is to put those in all the machine files and set it to a blank string for now.  I forgot to do that in one of my PRs.  Could you try that and see if it solves the issue with `deactivate_tasks`?  Thanks.

mkavulich: @gsketefian I think you are missing my point: this "TEST_VX_FCST_INPUT_BASEDIR" needs to be set only in the case where the user is doing a VX-only run, correct? So shouldn't there be a flag for VX-only runs? Because VX-only requires a completely different process for setting up directories in the VX tasks, unless I'm mistaken.

JeffBeck-NOAA:
Hey @mkavulich, I see two separate if statements in your latest commit that both address dt_atmos > 40.  What use cases are they for?  Is one for the automated selection of the pre-defined domain, and the other if the user manually sets dt_atmos in their config file?

mkavulich:
@JeffBeck-NOAA That's correct: the first statement is if the user has set DT_ATMOS in their config file, the second is if DT_ATMOS is coming from the predefined domain parameters.

tests/WE2E/run_WE2E_tests.py
gsketefian: @mkavulich Yes, `TEST_VX_FCST_INPUT_BASEDIR` needs to be set only when doing a test that runs only vx.  Testing for that requires obtaining values of other flags, specifically `RUN_TASK_RUN_FCST`, `WRITE_DOPOST`, and `RUN_TASK_RUN_POST`.  That's because the check for whether a vx-only run is to see whether UPP is going to be run (either inline or separately) because the forecast inputs for vx tasks are the output files from UPP.  If so, then it's not a vx-only experiment.  Otherwise, it is.  For example, in `config_defaults.yaml`, `VX_FCST_INPUT_BASEDIR` is set like this:
```
VX_FCST_INPUT_BASEDIR: '{{ workflow.EXPTDIR if ((workflow_switches.RUN_TASK_RUN_FCST and task_run_fcst.WRITE_DOPOST) or workflow_switches.RUN_TASK_RUN_POST) }}'
```
We could implement such a check in `run_WE2E_tests.py`, but a lot more vx changes are coming in the next few vx PRs, I decided it's much easier to always set `VX_FCST_INPUT_BASEDIR` by calling `check_task_verification` outside this if-statement.  That in turn requires `TEST_VX_FCST_INPUT_BASEDIR` to at least be defined in the machine files (it should be ok to set it to an empty string), which is something I forgot about in my PR.

I guess we could create a new derived variable/flag for vx-only runs, but that seems redundant since the info can be obtained from `RUN_TASK_RUN_FCST`, `WRITE_DOPOST`, and `RUN_TASK_RUN_POST`.  It would be fine as an internal local variable in `run_WE2E_tests.py` if you want to implement the checks above for when `TEST_VX_FCST_INPUT_BASEDIR` is really necessary, but we should probably just define `TEST_VX_FCST_INPUT_BASEDIR` in all the machine files anyway for completeness (setting it to null strings on platforms on which the forecast files are not staged).

gsketefian: Another easy fix for now is to set `TEST_VX_FCST_INPUT_BASEDIR` to a null string if it's not defined in the machine file (instead of erroring out).

mkavulich: Maybe it's best to take this conversation over to your latest PR; I'll remove my change from this one.

JeffBeck-NOAA:
> @JeffBeck-NOAA That's correct: the first statement is if the user has set DT_ATMOS in their config file, the second is if DT_ATMOS is coming from the predefined domain parameters.

Thanks, @mkavulich!  This looks good!

JeffBeck-NOAA (approved):

MichaelLueken (approved):
@mkavulich These changes look good to me!  Thanks for getting these changes in so quickly to help reduce the use of core hours on the machines!

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1249350329/20230222202016/ufs-srweather-app
Build was Successful
Rocoto jobs started
Long term tracking will be done on 10 experiments
ci-hera-intel-WE
Experiment Succeeded on hera: pregen_grid_orog_sfc_climo
Experiment Succeeded on hera: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_2017_gfdlmp_regional_plot
Experiment Succeeded on hera: community_ensemble_2mems_stoch
Experiment Succeeded on hera: MET_ensemble_verification
All experiments completed

mkavulich:
@MichaelLueken I am assuming these Jenkins failures are still due to the Hera problem? Let me know if there's anything I need to address.

MichaelLueken:
@mkavulich No issues with Hera, the machine is no longer being called in the Jenkinsfile.  Due to Orion being down for maintenance, I did need to abort the tests for that machine, which would lead to a message saying that a job was aborted.

Jenkins failed on Gaea due to a disconnect on the machine.  Looking through the experiments, all of the tests successfully passed.  The experiment directories can be found:

`/lustre/f2/dev/wpo/role.epic/jenkins/workspace/fs-srweather-app_pipeline_PR-624/expt_dirs`

Jenkins also failed on Jet.  Looking through the experiments, I'm seeing that make_ics, make_lbcs, and run_fcst tasks were failing.  The logs for these tasks show the following:

```
FATAL ERROR:
ERROR:
  From script:  "load_modules_run_task.sh"
  Full path to script:  "/mnt/lfs4/HFIP/hfv3gfs/role.epic/jenkins/workspace/fs-srweather-app_pipeline_PR-624/ush/load_modules_run_task.sh"
  Loading of platform- and compiler-specific module file (BUILD_MOD_FN)
for the workflow task specified by task_name failed:
  task_name = "make_ics"
  BUILD_MOD_FN = "build_jet_intel"
Exiting with nonzero status.
```
```
FATAL ERROR:
ERROR:
  From script:  "load_modules_run_task.sh"
  Full path to script:  "/mnt/lfs4/HFIP/hfv3gfs/role.epic/jenkins/workspace/fs-srweather-app_pipeline_PR-624/ush/load_modules_run_task.sh"
  Loading of platform- and compiler-specific module file (BUILD_MOD_FN)
for the workflow task specified by task_name failed:
  task_name = "make_lbcs"
  BUILD_MOD_FN = "build_jet_intel"
Exiting with nonzero status.
```
```
FATAL ERROR:
ERROR:
  From script:  "load_modules_run_task.sh"
  Full path to script:  "/mnt/lfs4/HFIP/hfv3gfs/role.epic/jenkins/workspace/fs-srweather-app_pipeline_PR-624/ush/load_modules_run_task.sh"
  Loading of platform- and compiler-specific module file (BUILD_MOD_FN)
for the workflow task specified by task_name failed:
  task_name = "run_fcst"
  BUILD_MOD_FN = "build_jet_intel"
Exiting with nonzero status.
```
Some of the logs include the following as well:

```
+ load_modules_run_task.sh[131]: module load build_jet_intel
++ bash[82]: /apps/lmod/8.5.2/libexec/lmod bash load build_jet_intel
Lmod has detected the following error: The following module(s) are unknown:
"sutils"

Please check the spelling or version number. Also try "module spider ..."
It is also possible your cache file is out-of-date; it may help to try:
  $ module --ignore_cache load "sutils"

Also make sure that all modulefiles written in TCL start with the string
#%Module
```

The experiment directories on Jet can be found:

`/mnt/lfs4/HFIP/hfv3gfs/role.epic/jenkins/workspace/fs-srweather-app_pipeline_PR-624/expt_dirs`

It is important to note that, while this behavior wasn't seen during the previous Jenkins tests at the end of last week, this is being seen on not only your branch, but also @panll and @gsketefian branches are encountering this as well.  Maybe something happened during the maintenance on Tuesday that is causing issues now, I'm not sure.  Having said that, running tests manually, if `rocotoboot` is used to relaunch the failed tasks, they run without issue.

Since the Gaea tests passed and the Jet tests can pass with a little extra work, I'm okay moving forward with these changes.  Having said that, I'm interested to see if you have encountered issues on Jet following Tuesday's maintenance.

MichaelLueken:
@mkavulich This morning, I was able to manually submit the fundamental tests on Jet without the issues encountered yesterday.  I have resubmitted the Jenkins tests on Jet to see if they will pass now.

MichaelLueken:
@mkavulich The retest on Jet successfully passed.  While resubmitting the Jet test, the Orion test relaunched as well, which is why the aborted message is now shown.

```
======================================
Checking workflow status of experiment "custom_ESGgrid" ...
Workflow status:  SUCCESS
======================================

======================================
Checking workflow status of experiment "custom_GFDLgrid" ...
Workflow status:  SUCCESS
======================================

======================================
Checking workflow status of experiment "grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2" ...
Workflow status:  SUCCESS
======================================

======================================
Checking workflow status of experiment "grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16" ...
Workflow status:  SUCCESS
======================================

======================================
Checking workflow status of experiment "grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_RAP_suite_HRRR" ...
Workflow status:  SUCCESS
======================================

======================================
Checking workflow status of experiment "grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v15p2" ...
Workflow status:  SUCCESS
======================================

======================================
Checking workflow status of experiment "nco_grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR" ...
Workflow status:  SUCCESS
======================================

======================================
Checking workflow status of experiment "specify_DOT_OR_USCORE" ...
Workflow status:  SUCCESS
======================================

======================================
Checking workflow status of experiment "specify_DT_ATMOS_LAYOUT_XY_BLOCKSIZE" ...
Workflow status:  SUCCESS
======================================

======================================
Checking workflow status of experiment "specify_RESTART_INTERVAL" ...
Workflow status:  SUCCESS
======================================
```

Will now move forward with merging this work.

Issue #619, opened 2023-02-17
Rename vx tasks so they follow METplus-tool based convention


gsketefian:
## Description

To avoid various sorts of file conflicts (e.g. multiple workflow tasks writing to the same output file), the eventual goal for the verification (vx) tasks in the ROCOTO workflow is for each task to only call one METplus tool (e.g. `GridStat`, `PointStat`, `EnsembleStat`, etc) and operate on one field or group of fields (see Issue #620 ).  For clarity, we would like the task names along with the corresponding j-job, ex-script, and log file names to reflect this.  

## Solution

In this issue, I propose that each item (task or file name) be renamed so that it contains the name of the METplus tool the task runs as well as the name of the field or group of fields on which it is run (e.g. 1-hour precip (APCP01h), reflectivity (REFC), etc).  

I propose to rename the j-job scripts as follows (note that NCO requires that all j-job names be in upper case):
```
JREGIONAL_RUN_VX_GRIDSTAT       --->  JREGIONAL_RUN_MET_GRIDSTAT_VX
JREGIONAL_RUN_VX_POINTSTAT      --->  JREGIONAL_RUN_MET_POINTSTAT_VX
JREGIONAL_RUN_VX_ENSGRID        --->  JREGIONAL_RUN_MET_ENSEMBLESTAT_VX_GRID
JREGIONAL_RUN_VX_ENSPOINT       --->  JREGIONAL_RUN_MET_ENSEMBLESTAT_VX_POINT
JREGIONAL_RUN_VX_ENSGRID_MEAN   --->  JREGIONAL_RUN_MET_GRIDSTAT_VX_ENSMEAN
JREGIONAL_RUN_VX_ENSPOINT_MEAN  --->  JREGIONAL_RUN_MET_POINTSTAT_VX_ENSMEAN
JREGIONAL_RUN_VX_ENSGRID_PROB   --->  JREGIONAL_RUN_MET_GRIDSTAT_VX_ENSPROB
JREGIONAL_RUN_VX_ENSPOINT_PROB  --->  JREGIONAL_RUN_MET_POINTSTAT_VX_ENSPROB
```
I propose to rename the ex-scripts as follows (note that NCO requires that all ex-script names be in lower case):
```
exregional_run_vx_ensgrid.sh        --->  exregional_run_met_ensemblestat_vx_grid.sh
exregional_run_vx_enspoint.sh       --->  exregional_run_met_ensemblestat_vx_point.sh
[...]

PR #631, opened 2023-02-28
[develop] Add RRFS product generation job to the SRW app

### LINKED ISSUE #580: Add product generation job.
The script for this one is [here](https://github.com/NOAA-GSL/regional_workflow/blob/feature/RRFS_dev1/scripts/exregional_run_prdgen.sh) in RRFS_dev1.

This task uses wgrib2 to generate various subdomain RRFS grib2 files from the raw-UPP generated grib2 output from the run_post task.  It will be needed for the real-time operational RRFS runs.  

The rrfs_prdgen_subpiece.sh script is called using CFP to parallelize the job and speed up the overall runtime.  The current version of the script resides in $PRDGENdir (ush/prdgen) but we may want to consider re-locating it.

BenjaminBlake-NOAA:
## DESCRIPTION OF CHANGES: 
This pull request adds the RRFS product generation (prdgen) job to the SRW app.  The prdgen job uses wgrib2 to generate various subdomain RRFS grib2 files from the raw-UPP generated grib2 output from the run_post task.  It will be needed for the real-time operational RRFS runs, which will be over the RRFS_NA_3km grid.  In the config.yaml file, two new configuration flags have been added.  RUN_TASK_RUN_PRDGEN needs to be set to true if you want to run the prdgen job, and DO_PARALLEL_PRDGEN can be set to true if you are running over the RRFS_NA_3km grid and want to use CFP to speed up the runtime of this task.  Two ex-scripts have also been added - exregional_run_prdgen.sh, which is the main script, and exregional_run_prdgen_subpiece.sh, which is only called if DO_PARALLEL_PRDGEN = TRUE.  In addition, a new WE2E test was added to the wflow_features directory for running the prdgen job over the RRFS_NA_3km grid.

A couple of notes about this task:
1. A testbed text file is optionally needed for this job to create testbed files.  In the RRFS workflow, it is included in sorc/UPP/parm, which gets linked into parm/upp.
2. RUN_CMD_PRDGEN has been added to all the machine.yaml files, but may need to be modified in the future for all machines besides WCOSS2.  RUN_CMD_PRDGEN is invoked when DO_PARALLEL_PRDGEN = TRUE and therefore will need to include a call to cfp.
3. If DO_PARALLEL_PRDGEN = TRUE with the RRFS_NA_3km grid, then PPN_RUN_PRDGEN needs to be 22 and not 1.

### Type of change

- [X ] New feature (non-breaking change which adds functionality)
- [X ] This change requires a documentation update

## TESTS CONDUCTED: 

- [X ] hera.intel
- [ ] orion.intel
- [ ] cheyenne.intel
- [ ] cheyenne.gnu
- [ ] gaea.intel
- [ ] jet.intel
- [X ] wcoss2.intel
- [ ] Jenkins
- [ ] fundamental test suite
- [X ] comprehensive tests (grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16, grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_RAP_suite_HRRR, MET_verification, community_ensemble_2mems_stoch, prdgen)

## DEPENDENCIES:
N/A

## DOCUMENTATION:
The documentation will need to be updated but I don't have any experience with doing this myself.

## ISSUE: 
Fixes issue [#580 ](https://github.com/ufs-community/ufs-srweather-app/issues/580)

## CHECKLIST

- [X ] My code follows the style guidelines in the Contributor's Guide
- [X ] I have commented my code, particularly in hard-to-understand areas
- [X ] My changes generate no new warnings
- [X ] New and existing tests pass with my changes

## LABELS (optional): 

- [ ] Work In Progress
- [ ] bug
- [X ] enhancement
- [ ] documentation
- [ ] release
- [ ] high priority
- [ ] run_ci
- [ ] Needs Cheyenne test 
- [ ] Needs Jet test 
- [ ] Needs Hera test 
- [ ] Needs Orion test 
- [X ] help wanted

## CONTRIBUTORS (optional): 
@hu5970, @ShunLiu-NOAA, @danielabdi-noaa


jobs/JREGIONAL_RUN_PRDGEN
```diff
@@ -0,0 +1,140 @@
...
+#
+# Set the run directory.
+#
+#-----------------------------------------------------------------------
+#
+CYCLE_TYPE=${CYCLE_TYPE:-prod}
+if [ ${CYCLE_TYPE} == "spinup" ]; then
+  DATA="${DATA:-${COMIN}${SLASH_ENSMEM_SUBDIR}}"
+else
+  DATA="${DATA:-${COMIN}${SLASH_ENSMEM_SUBDIR}}"
```
christinaholtNOAA: Should these two be the same? If so, then can the if statement be removed?

parm/FV3LAM_wflow.xml
```diff
@@ -753,6 +760,49 @@ the <task> tag to be identical to the ones above for other output times.
...
+      <envar><name>GLOBAL_VAR_DEFNS_FP</name><value>&GLOBAL_VAR_DEFNS_FP;</value></envar>
+      <envar><name>USHdir</name><value>&USHdir;</value></envar>
+      <envar><name>PDY</name><value><cyclestr>@Y@m@d</cyclestr></value></envar>
+      <envar><name>cyc</name><value><cyclestr>@H</cyclestr></value></envar>
+      <envar><name>subcyc</name><value><cyclestr>@M</cyclestr></value></envar>
+      <envar><name>LOGDIR</name><value>&LOGDIR;</value></envar>
+      <envar><name>SLASH_ENSMEM_SUBDIR</name><value><cyclestr>{{ slash_ensmem_subdir }}</cyclestr></value></envar>
+      <envar><name>ENSMEM_INDX</name><value><cyclestr>#{{ ensmem_indx_name }}#</cyclestr></value></envar>
+
+      <envar><name>CYCLE_TYPE</name><value><cyclestr>prod</cyclestr></value></envar>
```
christinaholtNOAA: Would this be set differently, or is it always `CYCLE_TYPE=prod`? If it is more or less always the same, does it make sense to move it to be hard-coded in the ex-script? Is it an NCO requirement to be exported?

I suppose the same questions hold for `CYCLE_SUBTYPE` and `TMMARK` below.

scripts/exregional_run_prdgen.sh
```diff
@@ -0,0 +1,437 @@
...
+# 4 subpieces for CONUS and Alaska grids
+sed -n -e '1,250p' $DATAprdgen/prslevf${fhr}.txt >& $DATAprdgen/conus_ak_1.txt
+sed -n -e '251,500p' $DATAprdgen/prslevf${fhr}.txt >& $DATAprdgen/conus_ak_2.txt
+sed -n -e '501,750p' $DATAprdgen/prslevf${fhr}.txt >& $DATAprdgen/conus_ak_3.txt
+sed -n -e '751,$p' $DATAprdgen/prslevf${fhr}.txt >& $DATAprdgen/conus_ak_4.txt
+
+# 2 subpieces for Hawaii and Puerto Rico grids
+sed -n -e '1,500p' $DATAprdgen/prslevf${fhr}.txt >& $DATAprdgen/hi_pr_1.txt
+sed -n -e '501,$p' $DATAprdgen/prslevf${fhr}.txt >& $DATAprdgen/hi_pr_2.txt
+
```
christinaholtNOAA, line 214: Would a loop over the following section work to reduce repetition of information? I was thinking something like (warning! totally untested bash):
```
echo "#!/bin/bash" > $DATAprdgen/poescript_${fhr}
echo "export DATA=${DATAprdgen}" >> $DATAprdgen/poescript_${fhr}
echo "export COMOUT=${COMOUT}" >> $DATAprdgen/poescript_${fhr}

tasks=(4 4 2 2 10)
domains=(conus ak hi pr namerica)
for i in $(seq ${#domains[@]}), do
  domain=${domains[$i]}
  for task in $(seq ${tasks[$i]}); do 
    mkdir -p $DATAprdgen/prdgen_${domain}_${task}
    echo "$SCRIPTSdir/exregional_run_prdgen_subpiece.sh $fhr $cyc $task $domain ${DATAprdgen} ${COMOUT} &" >> $DATAprdgen/poescript_${fhr}
    
  done
done
echo "wait" >> $DATAprdgen/poescript_${fhr}
chmod 775 $DATAprdgen/poescript_${fhr}
```

scripts/exregional_run_prdgen.sh
```diff
@@ -0,0 +1,437 @@
...
+eval ${RUN_CMD_PRDGEN} ${CMDFILE} ${REDIRECT_OUT_ERR} || print_err_msg_exit "\
+Call to parallel prdgen for forecast hour $fhr returned with non-
+zero exit code."
+POST_STEP
+
+#----------------------------------------
+# reassemble the output grids
+#----------------------------------------
+
+cat $DATAprdgen/prdgen_conus_1/conus_1.grib2 \
```
christinaholtNOAA: A similar loop could be applied in this section, too.

scripts/exregional_run_prdgen.sh
```diff
@@ -0,0 +1,437 @@
...
+fi
+
+# replace fhr with subh_fhr
+echo "fhr=${fhr} and subh_fhr=${subh_fhr}"
+fhr=${subh_fhr}
+
+#
+# background files
+#
+prslev=${COMOUT}/${NET}.${cycle}${dot_ensmem}.prslev.f${fhr}.${POST_OUTPUT_DOMAIN_NAME}.grib2
```
christinaholtNOAA, line 142: Where does `dot_ensmem` get set? Apologies if I missed that one.

scripts/exregional_run_prdgen.sh
```diff
@@ -0,0 +1,437 @@
...
+      wait 
+
+      # Remove temporary files
+      rm -f ${subdir}/${fhr}/tmp_${grid}.grib2
+
+      # Save to com directory 
+      mkdir -p ${COMOUT}/${grid}_grid
+      cp_vrfy ${bg_remap} ${COMOUT}/${grid}_grid/${NET}.${cycle}${dot_ensmem}.${leveltype}.f${fhr}.${POST_OUTPUT_DOMAIN_NAME}.grib2
+
+      if [ $leveltype = 'prslev' ]; then
```
christinaholtNOAA: I may have overlooked it, but each of these if statements contain roughly the same information. Would this one command work to replace them all? I've replaced the hardcoded `prslev` with `${leveltype}` instead.

```
         ln_vrfy -fs --relative ${COMOUT}/${grid}_grid/${NET}.${cycle}${dot_ensmem}.${leveltype}.f${fhr}.${POST_OUTPUT_DOMAIN_NAME}.grib2 ${COMOUT}/${net4}.${cycle}.${leveltype}.f${fhr}.${gridname}grib2
         wgrib2 ${COMOUT}/${net4}.${cycle}.prslev.f${fhr}.${gridname}grib2 -s > ${COMOUT}/${net4}.${cycle}.${leveltype}.f${fhr}.${gridname}grib2.idx
```

scripts/exregional_run_prdgen_subpiece.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# Use different parm file for each subpiece
+wgrib2 $comout/rrfs.t${cyc}z.prslev.f${fhr}.grib2 | grep -F -f ${parmfile} | wgrib2 -i -grib inputs.grib${domain} $comout/rrfs.t${cyc}z.prslev.f${fhr}.grib2
+wgrib2 inputs.grib${domain} -new_grid_vectors "UGRD:VGRD:USTM:VSTM" -submsg_uv inputs.grib${domain}.uv
+wgrib2 inputs.grib${domain}.uv -set_bitmap 1 -set_grib_type ${compress_type} \
+  -new_grid_winds grid -new_grid_vectors "UGRD:VGRD:USTM:VSTM" \
+  -new_grid_interpolation neighbor \
+  -if ":(WEASD|APCP|NCPCP|ACPCP|SNOD):" -new_grid_interpolation budget -fi \
+  -new_grid ${gridspecs} ${domain}_${subpiece}.grib2
+
```
christinaholtNOAA, line 58: I think the lines below this point may be unnecessary.

tests/WE2E/test_configs/wflow_features/config.prdgen.yaml
```diff
@@ -0,0 +1,43 @@
...
+    completes successfully on the RRFS_NA_3km grid using the RRFS_v1beta
+    physics suite with ICs and LBCs derived from the FV3GFS.
+    Note that this test also sets various resource parameters for several
+    of the rocoto tasks in order to more efficiently run the code on this
+    (very large) grid.
+user:
+  RUN_ENVIR: community
+workflow:
+  CCPP_PHYS_SUITE: FV3_RRFS_v1beta
+  PREDEF_GRID_NAME: RRFS_NA_3km
```
christinaholtNOAA: That's an expensive test! Could we leverage any of the existing tests (if we already have a NA domain test) to also test this task? Could we alternatively test just the prodgen task with staged data?

scripts/exregional_run_prdgen_subpiece.sh
```diff
@@ -0,0 +1,63 @@
...
+#export err=$?; err_chk
```
RatkoVasic-NOAA: Is this supposed to be commented out?

ush/machine/odin.yaml
```diff
@@ -9,8 +9,11 @@ platform:
   QUEUE_FCST: workq
   PARTITION_HPSS: workq
   QUEUE_HPSS: workq
+  PARTITION_PRDGEN: workq
+  QUEUE_PRDGEN: workq
```
christinaholtNOAA: Would it work for the prdgen task to use `QUEUE_` and `PARTITION_DEFAULT` instead of introducing a new queue? Are there platforms that may need a totally separate queue to do the thing for this job that's different than others?

I think we've accumulated 3 so far for a catchall (DEFAULT), the forecast (because we may want beefier nodes than a default on the cloud, or on Jet), and then HPSS for access nodes that have connectivity to the internet.

christinaholtNOAA (commented):
@BenjaminBlake-NOAA This is great! Thanks for preparing this. I have just a few suggestions and questions below.

chan-hoo:
@BenjaminBlake-NOAA, 1) Please merge the `develop` branch into your branch. There are some conflicts with the latest develop branch. 2) What do you think about changing the task name from `run_prdgen` to `gen_product` or similar?

ShunLiu-NOAA:
@BenjaminBlake-NOAA and @chan-hoo, Good suggestion about the task name. It is reasonable to remove "run_" in job name. Maybe we can just use "prdgen". We also consider removing "run_" for "prestart", "post", "fcst", "enkf" and "analysis" job in regional workflow in RRFS_DEV1.

BenjaminBlake-NOAA:
@chan-hoo  I just merged all the changes from the develop branch into my feature branch, hopefully I did that right.  I agree with @ShunLiu-NOAA and think we could call this task the "prdgen" task instead of "run_prdgen".  The operational NAM has a prdgen task.  @MatthewPyle-NOAA  what are your thoughts about the name of this task?

@christinaholtNOAA and @RatkoVasic-NOAA , thank you for your suggestions and feedback!  They all sound quite reasonable to me and I will try and address as many of them as I can before I go on leave (tomorrow through next Tuesday).

JeffBeck-NOAA:
> @BenjaminBlake-NOAA and @chan-hoo, Good suggestion about the task name. It is reasonable to remove "run_" in job name. Maybe we can just use "prdgen". We also consider removing "run_" for "prestart", "post", "fcst", "enkf" and "analysis" job in regional workflow in RRFS_DEV1.

@gsketefian, thoughts on this?

scripts/exregional_run_prdgen_subpiece.sh
BenjaminBlake-NOAA: Yes it is not needed, I will remove the line.

scripts/exregional_run_prdgen_subpiece.sh
BenjaminBlake-NOAA: I agree, I will remove them.

ush/machine/odin.yaml
BenjaminBlake-NOAA: Yes I think it should.  I will use QUEUE_DEFAULT and PARTITION_DEFAULT for this task.

tests/WE2E/test_configs/wflow_features/config.prdgen.yaml
BenjaminBlake-NOAA: There is one WE2E test for the RRFS_NA_3km grid in the grids_extrn_mdls_suites_community directory.  It is called grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta, so it is very similar to the one I added here.  I'm willing to add the prdgen job to this pre-existing WE2E test if that makes sense and I can remove this one.  What do you think?

gsketefian:
@JeffBeck-NOAA I like having task names with verbs in them, like "run_this_or_that" or "gen_something" (generate) because it allows for a more descriptive/informative name.  But I suppose they can all be nouns, like "fcst" instead of "run_fcst", as long as the naming convention is consistent (verb-like or noun-like).  But I suspect this path will lead to very short names that sound cryptic to a community user of the SRW who's not involved in the day-to-day weeds of development.  In this particular case, for now I'd go with `gen_product` as @chan-hoo suggests to be consistent with how the other tasks are named (verb-like).  If task renaming is something people are interested in discussing, we should do so at the next code management meeting and/or open an issue in the repo.

JeffBeck-NOAA:
Thanks, @gsketefian!  I agree that we should keep it consistent across task names and maintain a descriptive quality for those who aren't as familiar with the workflow.

chan-hoo:
@JeffBeck-NOAA, @gsketefian, I basically agree with you about the consistent and readable naming. However, the use of the prefix `run_` is questionable to me. The task switch names in `config_default.yaml` start with `run_task_`. If the task name begins with `run_`, the switch name becomes `run_task_run_`. This looks a little bit weird to me. We know that the tasks will run without the prefix `run_`. This is the reason why all the extra tasks for air quality modeling (Online-CMAQ) do not have a verb on them. I agree that `prdgen`, `enkf`, `analysis`, and even `fcst` look too short.

jobs/JREGIONAL_RUN_PRDGEN
MichaelLueken: It looks like line 67 be:

DATA="${DATA:-${COMIN}${SLASH_ENSMEM_SUBDIR}}/fcst_fv3lam_spinup"

while line 69 should be:

DATA="${DATA:-${COMIN}${SLASH_ENSMEM_SUBDIR}}/fcst_fv3lam".

scripts/exregional_run_prdgen.sh
MichaelLueken: @christinaholtNOAA It looks like `dot_ensmem` is set in `ush/job_preamble.sh` in the current `develop` branch:

https://github.com/ufs-community/ufs-srweather-app/blob/b34ab05a13d1fc571afd5f99f790e5ec71bc9a14/ush/job_preamble.sh#L27-L31

scripts/exregional_run_prdgen.sh
BenjaminBlake-NOAA: Yes, I think something like this should work!  Good idea.  I can test it next week after I return from leave.

scripts/exregional_run_prdgen.sh
BenjaminBlake-NOAA: Agreed, will look into doing that.

gsketefian:
@chan-hoo Yes, I agree `run_task_run_fcst` is a bit strange.  The first `run_` can be changed to something else, like `incl_` for "include" or `act_` or `actv_` for "activate" a task in the workflow, so those flags would be renamed `incl_task_run_fcst` or `actv_task_run_fcst`.

chan-hoo:
@gsketefian, @JeffBeck-NOAA, what do you think about the consistency to 'noun-like' naming? If we put the prefix `run_` to the scripts, the majority of the tasks for RRFS and AQM will have it. I think it will make the task name too long. Many other applications do not use a verb on the file names of their scripts. For example, you can check the global workflow (https://github.com/NOAA-EMC/global-workflow/tree/develop/scripts). Even in the example of the j-job or ex-script in the NCO standards, the file name is `JPMB_FORECAST` or `expmb_forecast.sh` (I don't think it is a requirement though). In my opinion, it would be good to have consistency with other applications. @MatthewPyle-NOAA, @aerorahul, can you let us know how other applications name the j-job and ex- script files? I'd like to hear from others too; @MichaelLueken, @mkavulich, @christinaholtNOAA, @danielabdi-noaa, @ShunLiu-NOAA, @hu5970.

MatthewPyle-NOAA:
@chan-hoo I'd say there isn't a single standard, but most follow something close to this idea.  For a production model system we'll call "ABC":

J-job file name would start with a `JABC_` string
ex-script file names would have a format like `exabc_*.sh`

There will be a correspondence between the J-jobs and ex-scripts, such as:

JABC_POST would execute something like exabc_post.sh

Personally I think the inclusion of "run" in task names is superfluous.

JeffBeck-NOAA:
@MatthewPyle-NOAA, @chan-hoo, these are fair points.  I think "noun-like" naming would be fine.  The only comment I'd add is that, not being familiar with the global workflow, when I see "JPMB_FORECAST", I don't know what "pmb" stands for.  So, if possible, it'd be good to have a more descriptive name, without making the name too long, as @chan-hoo mentioned.  Our current configuration has "JREGIONAL" and "exregional" for the j-jobs and ex-scripts prefixes, which I think follows the general standard.

aerorahul:
> @chan-hoo I'd say there isn't a single standard, but most follow something close to this idea. For a production model system we'll call "ABC":
> 
> J-job file name would start with a `JABC_` string ex-script file names would have a format like `exabc_*.sh`
> 
> There will be a correspondence between the J-jobs and ex-scripts, such as:
> 
> JABC_POST would execute something like exabc_post.sh
> 
> Personally I think the inclusion of "run" in task names is superfluous.

The convention is `JRUN_`  where `RUN` is the "name of the model run (third level of the COM directory structure).
If a j-job serves multiple values for "RUN", then something common e.g. `JGLOBAL_`, instead of `JGDAS_` or `JGFS_` is used.

Each `JRUN_` has a corresponding `exRUN_`.  Same logic applies when an ex-script serves multiple values for "RUN"

The top-level job card `ecf` driver of the `JRUN_` scripts is named uniquely e.g. `jRUN_.ecf`  There is a unique job card `ecf` driver for each value of `RUN`.

Please see Section VII of the EE2 standards.
"All examples are for job jpmb_forecast. Model name is nco and type of model run is pmb"
This statement implies `RUN = pmb`

gsketefian:
@chan-hoo If you're proposing changing to noun-like task name, can you create an issue (like the one I did for renaming of vx tasks, #619) so we can see exactly what the new names will be of each task?  There are already issues that come to my mind.  For example, if `get_extrn_ics` becomes `extrn_ics` and `make_ics` becomes `ics`, it will not be clear (especially to a relatively inexperienced user) what the difference between these two is.  I suspect that after considering all the tasks, it's going to be difficult to stick to a strictly noun-like or verb-like naming convention AND maintain clarity AND have names that aren't too long, so it may end up being a mix of nouns and verbs.  And I strongly suggest doing this in a separate PR, since it will be a significant change.

chan-hoo:
@gsketefian, sure, I'll open an issue for this soon.

jobs/JREGIONAL_RUN_PRDGEN
BenjaminBlake-NOAA: I think eventually that will be the plan once more RRFS tasks/features are added, but for now I'll remove this if statement.

parm/FV3LAM_wflow.xml
BenjaminBlake-NOAA: These variables can be revisited in a later PR, I'll remove them for this one.

scripts/exregional_run_prdgen.sh
BenjaminBlake-NOAA: Yes I agree.  I'm about to test a new version of the script with this change and the ones you recommended above with adding loops.  Once that's done I'll commit the changes and mark these comments as resolved.

scripts/exregional_run_prdgen.sh
BenjaminBlake-NOAA: Just committed a revised version of the script.

tests/WE2E/test_configs/wflow_features/config.prdgen.yaml
BenjaminBlake-NOAA: I added the prdgen job to the pre-existing WE2E task for the RRFS_NA_3km grid.

jobs/JREGIONAL_RUN_PRDGEN
```diff
@@ -0,0 +1,127 @@
...
+# e.g. from the workflow task that runs this script failing and then being 
+# called again.  Thus, we first make sure preexisting versions are deleted.
+#
+#-----------------------------------------------------------------------
+#
+if [ "${RUN_ENVIR}" = "nco" ]; then
+  export COMOUT="${COMOUT}"
+else
+  export COMOUT="${DATA}/postprd"
+fi
```
christinaholtNOAA, line 86: This could be simplified to:

```
if [ "${RUN_ENVIR}" = "community" ]; then
  export COMOUT="${DATA}/postprd"
fi
mkdir_vrfy -p "${COMOUT}"
```

parm/FV3LAM_wflow.xml
```diff
@@ -1188,6 +1189,46 @@ the <task> tag to be identical to the ones above for other output times.
   {%- endif %}
 {%- endif %}
 {%- endif %}
+{%- if run_task_run_prdgen %}
```
christinaholtNOAA, line 1192: Just checking...is this only meant to run for the hourly graphics (as opposed to subhourly)? _Yes_ is totally fine, just curious.

scripts/exregional_run_prdgen.sh
```diff
@@ -0,0 +1,366 @@
...
+POST_STEP
+
+#----------------------------------------
+# reassemble the output grids
+#----------------------------------------
+
+tasks=(4 4 2 2 10)
+domains=(conus ak hi pr namerica)
+count=0
+for i in ${domains[@]}
```
christinaholtNOAA: It's a little more readable with 

```
for domain in ${domains[@]}
```

scripts/exregional_run_prdgen.sh
```diff
@@ -0,0 +1,366 @@
...
+if [ ${DO_PARALLEL_PRDGEN} == "TRUE" ]; then
+#
+#  parallel run wgrib2 for product generation
+#
+
+if [ ${PREDEF_GRID_NAME} = "RRFS_NA_3km" ]; then
+
+DATA=$COMOUT
+DATAprdgen=$DATA/prdgen_${fhr}
+mkdir $DATAprdgen
```
christinaholtNOAA: There are several instances of `mkdir` instead of `mkdir_vrfy` that is more commonly used.

ush/config_defaults.yaml
```diff
@@ -1997,6 +2007,10 @@ task_run_post:
   # used for post-processing. This is only used if CUSTOM_POST_CONFIG_FILE
   # is set to true.
   #
+  # TESTBED_FIELDS_FN
+  # The file which lists grib2 fields to be extracted to bgsfc for testbed
+  # Empty string means no need to generate bgsfc for testbed
```
christinaholtNOAA: Would you mind clarifying what bgsfc means here in the comment?

christinaholtNOAA (approved):
@BenjaminBlake-NOAA This looks really good. I left a couple of nit-picky comments/suggestions below, but overall, I think it looks great. Thanks for addressing my previous comments.

parm/FV3LAM_wflow.xml
BenjaminBlake-NOAA: Eventually we will likely want to run the prdgen job for subhourly output too, but initially with this PR I think it is fine to only include it for hourly output.

ush/config_defaults.yaml
BenjaminBlake-NOAA: bgsfc is an old naming convention that will no longer be used going forward.  I'll change this to something like: "The file which lists grib2 fields to be extracted for testbed files.  Empty string means no need to generate testbed files."

jobs/JREGIONAL_RUN_PRDGEN
BenjaminBlake-NOAA: Agreed, making this change now.

scripts/exregional_run_prdgen.sh
BenjaminBlake-NOAA: Yes that looks better, I'll modify both loops to use for domain in ${domains[@]}

scripts/exregional_run_prdgen.sh
BenjaminBlake-NOAA: I'll update those to mkdir_vrfy.

BenjaminBlake-NOAA:
I've addressed all of @christinaholtNOAA 's suggestions and added documentation for the run_prdgen variables to ConfigWorkflow.rst with @gspetro-NOAA 's help.  The Python unittests and functional tests have passed, so I think this PR is now ready to be merged.  Please review/approve when you have time - thanks!

christinaholtNOAA (approved):
LGTM! Thanks @BenjaminBlake-NOAA

danielabdi-noaa (dismissed):
I believe this should be close to the last PR to go in so that it can be tested it works well the `run_post` job and the `run_pygraf` job that comes after it. We can't be sure it works well with them unless it is tested that way. That is a major downside of the standalone approach. For example:
  - The changes I made to make "spinup" and "prod" cycles is removed from ex- and j-job scripts. That guarantees it will not work well with `run_post` and `run_pygraf`
  - The workflow added with this PR is rudimentary to say the least. It just gets in the way later IMO.
  - The "standalone" testing approach is not demonstrated with this PR. Where are the inputs, where are the standalone scripts to test it with?

docs/UsersGuide/source/ConfigWorkflow.rst
```diff
@@ -1222,6 +1225,48 @@ Set parameters for customizing the :term:`UPP`.
...
+   Intel Thread Affinity Interface for the ``run_prdgen`` task. See :ref:`this note <thread-affinity>` for more information on thread affinity.
+
+``OMP_NUM_THREADS_RUN_PRDGEN``: (Default: 1) 
+   The number of OpenMP threads to use for parallel regions.
+
+``OMP_STACKSIZE_RUN_PRDGEN``: (Default: "1024m")
+   Controls the size of the stack for threads created by the OpenMP implementation.
+
+``DO_PARALLEL_PRDGEN``: (Default: false)
+   Flag that determines whether to use CFP to run the product generation job in parallel.  This should be used with the RRFS_NA_3km grid and PPN_RUN_PRDGEN should be set to 22.
```
gspetro-NOAA: @BenjaminBlake-NOAA What is CFP? Better to define for newbies even if seasoned forecasters would know. 
Also, could you add backticks around RRFS_NA_3km grid and PPN_RUN_PRDGEN?
Thanks! :)

docs/UsersGuide/source/ConfigWorkflow.rst
```diff
@@ -1222,6 +1225,48 @@ Set parameters for customizing the :term:`UPP`.
...
+RUN_PRDGEN Configuration Parameters
+=====================================
+
+Non-default parameters for the ``run_prdgen`` task are set in the ``task_run_prdgen:`` section of the ``config.yaml`` file.
+
+Basic Task Parameters
+---------------------------------
+For each workflow task, certain parameter values must be passed to the job scheduler (e.g., Slurm), which submits a job for the task.
+
+``TN_RUN_PRDGEN``: (Default: "run_prdgen")   Set the name of this Rocoto workflow task. Users typically do not need to change this value.
```
gspetro-NOAA: @BenjaminBlake-NOAA I think you might need a new line before the definition on this one.

docs/UsersGuide/source/ConfigWorkflow.rst
```diff
@@ -1222,6 +1225,48 @@ Set parameters for customizing the :term:`UPP`.
...
+``OMP_STACKSIZE_RUN_PRDGEN``: (Default: "1024m")
+   Controls the size of the stack for threads created by the OpenMP implementation.
+
+``DO_PARALLEL_PRDGEN``: (Default: false)
+   Flag that determines whether to use CFP to run the product generation job in parallel.  This should be used with the RRFS_NA_3km grid and PPN_RUN_PRDGEN should be set to 22.
+
+``ADDNL_OUTPUT_GRIDS``: (Default: [])
+   Set additional output grids for wgrib2 remapping, if any.  Space-separated list of strings, e.g., ( "130" "242" "clue").  Default is no additional grids.
+
+``TESTBED_FIELDS_FN``: (Default: "")
```
gspetro-NOAA, line 1267: @BenjaminBlake-NOAA I think config_defaults.yaml had a TESTBED_FIELDS_FN2 field as well (but it looks like that might be gone now?), and both TESTBED_FIELDS_FN* were defined in the MAKE_GRID Config Params section. Is there a reason the second one has been omitted and the first moved to this section?

docs/UsersGuide/source/ConfigWorkflow.rst
BenjaminBlake-NOAA: Sure I can add backticks to those two variables.  CFP allows you to run multiple executions of code at the same time, but I can't seem to find what it actually stands for!  Will attempt to track that down...

docs/UsersGuide/source/ConfigWorkflow.rst
BenjaminBlake-NOAA: Thanks for catching that, I'll get it fixed.

docs/UsersGuide/source/ConfigWorkflow.rst
BenjaminBlake-NOAA: I think what happened was I realized that defining them in the MAKE_GRID Config Params section was a mistake, and since it is only needed for the RUN_PRDGEN task I moved it to there.  Originally there was also a TESTBED_FIELDS_FN2 variable for defining what parameters should go into the testbed files, but I think initially it makes sense to just add one TESTBED_FIELDS_FN variable at this time.

docs/UsersGuide/source/ConfigWorkflow.rst
BenjaminBlake-NOAA: I couldn't actually find what the CFP acronym stands for but I'll add a sentence describing what CFP is used for.

BenjaminBlake-NOAA:
@danielabdi-noaa  I am fine with waiting to merge this PR until later, but I was under the impression that we were interested in merging it soon.  I removed the changes related to the "spinup" and "prod" cycles because those tasks haven't been added to the SRW app yet, so it didn't make sense to include them in this PR.  @christinaholtNOAA do you think it makes sense to merge this work now or wait until other RRFS tasks have been added?

christinaholtNOAA:
@danielabdi-noaa this task can be added to the existing workflow cleanly, doesn't depend on any other additions, and has been demonstrated here with a WE2E test addition. It takes in the output of an existing task and creates products. It is a rudimentary task, so that it's implementation is simple makes sense.

When added to the existing workflow, a new task doesn't necessarily need a standalone method of testing. I've been suggesting that some tasks use standalone testing to avoid a huge expansion of logic associated with the XML generation. This one does not cause any of those issues.

The addition of `spinup` and `prod` cycles for RRFS will essentially be at the workflow level, and I don't think that having prod gen enabled should not change any of that. The dependencies and cycles may need to be changed, just like any other task that goes into the workflow when we are defining those cycles.

@BenjaminBlake-NOAA I think it's a good idea to go ahead with the merge now before your branch goes stale, causing you additional work.

MichaelLueken:
Since this is the next component that will be merged into the SRW App for the RRFS implementation, I would like to ask @gspetro-NOAA and @RatkoVasic-NOAA to review the changes one last time and, if @BenjaminBlake-NOAA has made the desired changes, to please go ahead and approve the PR.

@danielabdi-noaa If @christinaholtNOAA's above explanation is okay, please also provide an approval once you are okay with the changes.

Thanks!

MichaelLueken (approved):
@BenjaminBlake-NOAA These changes look good to me!  I do have one quick question - I see that you added a `run_prdgen.local.lua` file for WCOSS2.  Do the other machines require a run_prdgen task modulefile as well, or is this only an issue for WCOSS2?  Will go ahead and approve now.

danielabdi-noaa:
@christinaholtNOAA  I disagree. If I had not tested this within a workflow, it would be a stretch to say this will work within the RRFS dev workflow. 

- I do not think a PR that removes the "spinup" and "prod" cycles in the ex-scripts and j-jobs, which will definately not work later, is a good idea.

- I also do not think adding bits of the workflow that will definately rewritten for "spinup" and "prod" cycles later is a good idea either.

These two just add work for the next PRs to come. It would be best to leave out the changes to the xml file and add back the support for "spinup" and "prod" cycles in the scripts.

Also it is good to know that the strictly "standalone" approach is not a must have.

Having raised my concerns, I will remove my request for change though I will not approve it.

BenjaminBlake-NOAA:
@MichaelLueken  Thanks for reviewing the changes!  The other machines may also require a run_prdgen.local.lua file, unless we only plan on supporting the RRFS_NA_3km grid on WCOSS2.  That file is needed to load the CFP module when using DO_PARALLEL_PRDGEN = true with the RRFS_NA_3km grid.

tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta.yaml
```diff
@@ -37,3 +37,7 @@ task_run_fcst:
 task_run_post:
   NNODES_RUN_POST: 8
   PPN_RUN_POST: 12
+task_run_prdgen:
+  RUN_TASK_RUN_PRDGEN: true
```
danielabdi-noaa: This should be in `workflow_sections`.

scripts/exregional_run_prdgen.sh
```diff
@@ -0,0 +1,364 @@
...
+  gridname=""
+fi
+basetime=$( date +%y%j%H%M -d "${yyyymmdd} ${hh}" )
+
+# Create index (idx) files
+net4=$(echo ${NET:0:4} | tr '[:upper:]' '[:lower:]')
+for leveltype in prslev natlev ififip testbed
+do
+  if [ -f ${COMOUT}/${NET}.${cycle}${dot_ensmem}.${leveltype}.f${fhr}.${POST_OUTPUT_DOMAIN_NAME}.grib2 ]; then
+    ln_vrfy -sf --relative ${COMOUT}/${NET}.${cycle}${dot_ensmem}.${leveltype}.f${fhr}.${POST_OUTPUT_DOMAIN_NAME}.grib2 ${COMOUT}/${net4}.${cycle}.${leveltype}.f${fhr}.${gridname}grib2
```
danielabdi-noaa, line 170: There was a question whether to use `POST_OUTPUT_DOMAIN_NAME` or `TMMARK`. Last time I checked with Ming Hu, he mentioned we should use the latter even though I thought that `POST_OUTPUT_DOMAIN_NAME` fits with SRW's way of doing things better. May want to check with him again.

scripts/exregional_run_prdgen.sh
```diff
@@ -0,0 +1,364 @@
...
+elif  [ ${PREDEF_GRID_NAME} = "RRFS_NA_3km" ]; then
+  gridname=""
+fi
+basetime=$( date +%y%j%H%M -d "${yyyymmdd} ${hh}" )
+
+# Create index (idx) files
+net4=$(echo ${NET:0:4} | tr '[:upper:]' '[:lower:]')
+for leveltype in prslev natlev ififip testbed
+do
+  if [ -f ${COMOUT}/${NET}.${cycle}${dot_ensmem}.${leveltype}.f${fhr}.${POST_OUTPUT_DOMAIN_NAME}.grib2 ]; then
```
danielabdi-noaa, line 169: I have not tested the naming of files with "ensemble" members properly -- because RRFS CONUS 3km did not have it IIRC. So the `${dot_ensmem}` additions are not tested although I have tested the change to `f${fhr}` from `${fhr}` which was represented as `{fhr}` works

scripts/exregional_run_prdgen.sh
```diff
@@ -0,0 +1,364 @@
...
+# extract the output fields for the testbed files
+touch ${testbed}
+if [[ ! -z ${TESTBED_FIELDS_FN} ]]; then
+  if [[ -f "${PARMdir}/upp/${TESTBED_FIELDS_FN}" ]]; then
+    wgrib2 ${prslev} | grep -F -f "${PARMdir}/upp/${TESTBED_FIELDS_FN}" | wgrib2 -i -grib ${testbed} ${prslev}
+  else
+    echo "${PARMdir}/upp/${TESTBED_FIELDS_FN} not found"
+  fi
+fi
+if [[ ! -z ${TESTBED_FIELDS_FN2} ]]; then
```
danielabdi-noaa: This is a bug if `TESTBED_FIELDS_FN2` is not defined? I think @gspetro-NOAA also poointed this out.

ush/config_defaults.yaml
```diff
@@ -2015,6 +2029,50 @@ task_run_post:
...
   CUSTOM_POST_CONFIG_FP: ""
   POST_OUTPUT_DOMAIN_NAME: '{{ workflow.PREDEF_GRID_NAME }}'
+  TESTBED_FIELDS_FN: ""
+
+#----------------------------
+# RUN PRDGEN config parameters
+#-----------------------------
+task_run_prdgen:
+  TN_RUN_PRDGEN: "run_prdgen"
+  NNODES_RUN_PRDGEN: 1
```
danielabdi-noaa, line 2039: I believe the default should be 22 since that was what was hardcoded in the command
```
mpiexec -np 22 --cpu-bind core cfp $CMDFILE  #Daniel: Move this somewhere else
```

tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta.yaml
```diff
@@ -37,3 +37,7 @@ task_run_fcst:
...
+  DO_PARALLEL_PRDGEN: true
+  PPN_RUN_PRDGEN: 22
```
danielabdi-noaa, line 44: I believe 22 should be assigned to `NNODES_RUN_PRDGEN` instead of `PPN_RUN_PRDGEN`

tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta.yaml
christinaholtNOAA: I believe that is `workflow_switches`

BenjaminBlake-NOAA: Yes it looks like it should fall under workflow_switches.  I'll fix that along with your other suggested changes below and push the fixes to GitHub soon - thanks for reviewing!

scripts/exregional_run_prdgen.sh
BenjaminBlake-NOAA: Good catch, I missed that somehow.

ush/config_defaults.yaml
BenjaminBlake-NOAA: Yeah setting the default as 22 sounds good to me.

scripts/exregional_run_prdgen.sh
danielabdi-noaa: Also both of these files are not available in current SRW app. So we would need to have those two fix files incorporated first. In my test, I pointed to the fix files from `RRFS_dev1` clone temporarily until a way is settled on how to incorporate the fix files.
```
  CUSTOM_POST_CONFIG_FP: '{{ [user.PARMdir, "upp/postxconfig-NT-fv3lam_rrfs.txt"]|path_join }}'
  CUSTOM_POST_PARAMS_FP: '{{ [user.PARMdir, "upp/params_grib2_tbl_new"]|path_join }}'
  TESTBED_FIELDS_FN: "testbed_fields_bgdawp.txt"
  TESTBED_FIELDS_FN2: "testbed_fields_bgrd3d.txt"
```
Moreover, these variables are first used by the `run_post` job so they should be defined there.
Another reason why order of tasks is important.

tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta.yaml
BenjaminBlake-NOAA: Actually I intended for PPN_RUN_PRDGEN to be 22, so the task would run on 1 node but using 22 tasks per node, instead of 22 nodes with 1 task on each node.

scripts/exregional_run_prdgen.sh
BenjaminBlake-NOAA: @hu5970 Are you ok with us using POST_OUTPUT_DOMAIN_NAME or would TMMARK be preferred?  I think I agree with @danielabdi-noaa and would rather use POST_OUTPUT_DOMAIN_NAME.

scripts/exregional_run_prdgen.sh
BenjaminBlake-NOAA: I was thinking about the best way to handle this too.  I tested the prdgen task on WCOSS2 without these files and it completed successfully for me.  If the files aren't defined in the config.yaml file, then the testbed file generation part of the job should be skipped.  But once the run_post job is updated and the necessary files are added, this logic will already be in place for the testbed file generation and it can be turned on then.

tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta.yaml
danielabdi-noaa: I am not sure about the parallel prdgen code. The script  generated on the fly that `mpiexec` calls launches 22 background jobs and waits, i.e. it is already parallel. However, we call the script with `mpiexec -np 22` or RUN_CMD_PRDGEN. Wouldn't that be  effectively calling 22x22 jobs?

parm/FV3LAM_wflow.xml
```diff
@@ -1188,6 +1189,46 @@ the <task> tag to be identical to the ones above for other output times.
...
+<!--
+************************************************************************
+************************************************************************
+-->
+<metatask name="&TN_RUN_PRDGEN;{{ uscore_ensmem_name }}">
+
+    <var name="fhr">{% for h in range(0, fcst_len_hrs+1) %}{{ " %03d" % h }}{% endfor %}
+</var>
+    <task name="&TN_RUN_PRDGEN;{{ uscore_ensmem_name }}_f#fhr#" cycledefs="forecast" max
+tries="{{ maxtries_run_prdgen }}">
```
danielabdi-noaa: There is a mistake here.

BenjaminBlake-NOAA: Oh yeah the formatting got messed up somehow here.  I'll fix that

tests/WE2E/test_configs/grids_extrn_mdls_suites_community/config.grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta.yaml
BenjaminBlake-NOAA: So the CFP module is actually what spreads out the tasks.  We define RUN_CMD_PRDGEN as "mpiexec -n $nprocs --cpu-bind core cfp".  In the ex-script, it gets called with eval $RUN_CMD_PRDGEN $CMDFILE, and cfp will execute each line of the script using 1 node.  For those 22 executions we just need 22 tasks on 1 node.

danielabdi-noaa: Well this doesn't work on Hera since RUN_CMD_PRDGEN's defintion does not have "cfp". The prdgen task failed with out of memory error
```
slurmstepd: error: *** JOB 42907778 ON h33m43 CANCELLED AT 2023-03-15T10:16:19 DUE TO TIME LIMIT ***
slurmstepd: error: Detected 1786 oom-kill event(s) in StepId=42907778.batch. Some of your processes may have been killed by the cgroup out-of-memory handler.

```
So I think the RUN_CMD_PRDGEN defintion for all systems other than wcoss2 should be executed serially (mpirun -np 1) since there is no cfp there.

BenjaminBlake-NOAA: Yes I agree.  The job will run more slowly as a serial job without CFP, but it should still work - the wall clock will need to be longer though.  Since the primary location this job will run is on WCOSS2 I think that is acceptable though.  If we will only use CFP on WCOSS2, then run_prdgen.local.lua files are not needed for the other supported platforms.

gspetro-NOAA:
@MichaelLueken @BenjaminBlake-NOAA I'm all set on the doc side! Didn't do an actual approval review because I can't speak to the code side. ;) But presuming that's all set, I agree this PR is ready to go!

MichaelLueken:
@gspetro-NOAA Thanks for approving the changes to the doc side!

christinaholtNOAA:
@MichaelLueken Will this one be ready for merge this morning, as well? It could be helpful to get it merged into my PR for the next round of testing. Thanks!

MichaelLueken:
@christinaholtNOAA This PR likely won't be merged this morning:

 1. I still need to submit the Jenkins tests (I can go ahead and do so now),
 2. following our discussion from two weeks ago, we still need an approval from an EMC code manager, and
 3. there appears to be an open discussion still related to the `POST_OUTPUT_DOMAIN_NAME` vs `TMMARK` naming convention.

As noted in 1, I will go ahead and submit the Jenkins tests now.  Is this update fine with not including an approval from an EMC code manager and should we apply some additional pressure to @hu5970 for an answer to the still open discussion on `scripts/exregional_run_prdgen.sh`?

BenjaminBlake-NOAA:
@MichaelLueken  It would be good if an EMC code manager could also submit an approval.  @RatkoVasic-NOAA or @chan-hoo do either of you have time to review/approve this PR?  As for the POST_OUTPUT_DOMAIN_NAME vs TMMARK issue, I am not sure what would be best.

RatkoVasic-NOAA:
@BenjaminBlake-NOAA did you get opinion from Ming Hu for POST_OUTPUT_DOMAIN_NAME vs TMMARK?

BenjaminBlake-NOAA:
@RatkoVasic-NOAA  Not yet.  I thought we wanted to move away from using $TMMARK, but I'd like for Ming or others who are more knowledgeable to comment on that.

hu5970:
$TMMARK is for NAM type spin-up cycle. The current RRFS is using RAP/HRRR type spin-up cycle. I did not explain those two type of sin-up cycle here so you will feel confused. To be simple, in current RRFS cycles, the $TMMARK is always "tm00", it is means the current cycle minus 0. So for a file "RRFS_CONUS.t10z.bgdawpf003.tm00.grib2", I will see this is 3-h forecast from 10Z cycle. So "$TMMARK" should be removed or replaced with the "POST_OUTPUT_DOMAIN_NAME".

MichaelLueken:
@hu5970 Thank you very much for confirming that replacing `$TMMARK` with `$POST_OUTPUT_DOMAIN_NAME` is the correct move to make!  This has been done in this PR.

@RatkoVasic-NOAA With confirmation from Ming, I think we can continue moving forward with this update.  Would either you or @chan-hoo be able to review and approve this PR, so that the second issue is addressed?

@BenjaminBlake-NOAA Once the Jenkins tests are complete, I think this work will be ready to be merged.

RatkoVasic-NOAA (approved):
Looks good to me.

MichaelLueken:
The Jenkins tests have passed.  With @RatkoVasic-NOAA's approval, I will now merge this work to develop.

PR #459, opened 2022-11-09
Minor bug fixes

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 

Minor bug fixes related to
- building (without installing) the ufs weather model
- deletion of forecast directory
- running `setup_WE2e_tests.sh` with conda activated, see issue #458 

### Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## TESTS CONDUCTED: 

- [ ] hera.intel
- [ ] orion.intel
- [ ] cheyenne.intel
- [ ] cheyenne.gnu
- [ ] gaea.intel
- [ ] jet.intel
- [ ] wcoss2.intel
- [ ] NOAA Cloud (indicate which platform)
- [ ] Jenkins
- [ ] fundamental test suite
- [ ] comprehensive tests (specify *which* if a subset was used)

## DEPENDENCIES:

None

## ISSUE: 

- #458 

## CHECKLIST

- [X] My code follows the style guidelines in the Contributor's Guide
- [ ] I have performed a self-review of my own code using the Code Reviewer's Guide
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes need updates to the documentation. I have made corresponding changes to the documentation
- [ ] My changes do not require updates to the documentation (explain).
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] Any dependent changes have been merged and published

## LABELS (optional): 

A Code Manager needs to add the following labels to this PR: 
- [ ] Work In Progress
- [ ] bug
- [ ] enhancement
- [ ] documentation
- [ ] release
- [ ] high priority
- [ ] run_ci
- [ ] run_we2e_fundamental_tests
- [ ] run_we2e_comprehensive_tests
- [ ] Needs Cheyenne test 
- [ ] Needs Jet test 
- [ ] Needs Hera test 
- [ ] Needs Orion test 
- [ ] help wanted

Issue #635, opened 2023-03-03
Failure to clone UFS_UTILS/ccpp-physics/physics/rte-rrtmgp on Orion


MichaelLueken:
## Expected behavior
On Orion, using `./manage_externals/checkout_externals` should clone all of the externals required to build the SRW App.

## Current behavior
On Orion, using `./manage_externals/checkout_externals` fails while cloning `UFS_UTILS` with the following message:

```
Orion-login-1[57] mlueken$ ./manage_externals/checkout_externals 
Processing externals description file : Externals.cfg
Checking status of externals: ufs-weather-model, aqm-utils, rrfs_utl, arl_nexus, ufs_utils, gsi, upp, 
Checking out externals: aqm-utils, ufs_utils, ERROR:root:Command '[u'git', u'submodule', u'update', u'--init', u'--recursive']' returned non-zero exit status 1
ERROR:root:Failed with output:
    Submodule 'ccpp-physics' (https://github.com/NCAR/ccpp-physics.git) registered for path 'ccpp-physics'
    Cloning into 'ccpp-physics'...
    Submodule path 'ccpp-physics': checked out '12c115e992d3a265eaaa67d72fcbdb3a6f21195f'
    Submodule 'physics/rte-rrtmgp' (https://github.com/earth-system-radiation/rte-rrtmgp) registered for path 'physics/rte-rrtmgp'
    Cloning into 'physics/rte-rrtmgp'...
    fatal: reference is not a tree: 7f01618c92409658bddd3afa9acb004c608f6a0d
    Unable to checkout '7f01618c92409658bddd3afa9acb004c608f6a0d' in submodule path 'ccpp-physics/physics/rte-rrtmgp'
    Failed to recurse into submodule path 'ccpp-physics'

ERROR: In directory
    /work/noaa/epic-ps/mlueken/ufs-srweather-app/sorc/UFS_UTILS
Process did not run successfully; returned status 1:
    git submodule update --init --recursive
See above for output from failed command.

ERROR:root:Failed with output:
    Submodule 'ccpp-physics' (https://github.com/NCAR/ccpp-physics.git) registered for path 'ccpp-physics'
    Cloning into 'ccpp-physics'...
    Submodule path 'ccpp-physics': checked out '12c115e992d3a265eaaa67d72fcbdb3a6f21195f'
[...]

PR #637, opened 2023-03-06
[develop] Replace shell-based WE2E scripts with python versions

### LINKED ISSUE #462: Running comprehensive suite of WE2E tests overwhelms some platforms

### LINKED ISSUE #568: Script `run_WE2E_tests.sh` silently discards invalid test options
## Expected behavior
If an invalid option is provided in one of the config.yaml files for a WE2E test, we definitely want to know about it! Either the script should fail and call attention to the offending variable if an invalid option is provided, or that option should be fed directly to the generate_FV3LAM_workflow.py script directly where it can fail as the intended behavior for an invalid option.

## Current behavior
Some WE2E tests require manipulation before they can be submitted on a specific platform; for example, to test the ability to read user-staged files, the run_WE2E_tests script needs to manually substitute a "dummy" value for the location of staged test input data for a given machine. Because of the way that run_WE2E_tests.sh constructs a `config.yaml` file after these manipulations, any options that are not valid are silently discarded rather than being carried through.

## Machines affected
All

## Steps To Reproduce
Just a simple example, though there are countless ways this could be reproduced:
1. Append file `ufs-srweather-app/tests/WE2E/test_configs/release_SRW_v1/config.GST_release_public_v1.yaml` to include the line:
```
  this_option: doesn't exist
```
2. Run the `run_WE2E_tests.sh` script for the `GST_release_public_v1` test
```
cd ufs-srweather-app/tests/WE2E
echo "GST_release_public_v1" > test
./run_WE2E_tests.sh tests_file=test machine=hera account=fv3lam
```
3. Observe that the resulting config.yaml file (which should be found at `../../../expt_dirs/GST_release_public_v1/config.yaml` in this example) does not include the new fancy option you wanted.

## Possible Implementation
The newly provided pythonized version of the WE2E script (introduced in #558) takes a different approach to constructing the config.yaml for a given test, ensuring that all provided values are carried forward in the resulting config.yaml. When this script replaces the older bash version, any invalid options will be caught by the `generate_FV3LAM_workflow` function, as expected.

### LINKED ISSUE #571: `run_WE2E_tests.sh` does not run the intended tests on Hera if COMPILER is not explicitly set
## Expected behavior
The `run_WE2E_tests.sh` script should check to see if the following pre-defined file lists exist in the directory `ufs-srweather-app/tests/WE2E/machine_suites/`, in order of priority:
1. `${test_type}.${machine}.${compiler}.nco`
2. `${test_type}.${machine}.${compiler}.com`
3. `${test_type}.${machine}.${compiler}`
4. `${test_type}.${machine}`
5. `${test_type}`

So on Hera, running the fundamental tests with no other flags should run the tests found in the file `fundamental.hera.intel.nco`, since that file exists in `machine_suites/` and it is first in the priority list. If the user specifies "--compiler=gnu", then the tests in `fundamental.hera.gnu.com` should be run.

## Current behavior
With commit d03e64619 (PR #477), the logic for checking the set of "fundamental" or "comprehensive" tests to run has been broken. The current logic assumes that "compiler" has been set, but if the user has not provided the argument then the default value will not have been filled in yet. So running the fundamental tests on, for example, Hera without specifying the compiler will default to the set of tests listed in `machine_suites/fundamental`, rather than the intended `machine_suites/fundamental.hera.intel.nco`, because ${compiler} will resolve as an empty string, meaning the logic looks for `fundamental.hera..nco`, which does not exist. Likewise, `fundamental.hera..nco`, `fundamental.hera.`, and `fundamental.hera` do not exist, so finally the logic settles on the tests found in `machine_suites/fundamental`.

If `compiler=intel` or `compiler=gnu` is explicitly set, the correct set of tests is used.

## Machines affected
This logic will be broken if compiler is not explicitly set for any platform with a compiler-specific test suite for the default compiler; currently this is only Hera.

## Steps To Reproduce
1. On Hera, run the WE2E tests without using the `compiler=` flag:

2. Observe that the set of tests run is incorrect:

```
After processing the user-specified list of WE2E tests to run, the number 
of tests to run (num_tests_to_run) is
  num_tests_to_run = 9
and the list of WE2E tests to run (one test per line) is
  'grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2'
  'grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16'
  'grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_RAP_suite_HRRR'
  'grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v15p2'
  'grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_HRRR'
  'grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_RRFS_v1beta'
  'grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR'
  'grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta'
  'nco_grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR'
```

This is the list of tests specified in `machine_suites/fundamental`, rather than the intended tests in `machine_suites/fundamental.hera.intel.nco`.

## Detailed Description of Fix (optional)
The logic filling in default values for input arguments should be moved earlier to before this logic.

The new python-based script (#558) does not suffer from this error, so when that is adopted this problem will go away.

## Additional Information
Note: As I have previously mentioned, Il do not think that this testing strategy is a good idea. The "fundamental" set of tests should be the same on each platform for consistency, and it should be a small subset of tests checking the "most important" functionality of the workflow to ensure that nothing major has been broken, regardless of which platform is used for development/testing. But I felt it was important to point out that the intended functionality is currently broken.

### LINKED ISSUE #586: Replace shell-based WE2E test scripts and `run_srw_tests.py` script with `run_WE2E_tests.py` and `monitor_jobs.py`
## Description
#558 introduced the first implementation of `run_WE2E_tests.py` and `monitor_jobs.py`. The current version accomplishes most of the existing functionality of `run_WE2E_tests.sh`, but stopped short of a full replacement to reduce the scope of the effort. Now that the first PR has been merged, we should identify the work needed to completely deprecate `run_WE2E_tests.sh` and shift all testing efforts to the python script.

The scripts that will be deprecated are:
 - `run_WE2E_tests.sh`: This is the original, bash-based script for generating and submitting the workflow end-to-end tests
 - `run_srw_tests.py`: Introduced in #466, this was a temporary solution to platforms that had limitations running full experiment suites in crontabs. `monitor_jobs.py` offers a more robust solution.
 - `get_WE2Etest_names_subdirs_descs.sh`: This is a script that creates a pipe-delimited .csv file describing the various WE2E tests and settings within. There is a lot of other functionality used by the `run_WE2E_tests.sh` script, but this is unnecessary in the python version.
 - `get_expts_status.sh` This script 
 - `create_WE2E_resource_summary.py` This script prints out a summary of the number of core hours used by a given experiment or set of experiments based on information in the rocoto database files for each experiment.

## Solution
This issue will track the steps needed to achieve this replacement without disruption to the normal development process.

## Requirements
 - [ ] `run_WE2E_tests.py` and its related functions should include all functionality present in scripts to be replaced
   - [ ] `run_WE2E_tests.sh`
   - [ ] `run_srw_tests.py`
   - [x] `get_WE2Etest_names_subdirs_descs.sh`
     - [x] The test details file functionality has been added to run_WE2E_tests.py with new `--print_test_details` flag
     - [x] Most other functionality is unnecessary with new scripts
   - [x] `get_expts_status.sh`
     - [x]  This functionality will now be available from both monitor_jobs.py and WE2E_summary.py
   - [x] `create_WE2E_resource_summary.py` 
     - [x] This information will now be available from `WE2E_summary.py`, with several improvements:
       - Core hours are calculated more correctly (by charging full use of node)
       - Can pass either an experiment directory or a yaml experiment file
 - [ ] OR, give proper justification for why a feature will be removed/deprecated, with approval from the code management team
 - [ ] Jenkins testing must be updated to use the new script
 - [ ] Automated label-based WE2E tests must be updated to use the new script
 - [ ] Any additional use cases not mentioned above should be identified and accounted for

## Acceptance Criteria
 - [ ] Requirements must be met.

## Dependencies 
Will track relevant PRs here.

mkavulich:
## DESCRIPTION OF CHANGES: 
This PR improves on the new `./run_WE2E_tests.py` script (introduced in #558), implementing all the features present in the previous shell-based workflow. Some new files are also introduced for better organization and additional functionality:

 - `tests/WE2E/utils.py` This is a collection of functions used by other scripts, contained here to avoid circular dependencies.
 - `tests/WE2E/WE2E_summary.py` Given an experiment directory or .yaml file, outputs a summary to screen of each experiment, its status, and the number of core hours used. It also prints a summary file with detailed information about each task for each experiment.
 - `tests/WE2E/print_test_info.py` Will print a file `WE2E_test_info.txt`, very similar to the legacy `WE2E_test_info.csv` with just a few minor format differences.

Any scripts can be run with the `-h` argument to print information about all available options (not including `utils.py`, which is not designed to be run stand-alone).

**With this PR, the old superseded shell-based tools are removed.** This will require action by anyone whose workflow relies on those tools. I will reach out to those that I know have automated testing based on these tools, but please let me know if any other capabilities or modifications are needed to keep existing workflows working.

### New options
 - `--procs=##` For monitoring and submitting jobs, this will call `rocotorun` and read rocoto database files in parallel for the specified number of parallel tasks. This can greatly speed up the submission of large test suites, such as the comprehensive tests. **Only use this option if you are sure you have access to the specified number of parallel cores**; python will behave badly if you try to over-subscribe your available tasks.
 - `--opsroot=` If test is for NCO mode, sets OPSROOT
 - `--print_test_info` Create a `WE2E_test_info.txt` file summarizing each test prior to starting experiment (False by default)

### Other updates
 - The `-d` flag (debug mode) now captures output from `rocotorun` in log files, such as job cards and job submission messages. This is not enabled by default because it is very slow.
 - Job monitoring can now be paused with `ctrl-c`, and a message on how to resume testing is printed to screen
 - All auto-generated experiment monitoring files now use the naming convention `WE2E_tests_DATETIME.yaml`
 - functions `check_task_get_extrn_ics()` and `check_task_get_extrn_lbcs()` are consolidated into a single `check_task_get_extrn_bcs()` function
 - If `TEST_VX_FCST_INPUT_BASEDIR` is not set for this platform, do not exit with an error, instead set to empty string
 - Added more detailed checks to ensure that an experiment is actually finished before script exits
 - Fixed broken capabilities (calculating "relative cost" and number of forecasts) in `WE2E_test_info.txt` file

### Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## TESTS CONDUCTED: 
I have tested on all platforms I have access to; I would appreciate help testing on other platforms.

- [x] hera.intel
  - Ran three identical suites (from `tests/WE2E/machine_suites/fundamental`) with the old `run_WE2E_tests.sh`, and with `run_WE2E_tests.py` in the default configuration, as well as using the ``--use_cron_to_relaunch`` option. All tests produced identical output.
  - Ran all test suite options (`fundamental`,`comprehensive`, and `all`), as well as individual tests and input files with test names to ensure expected behavior for all `-t` options
  - Quit script using `ctrl-c` at various points during the execution to ensure expected behavior
  - Ran `WE2E_summary.py` on several previous experiment directories to ensure expected output
- [x] orion.intel
  - Ran fundamental tests
- [x] cheyenne.intel
  - Ran fundamental tests
- [x] cheyenne.gnu
  - Ran fundamental tests
- [ ] gaea.intel
- [x] jet.intel
  - Ran all tests
- [ ] wcoss2.intel
- [ ] Jenkins

## DOCUMENTATION:
Documentation is updated for relevant WE2E capabilities that have been changed/added. A lot of documentation related to individual tests and suites is still outdated, but this will be updated at a future date with changes related to #587. 

This version of the docs can be viewed here; changes were made to chapters 10 and 12: https://ufs-srweather-app-mkavulich.readthedocs.io/en/latest/index.html

## ISSUE: 
 - Resolves #586 
 - Resolves #462
   - Users on Cheyenne can submit `./run_WE2E_tests.py` to avoid use of crontab; to run faster it can be run on a compute node with the `-p` option.
   - Users on Hera, where occasional failures occurred due to reaching a job submission limit, will no longer see this error (unless legacy crontab option is used)
 - Fixes #571 
 - Fixes #568 

## CHECKLIST

- [x] My changes need updates to the documentation. I have made corresponding changes to the documentation

## CONTRIBUTORS: 
@christinaholtNOAA Contributed a better way of setting relative dates (as used in test `get_from_NOMADS_ics_FV3GFS_lbcs_FV3GFS`)


tests/WE2E/utils.py
```diff
@@ -0,0 +1,597 @@
...
+                nnodes = vdf[nnodes_var]
+                # Users are charged for full use of nodes, so core hours are CPN * nodes * time in hrs
+                core_hours = cores_per_node * nnodes * expt_dict[expt][task]['walltime'] / 3600
+                expt_dict[expt][task]['exact_count'] = True
+            else:
+                # If we can't find the number of nodes, assume full usage (may undercount)
+                core_hours = expt_dict[expt][task]['cores'] * expt_dict[expt][task]['walltime'] / 3600
+                expt_dict[expt][task]['exact_count'] = False
+            expt_dict[expt][task]['core_hours'] = round(core_hours,2)
+    return expt_dict
```
christinaholtNOAA: No need to return expt_dict here. It's updated in-place.

tests/WE2E/utils.py
```diff
@@ -0,0 +1,597 @@
...
+            expt_dict[item] = dict()
+            expt_dict[item].update({"expt_dir": os.path.join(expt_dir,item)})
+            expt_dict[item].update({"status": "CREATED"})
+        else:
+            logging.debug(f'Skipping directory {item}, experiment XML file not found')
+            continue
+        #Update the experiment dictionary
+        logging.debug(f"Reading status of experiment {item}")
+        update_expt_status(expt_dict[item],item,True,False)
+    summary_file = f'WE2E_tests_{datetime.now().strftime("%Y%m%d%H%M%S")}.yaml'
```
christinaholtNOAA, line 135: This is bad indentation, I think.

tests/WE2E/WE2E_summary.py
```diff
@@ -0,0 +1,73 @@
...
+    # Set up dictionary of experiments
+    if args.expt_dir:
+        yaml_file, expt_dict = create_expt_dict(args.expt_dir)
+    elif args.yaml_file:
+        expt_dict = load_config_file(args.yaml_file)
+    else:
+        raise ValueError(f'Bad arguments; run {__file__} -h for more information')
+
+    # Calculate core hours and update yaml
+    expt_dict = calculate_core_hours(expt_dict)
```
christinaholtNOAA: This should probably just update the dict in-place, no?

tests/WE2E/run_WE2E_tests.py
```diff
@@ -283,121 +303,81 @@ def check_test(test: str) -> str:
...
-    Function for checking and updating various settings in task_get_extrn_lbcs section of test config yaml
-
-    Args:
-        cfg  : Dictionary loaded from test config file
-        mach : Dictionary loaded from machine settings file
-        dflt : Dictionary loaded from default config file
-    Returns:
-        cfg_lbcs : Updated dictionary for task_get_extrn_lbcs section of test config
-    """
+    I_OR_L = ics_or_lbcs.upper()
```
christinaholtNOAA, line 324: `I_OR_L` is not a great variable name.

mkavulich:
## Notes on platform-specific usage
I wasn't sure the best place to put these notes so I'll paste them here for future reference.

While it is still possible to run WE2E tests using the crontab option, in general the new script is faster. To speed up the test submission even further, you can use the `-p=PROCS` option to run PROCS parallel processes.

### Cheyenne

Usage on Cheyenne was a big part of why this script was introduced (#462). Use of crontabs is discouraged by sysadmins on that platform, and there are fairly strict time and memory limits on the use of login nodes. While smaller test suites may run successfully using the login node, it is recommended to run tests on a login node.

```
cd ufs-srweather-app/tests/WE2E
qinteractive -l walltime=7200 -A account_name
module use ../../modulefiles/
module load wflow_cheyenne
conda activate regional_workflow
./run_WE2E_tests.py -m=cheyenne -a=account_name --tests=fundamental --expt_basedir=fundamental_parallel -p=32
```

The `qinteractive` command puts the user on a login node for an hour by default. The `-l walltime=7200` option extends this to 2 hours (7200 seconds). Then we can run `./run_WE2E_tests.py` with the `-p=32` options to make sure we utilize all 32 cores for job monitoring/submission.

### Hera
The rocotorun tasks run quite fast on Hera, so for most testing even non-parallel runs will submit jobs with little delay. For very large test suites, such as the `comprehensive` suite, running in parallel can give a significant speed-up, but due to limitations of the Hera system, it is not recommended to use more than 4 processes. Using more may result in rocotorun tasks being unable to initialize, which can result in errors in job submission.

.cicd/scripts/srw_test.sh
```diff
@@ -38,58 +38,18 @@ else
...
-# Run the new run_srw_tests script if the machine is Cheyenne.
-if [[ "${platform}" = "cheyenne" ]]; then
-    cd ${workspace}/ush
-    ./run_srw_tests.py -e=${we2e_experiment_base_dir}
-    cd ${we2e_test_dir}
-fi
+    opsroot=${nco_dir} | tee ${progress_file}
 
 # Progress file
 progress_file="${workspace}/we2e_test_results-${platform}-${SRW_COMPILER}.txt"
```
MichaelLueken: These lines should be removed, since they have been moved to lines 41 and 42.

.cicd/scripts/srw_test.sh
```diff
@@ -38,58 +38,18 @@ else
 fi
 
 cd ${we2e_test_dir}
+# Progress file
+progress_file="${workspace}/we2e_test_results-${platform}-${SRW_COMPILER}.txt"
 ./setup_WE2E_tests.sh ${platform} ${SRW_PROJECT} ${SRW_COMPILER} ${test_type} \
     expt_basedir=${we2e_experiment_base_dir} \
```
MichaelLueken: Both `expt_basedir` and `opsroot` should have `--` added to them:

--expt_basedir=${we2e_experiment_base_dir}
--opsroot=${nco_dir}

tests/WE2E/setup_WE2E_tests.sh
```diff
@@ -54,16 +54,16 @@ test_type=${4:-fundamental}
 #----------------------------------------------------------------------
 opts=
 if [[ "$*" != *"debug"* ]]; then
-   opts="${opts} debug=TRUE"
+   opts="${opts} --debug=TRUE"
```
MichaelLueken: Since there is no boolean associated with `--debug` in `run_WE2E_tests.py`, `=TRUE` should be removed:

`opts="${opts} --debug`

tests/WE2E/setup_WE2E_tests.sh
```diff
@@ -54,16 +54,16 @@ test_type=${4:-fundamental}
...
 fi
 if [[ "$*" != *"verbose"* ]]; then
-   opts="${opts} verbose=TRUE"
+   opts="${opts} --verbose=TRUE"
```
MichaelLueken: Since there is no boolean associated with `--verbose` in `run_WE2E_tests.py`, `=TRUE` should be removed:

`opts="${opts} --verbose`

MichaelLueken (commented):
@mkavulich I was able to clone your branch and run the fundamental tests using `tests/WE2E/run_WE2E_tests.py` without issue.

I encountered issues while attempting to manually run the automated Jenkins tests (`.cicd/scripts/srw_test.sh`).  I have noted in the review the changes that I had to make in order to start the tests.  Following the required changes, the tests fail with the following:

```
calling function that monitors jobs, prints summary
Writing information for all experiments to WE2E_tests_20230307120101.yaml
Checking tests available for monitoring...
Starting experiment grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta running
Updating database for experiment grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta
03/07/23 12:01:02 CST :: FV3LAM_wflow.xml :: Workflow Manager Initialization failed.
03/07/23 12:01:02 CST :: FV3LAM_wflow.xml :: Permission denied - /.rocoto
/apps/contrib/rocoto/1.3.3/lib/workflowmgr/utilities.rb:187:in `initialize': No such file or directory - /.rocoto/1.3.3/FV3LAM_wflow/log.lock (Errno::ENOENT)
	from /apps/contrib/rocoto/1.3.3/lib/workflowmgr/utilities.rb:187:in `open'
	from /apps/contrib/rocoto/1.3.3/lib/workflowmgr/utilities.rb:187:in `log'
	from /apps/contrib/rocoto/1.3.3/lib/workflowmgr/workflowengine.rb:78:in `rescue in initialize'
	from /apps/contrib/rocoto/1.3.3/lib/workflowmgr/workflowengine.rb:37:in `initialize'
	from /apps/contrib/rocoto/1.3.3/bin/../sbin/rocotorun.rb:34:in `new'
	from /apps/contrib/rocoto/1.3.3/bin/../sbin/rocotorun.rb:34:in `<main>'

03/07/23 12:01:02 CST :: FV3LAM_wflow.xml :: Workflow Manager Initialization failed.
03/07/23 12:01:02 CST :: FV3LAM_wflow.xml :: Permission denied - /.rocoto
/apps/contrib/rocoto/1.3.3/lib/workflowmgr/utilities.rb:187:in `initialize': No such file or directory - /.rocoto/1.3.3/FV3LAM_wflow/log.lock (Errno::ENOENT)
	from /apps/contrib/rocoto/1.3.3/lib/workflowmgr/utilities.rb:187:in `open'
	from /apps/contrib/rocoto/1.3.3/lib/workflowmgr/utilities.rb:187:in `log'
	from /apps/contrib/rocoto/1.3.3/lib/workflowmgr/workflowengine.rb:78:in `rescue in initialize'
	from /apps/contrib/rocoto/1.3.3/lib/workflowmgr/workflowengine.rb:37:in `initialize'
	from /apps/contrib/rocoto/1.3.3/bin/../sbin/rocotorun.rb:34:in `new'
	from /apps/contrib/rocoto/1.3.3/bin/../sbin/rocotorun.rb:34:in `<main>'

Reading database for experiment grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta, updating experiment dictionary

*********************************************************************
FATAL ERROR:
Experiment generation failed. See the error message(s) printed below.
For more detailed information, check the log file from the workflow
generation script: log.run_WE2E_tests
*********************************************************************

  File "/work/noaa/epic-ps/mlueken/ufs-srweather-app/tests/WE2E/./run_WE2E_tests.py", line 534, in <module>
    run_we2e_tests(homedir,args)
  File "/work/noaa/epic-ps/mlueken/ufs-srweather-app/tests/WE2E/./run_WE2E_tests.py", line 222, in run_we2e_tests
    monitor_file = monitor_jobs(monitor_yaml, monitor_file=monitor_file, procs=args.procs,
  File "/work/noaa/epic-ps/mlueken/ufs-srweather-app/tests/WE2E/monitor_jobs.py", line 50, in monitor_jobs
    expts_dict[expt] = update_expt_status(expts_dict[expt], expt, True, debug)
  File "/work/noaa/epic-ps/mlueken/ufs-srweather-app/tests/WE2E/utils.py", line 291, in update_expt_status
    for task in db:
UnboundLocalError: local variable 'db' referenced before assignment
```
Looking in `WE2E_tests_20230307120101.yaml`, I see:

```
### WARNING ###
### THIS FILE IS AUTO_GENERATED AND REGULARLY OVER-WRITTEN BY WORKFLOW SCRIPTS
### EDITS MAY RESULT IN MISBEHAVIOR OF EXPERIMENTS RUNNING
grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta:
  expt_dir: /work/noaa/epic-ps/mlueken/ufs-srweather-app/expt_dirs/grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta
  status: CREATED
grid_SUBCONUS_Ind_3km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta:
  expt_dir: /work/noaa/epic-ps/mlueken/ufs-srweather-app/expt_dirs/grid_SUBCONUS_Ind_3km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
  status: CREATED
community_ensemble_008mems:
  expt_dir: /work/noaa/epic-ps/mlueken/ufs-srweather-app/expt_dirs/community_ensemble_008mems
  status: CREATED
deactivate_tasks:
  expt_dir: /work/noaa/epic-ps/mlueken/ufs-srweather-app/expt_dirs/deactivate_tasks
  status: CREATED
inline_post:
  expt_dir: /work/noaa/epic-ps/mlueken/ufs-srweather-app/expt_dirs/inline_post
  status: CREATED
```
It seems as though the step to start the tests running didn't work correctly, which led the `FV3LAM_wflow.db` file not to be generated.  Do you have any idea why this might be occurring for the `.cicd/scripts/srw_test.sh` script?

mkavulich:
@MichaelLueken Based on what I'm seeing in the error messages you posted, it looks like there is a missing or uninitialized variable specifying where the .rocoto directory is. On Hera and Cheyenne this is the $HOME directory, maybe $HOME is not set on the Jenkins machine?

MichaelLueken:
@mkavulich Thanks for making the changes to allow `.cicd/scripts/srw_test.sh` to run the updated `run_WE2E_tests.py` script.  I'll see what I can find with respect to how rocoto is set for Jenkins.  As I understand it currently, there are Jenkins runners deployed on all Tier-1 machines, linked to the EPIC role accounts on the various machines.  I don't think that there is currently a variable for specifying where the .rocoto directory is.

MichaelLueken:
@mkavulich If you would like to try and replicate the errors by manually running the Jenkins scripts, you will need to set the following parameters (this example is for Orion, but are still used for other machines as well):

```
export SRW_PLATFORM="orion"
export SRW_COMPILER="intel"
export WORKSPACE="/work/noaa/epic-ps/mlueken/ufs-srweather-app"
export SRW_WE2E_COMPREHENSIVE_TESTS="False"
export SRW_PROJECT="epic-ps"
```

Where `SRW_PLATFORM` is the machine you are running on, `SRW_COMPILER` is the compiler, `WORKSPACE` is the path to your local copy of ufs-srweather-app,  `SRW_WE2E_COMPREHENSIVE_TESTS` is a flag used to turn on or off the comprehensive tests, and `SRW_PROJECT` is the account.

Once these environmental variables are set, then you will need to rebuild the SRW App using `./srw_build.sh`.  Once built, you can submit the tests using `./srw_test.sh`.  Both of these commands will need to be used in the `.cicd/scripts` directory.

mkavulich:
@MichaelLueken Thanks for the instructions, I was just about to ask! I will let you know what I find.

mkavulich:
@MichaelLueken I have tracked down the problem to [this line in the setup script](https://github.com/mkavulich/ufs-srweather-app/blob/feature/replace_sh_WE2E_script/tests/WE2E/setup_WE2E_tests.sh#L2) that seems to un-set the $HOME directory. It came from [a bug fix PR #459 by Daniel Abdi](https://github.com/ufs-community/ufs-srweather-app/pull/459/files#diff-7153880786409fadf07acb91061319b86e7c7a30b9d1f3bb16e2ea3222859a49), but I do not understand exactly what this line is doing.

@danielabdi-noaa Do you remember what this line does and/or why it was included? Removing this line seems to fix the problem we're seeing.

mkavulich:
@MichaelLueken I confirmed that removing that line results in a successful Jenkins test on Jet. We'll need to wait to hear back from @danielabdi-noaa as to what the purpose of that line is; we need to make sure at the very least that the `$HOME` variable is set in the environment when running the script; this appears to be a limitation of rocoto, not our workflow.

danielabdi-noaa:
@mkavulich If I recall correctly, this line is added so that the `setup_WE2E_tests.sh` script gets a clean environment smiliar to one you would get with slurm's `--export=NONE`. Without it the setup script failed when you execute it from a shell where the conda environment is activated i.e.
```
(regional_workflow) $ ./setup_WE2_tests.sh
```
vs
```
$ ./setup_WE2_tests.sh
```
The first one was not working because you have conda activated twice, within and outside the setup script and paths were mixed up and it was not able to find `jinja2` or `yaml` IIRC. I can not find the chat we had with @MichaelLueken but there was an issue then for sure. Things may have changed since then with upgrades to modulefiles so you can try to remove it if you can't find other solution. Personally I like it since it will make sure the script behaves the same whether conda is activated outside the script or not.

mkavulich:
@danielabdi-noaa Thanks for the info! With that in mind, I was able to implement a fix that still keeps a clean environment, just passing $HOME as an argument to be exported later. This has allowed the Jenkins script to run successfully on Jet!

@MichaelLueken You can try the script again now and it should complete successfully. Just note that if you run on Hera you may see random failures; the reason for this is due to an NCO mode problem described in an issue I just opened (#652)

MichaelLueken:
@mkavulich and @danielabdi-noaa Thank you both very much for identifying the cause of the issue and correcting it!  The manual run of the Jenkins test script I ran on Orion successfully runs now.

MichaelLueken (approved):
@mkavulich Manually running `run_WE2E_tests.py`, manually running the Jenkins test script - `.cicd/scripts/srw_test.sh`, and relaunching tests using `monitor_jobs.py` were all successful.  Approving this work now.

MichaelLueken:
Please note that the Jenkins failure is due to issue #635.  The Jenkins test is failing for Orion.  I have manually submitted the tests on Orion and they have successfully passed, as noted above.

MichaelLueken:
@mkavulich Additional issues have been encountered with the Jenkins tests.

During the tarring of the log files at the end of the tests, the following is being encountered:

`*/log.launch_FV3LAM_wflow: Cannot stat: No such file or directory`

This is causing the testing phase to fail, since this file is required in the tarring step.

The Jenkinsfile includes the following line:
https://github.com/ufs-community/ufs-srweather-app/blob/ff6f103b451bb0886b90afdc4a2a7a80f877c36d/.cicd/Jenkinsfile#L180

Since `log.launch_FV3LAM_wflow` is no longer part of the workflow, this entry on line 180 of the .cicd/Jenkinsfile will need to be removed.

mkavulich:
@MichaelLueken thanks for the info, I removed the now-non-existent log file from the archive, and also added new log and status files to the archive. Let me know if this still does not work.

tests/WE2E/setup_WE2E_tests.sh
```diff
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
-[ -n "$HOME" ] && exec -c "$0" "$@"
+
+# `exec -c` runs this script with clean environment; this avoids some problems
+# with double-loading conda environments. Since we do need $HOME to be set for
+# rocoto to run properly, pass it as an argument and export it later
+
```
danielabdi-noaa, line 6: Ok good fix!

tests/WE2E/setup_WE2E_tests.sh
```diff
@@ -26,8 +31,9 @@
 
 function usage {
   echo
-  echo "Usage: $0 machine account [compiler] [test_type] [others] | -h"
+  echo "Usage: $0 homedir machine account [compiler] [test_type] [others] | -h"
   echo
+  echo "       homedir       [required] user's home directory; this space must be writable"
```
danielabdi-noaa: I don't think the user needs to know about the implictly passed $HOME directory, so probably remove this and the one above on the `Usage: $0 homedir` ?

tests/WE2E/setup_WE2E_tests.sh
```diff
@@ -42,43 +48,54 @@ function usage {
...
+machine=${2,,}
+account=$3
+compiler=${4:-intel}
+test_type=${5:-fundamental}
+
+echo $homedir
+echo $machine
+echo $account
+echo $compiler
+echo $test_type
```
danielabdi-noaa: These are probably debugging prints?

.cicd/scripts/srw_test.sh
```diff
@@ -38,58 +38,15 @@ else
...
-# Allow we2e cron jobs time to complete and clean up themselves
-# TODO: Create parameter that sets the interval for the we2e cron jobs; this
-# value should be some factor of that interval to ensure the cron jobs execute
-# before the workspace is cleaned up.
-if [[ "${platform}" != "cheyenne" ]]; then
-    sleep 600
-fi
+./setup_WE2E_tests.sh ${platform} ${SRW_PROJECT} ${SRW_COMPILER} ${test_type} \
+    --expt_basedir=${we2e_experiment_base_dir} \
+    --opsroot=${nco_dir} | tee ${progress_file}
```
danielabdi-noaa, line 46: I don't know how the new `run_WE2E` works but just wanted to make sure the various delays are taken care of
- initial delay of 300 sec 
- polling frequency of 60 sec
- after completion delay of 600 sec. This one is to make sure jenkins does not delete experiment directory prematurely.

tests/WE2E/setup_WE2E_tests.sh
```diff
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
```
danielabdi-noaa, line 1: Could you update the usage comments of this script below? I was able to run fundamental tests witch
```
./setup_WE2E_tests.sh hera zrtrr intel fundamental
```
But not a custom test file `custom.hera.intel.com`
```
./setup_WE2E_tests.sh hera zrtrr intel custom
```

.cicd/scripts/srw_test.sh
mkavulich: Unless I'm missing something, none of those delays should be relevant with the new script. 

1. The `./run_WE2E_tests.py` script takes care of both the setting up and monitoring of tests, so no initial delay is needed
2. The new script serially cycles through active experiments with no delay
3. The new script will not finish until all tasks have been confirmed complete. 

One thing that may need to change is that for larger tests (such as, if the comprehensive tests are re-enabled), it may be worth using the new `-p` parallel option. Because the Jenkins tests call the script with `-d`, calls to rocotorun are fairly slow, taking ~30 seconds each (and there are two calls to rocotorun per test check). But this should be done cautiously depending on the platform: too many parallel tasks may overload the login node if that's where this script is running.

tests/WE2E/setup_WE2E_tests.sh
mkavulich: The script should be able to work with any custom file name, suite name, or test name; I confirmed this worked through the `./setup_WE2E_tests.sh` script with my own test on Jet. Can you send me the command that is failing for you and the directory? The above two commands appear to be the same.

danielabdi-noaa: @mkavulich I think test files like `custom.hera.intel.com` are not recognized because such patterns are only parsed for `fundamental` and `comprehensive` tests according to the logic in line 72 of `run_WE2E_tests.py`
```
$ ./setup_WE2E_tests.sh hera zrtrr intel custom
/home/Daniel.Abdi
hera
zrtrr
intel
custom

Modules based on Lua: Version 8.5.2  2021-05-12 12:44 -05:00
    by Robert McLay mclay@tacc.utexas.edu


Currently Loaded Modules:
  1) rocoto/1.3.3   2) miniconda3/4.12.0   3) wflow_hera

 

Logging set up successfully
Checking if ['custom'] is a valid test name
Checking if ['custom'] is a valid test suite
Checking if ['custom'] is a file containing test names

*********************************************************************
FATAL ERROR:
Experiment generation failed. See the error message(s) printed below.
For more detailed information, check the log file from the workflow
generation script: log.run_WE2E_tests
*********************************************************************

  File "/scratch2/BMC/gsd-hpcs/Daniel.Abdi/ufs-srweather-app/tests/WE2E/./run_WE2E_tests.py", line 534, in <module>
    run_we2e_tests(homedir,args)
  File "/scratch2/BMC/gsd-hpcs/Daniel.Abdi/ufs-srweather-app/tests/WE2E/./run_WE2E_tests.py", line 109, in run_we2e_tests
    raise FileNotFoundError(dedent(f"""
FileNotFoundError: 
The specified 'tests' argument '['custom']'
does not appear to be a valid test name, a valid test suite, or a file
containing valid test names.

Check your inputs and try again.
```

.cicd/scripts/srw_test.sh
danielabdi-noaa: For the third one, note that the existing `./run_WE2E_Tests.sh` also does not return until all tasks are complete, so if it was needed for the shell script it may also be needed for the python script. I vaguely recall that server side jenkins code by @jessemcfarland  deletes experiment directories as soon as `srw_tests.sh` finishes. This could happen before all cronjobs are removed, since they need to be called one last time for them to be removed from the crontable.

The first one, initial delay, is required to given enough time for atleast one workflow to be launched. If `run_WE2E_tests.py` does not exit thinking all jobs completed, when in reality no jobs were launched after say 4 mins, then it should be fine.

The second one is ok i think since it looks like the behaviour of checking experiment status changed. The old `get_expt_status.sh` was time consuming so it checks status of experiments every minute and print to the screen.

mkavulich: >note that the existing ./run_WE2E_Tests.sh also does not return until all tasks are complete

The `launch_FV3LAM_wflow.sh` script *tried* to implement this but it was done in quite a hacky way, so there were instances where the script would assume the experiment was done when there were still jobs running (for example, if there was a failed task but other tasks were still running). The new implementation does not use that script (does not use cron jobs either) and so does not suffer from that problem. It will only exit if it has verified that all jobs are done, via reading the rocoto database file directly.

>If run_WE2E_tests.py does not exit thinking all jobs completed, when in reality no jobs were launched after say 4 mins, then it should be fine.

This is correct, the `run_WE2E_tests.py` both submits and monitors jobs, so unless something goes horribly wrong there is no possibility of "desync" there.

I won't promise that I have anticipated *all* potential problems, but if any of these problems you mention (or similar ones) happen in the future, I can fix it in the `run_WE2E_tests.py` script itself, we shouldn't rely on hand-wavy delays to avoid problems anyway.

tests/WE2E/setup_WE2E_tests.sh
mkavulich: @danielabdi-noaa Sorry, I still don't understand what behavior you are expecting. It appears you gave the script `custom` as an argument; if that's not a file name then I'm not aware of what functionality that's supposed to have.

danielabdi-noaa: The file I have is `custom.hera.intel.com` under `machine_suites` folder not `custom`. With the new `setup_WE2E_tests.sh`, none of the below two calls work
```
./setup_WE2E_tests.sh hera zrtrr intel custom.hera.intel.com
./setup_WE2E_tests.sh hera zrtrr intel custom
```
while with the old one, they both do. Line 72 of `run_WE2E_tests.py` is a new addition that was absent in the old one, and I believe is the cause of this problem. I can understand if the behaviour is changed now, but that should be documented if so.

.cicd/scripts/srw_test.sh
danielabdi-noaa: Ok I did not realize cron is no longer being used. I agree the initial/final delays are not required in this case.

tests/WE2E/setup_WE2E_tests.sh
mkavulich: Thanks for the explanation, sorry it took me a while to understand. The way you are using the old script is not documented (at least, not anywhere I'm aware of) so I did not account for it when consolidating the three test arguments (`test_type`, `tests_file`, `test_name`) into a single argument (`tests`). I would argue that your use is unintended behavior and should not be carried forward to the new script. 

The documented/intended way to specify a custom test list file is to provide the full or relative path location of the file name you wish to read. So to run a file named `custom.hera.intel.com` under `machine_suites`, you would run

`./setup_WE2E_tests.sh hera zrtrr intel machine_suites/custom.hera.intel.com`

dshawul: @mkavulich Ok sounds good. In that case the usage of the script should be modified accordingly
```
echo "       test_type     [optional] test type: fundamental or comprehensive or all or any other name"
```
The `test_type` can not be "any other name" now, and also due to the consoldation of the three arguments maybe add also that you can pass a specific test case name or a path to a file containing list of tests.

mkavulich: Thanks Daniel, I have updated the usage statement in `./setup_WE2E_tests.sh` to accurately reflect the new argument behavior

tests/WE2E/WE2E_summary.py
```diff
@@ -0,0 +1,71 @@
...
+    if debug:
+        console.setLevel(logging.DEBUG)
+    else:
+        console.setLevel(logging.INFO)
+    logging.getLogger().addHandler(console)
+    logging.debug("Logging set up successfully")
+
+
+
+if __name__ == "__main__":
```
christinaholtNOAA, line 32: It's been helpful for testing other scripts if all the stuff in the "main" part of the script goes into a function, and the only thing the script does is call that function. For example:

```
def create_summary(args):
     # All the stuff below this line


if __name__ == "__main__":
    create_summary(sys.argv[1:])
```

tests/WE2E/monitor_jobs.py
```diff
@@ -3,217 +3,100 @@
...
 
-from python_utils import (
-    load_config_file,
-    cfg_to_yaml_str
-)
+from python_utils import load_config_file
 
 from check_python_version import check_python_version
 
+from WE2E_summary import print_WE2E_summary
```
christinaholtNOAA: `print_WE2E_summary` seemed to be imported from `utils` in this script above.

tests/WE2E/run_WE2E_tests.py
```diff
@@ -162,32 +173,32 @@ def run_we2e_tests(homedir, args) -> None:
...
-        logging.debug(test_cfg['verification'])
+            test_cfg['verification'] = check_task_verification(test_cfg,machine_defaults,
+                                                               config_defaults)
 
-        logging.debug(f"Writing updated config.yaml for test {test_name}\nbased on specified command-line arguments:\n")
+        logging.debug(f"Writing updated config.yaml for test {test_name}\n"\
+                       "based on specified command-line arguments:\n")
         logging.debug(cfg_to_yaml_str(test_cfg))
-        with open(ushdir + "/config.yaml","w") as f:
+        with open(ushdir + "/config.yaml","w", encoding="utf-8") as f:
```
christinaholtNOAA: `os.path.join()` is more portable, in general.

tests/WE2E/run_WE2E_tests.py
```diff
@@ -283,121 +303,81 @@ def check_test(test: str) -> str:
...
+                does not exist."""))
+
+    # Different input data types have different directory structures; set data dir accordingly
+    if cfg_bcs[f'EXTRN_MDL_NAME_{I_OR_L}'] == 'FV3GFS':
+        if f'FV3GFS_FILE_FMT_{I_OR_L}' not in cfg_bcs:
+            cfg_bcs[f'FV3GFS_FILE_FMT_{I_OR_L}'] = \
+                    dflt[f'task_get_extrn_{ics_or_lbcs}'][f'FV3GFS_FILE_FMT_{I_OR_L}']
+        cfg_bcs[f'EXTRN_MDL_SOURCE_BASEDIR_{I_OR_L}'] = \
+                f"{mach['platform']['TEST_EXTRN_MDL_SOURCE_BASEDIR']}/"\
+                f"{cfg_bcs[f'EXTRN_MDL_NAME_{I_OR_L}']}/{cfg_bcs[f'FV3GFS_FILE_FMT_{I_OR_L}']}/${{yyyymmddhh}}"
```
christinaholtNOAA: There's a chance the `os.path.join()` route here would make this much more readable...and portable.

tests/WE2E/setup_WE2E_tests.sh
```diff
@@ -42,43 +48,46 @@ function usage {
...
 
 if [ "$1" = "-h" ] ; then usage ; fi
-[[ $# -le 1 ]] && usage
+[[ $# -le 2 ]] && usage
 
-machine=${1,,}
-account=$2
-compiler=${3:-intel}
-test_type=${4:-fundamental}
+homedir=$1
```
christinaholtNOAA, line 53: This shift in order of the input command line args doesn't seem to be reflected in the usage statement.

tests/WE2E/setup_WE2E_tests.sh
```diff
@@ -26,14 +31,15 @@
...
-  echo "       compiler      [optional] compiler used to build binaries (intel or gnu)"
-  echo "       test_type     [optional] test type: fundamental or comprehensive or all or any other name"
-  echo "       others        [optional] All other arguments are forwarded to run_WE2E_tests.sh"
-  echo "       -h            display this help"
+  echo "       machine   [required] is one of: ${machines[@]}"
+  echo "       account   [required] case sensitive name of the user-specific slurm account"
+  echo "       compiler  [optional] compiler used to build binaries (intel or gnu)"
+  echo "       tests     [optional] tests to run: can be a suite (all|comprehensive|fundamental)
+                            a filename, or a test name"
+  echo "       others    [optional] All other arguments are forwarded to run_WE2E_tests.sh"
```
christinaholtNOAA: This one should be .py now?

tests/WE2E/test_configs/wflow_features/config.specify_template_filenames.yaml
```diff
@@ -5,11 +5,11 @@ metadata:
...
   RUN_ENVIR: community
 workflow:
-  DATA_TABLE_TMPL_FN: data_table
+  DATA_TABLE_FN: data_table
   DIAG_TABLE_TMPL_FN: diag_table.FV3_GFS_v15p2
   FIELD_TABLE_TMPL_FN: field_table.FV3_GFS_v15p2
-  MODEL_CONFIG_TMPL_FN: model_configure
-  NEMS_CONFIG_TMPL_FN: nems.configure
+  MODEL_CONFIG_FN: model_configure
+  NEMS_CONFIG_FN: nems.configure
```
christinaholtNOAA, line 12: I'm curious if this test can just be deleted....it no longer serves the purpose since these values are the default values

tests/WE2E/utils.py
```diff
@@ -0,0 +1,584 @@
...
+        expt_details.append(f'{" "*40} | Status    | Walltime   | Core hours used')
+        expt_details.append('-'*REPORT_WIDTH)
+
+        for task in expts_dict[expt]:
+            # Skip non-task entries
+            if task in ["expt_dir","status"]:
+                continue
+            status = expts_dict[expt][task]["status"]
+            walltime = expts_dict[expt][task]["walltime"]
+            expt_details.append(f'{task[:40]:<40s}  {status:<12s} {walltime:>10.1f}')
```
christinaholtNOAA: Lots of 40s as magic numbers here if we every want to make that wider...

tests/WE2E/utils.py
```diff
@@ -0,0 +1,584 @@
...
+                # Users are charged for full use of nodes, so core hours = CPN * nodes * time in hrs
+                core_hours = cores_per_node * nnodes * expts_dict[expt][task]['walltime'] / 3600
+                expts_dict[expt][task]['exact_count'] = True
+            else:
+                # If we can't find the number of nodes, assume full usage (may undercount)
+                core_hours = expts_dict[expt][task]['cores'] * \
+                             expts_dict[expt][task]['walltime'] / 3600
+                expts_dict[expt][task]['exact_count'] = False
+            expts_dict[expt][task]['core_hours'] = round(core_hours,2)
+    return expts_dict
```
christinaholtNOAA, line 183: You're update `expts_dict` in place. No need to return it.

tests/WE2E/utils.py
```diff
@@ -0,0 +1,584 @@
...
+                           as DEAD, ERROR, or COMPLETE. Used for initial checks for experiments
+                           that may have been restarted.
+        debug   (bool):    Will capture all output from rocotorun. This will allow information such
+                           as job cards and job submit messages to appear in the log files, but can
+                           slow down the process drastically.
+        submit  (bool):    In addition to reading the rocoto database, script will advance the
+                           workflow by calling rocotorun. If simply generating a report, set this
+                           to False
+    Returns:
+        dict: The updated experiment dictionary.
```
christinaholtNOAA, line 252: No need to return the dict. It's being updated in place. (this applies to others below, too, but I'll quite harping here)

tests/WE2E/utils.py
```diff
@@ -0,0 +1,584 @@
...
+    output the updated dictionary.
+
+    parallelizes the call to update_expt_status across the given number of processes.
+    Making use of the python multiprocessing starmap functionality, takes
+
+    Args:
+        expts_dict (dict): A dictionary containing information for all experiments
+        procs      (int): The number of parallel processes
+        refresh   (bool): "Refresh" flag to pass to update_expt_status()
+        debug     (bool): Will capture all output from rocotorun. This will allow information such
```
christinaholtNOAA: Silly formatting pet peeve...either keep the colons aligned or use one space between variable name and the data type.

tests/WE2E/utils.py
```diff
@@ -0,0 +1,584 @@
...
+    Args:
+        txtfile (str): File name for test details file
+    """
+
+    testfiles = glob.glob('test_configs/**/config*.yaml', recursive=True)
+    testdict = dict()
+    links = dict()
+    for testfile in testfiles:
+        # Calculate relative cost of test based on config settings using legacy script
+        cost_array = calculate_cost(testfile)
```
christinaholtNOAA, line 420: Is this array an array or a list? Maybe even a list of lists?

There are a bunch of numbers referencing list items in this section. Perhaps it's helpful to the user here to see a comment on how you expect the cost_array to be structured? Or even split them out into variables like this to improve readability:
```
test_cost, _, baseline_cost = cost_array[1:3]
cost = test_cost / baseline_cost
```

tests/WE2E/utils.py
```diff
@@ -0,0 +1,584 @@
...
+                nf = diffh // testdict[testname]['workflow']["INCR_CYCL_FREQ"]
+                testdict[testname]["num_fcsts"] = nf
+            else:
+                testdict[testname]["num_fcsts"] = 1
+
+    # For each found link, add its info to the appropriate test dictionary entry
+    for key in links.keys():
+        link = links[key]
+        testdict[link[2]]["alternate_name"] = link[0]
+        testdict[link[2]]["alternate_directory_name"] = link[1]
```
christinaholtNOAA: This could be made a bit more readable:

```
for key, link in links.items():
    alt_name, alt_dir, link_name = link
    .....
```

tests/WE2E/utils.py
```diff
@@ -0,0 +1,584 @@
...
+            f.write(f"{d}" + get_or_print_blank(testdict[expt],'task_get_extrn_lbcs',
+                                                'EXTRN_MDL_NAME_LBCS'))
+            f.write(f"{d}" + get_or_print_blank(testdict[expt],'workflow','DATE_FIRST_CYCL'))
+            f.write(f"{d}" + get_or_print_blank(testdict[expt],'workflow','DATE_LAST_CYCL'))
+            f.write(f"{d}" + get_or_print_blank(testdict[expt],'workflow','INCR_CYCL_FREQ'))
+            f.write(f"{d}" + get_or_print_blank(testdict[expt],'workflow','FCST_LEN_HRS'))
+            f.write(f"{d}" + get_or_print_blank(testdict[expt],'task_run_fcst','DT_ATMOS'))
+            f.write(f"{d}" + get_or_print_blank(testdict[expt],'task_get_extrn_lbcs',
+                                                'LBC_SPEC_INTVL_HRS'))
+            f.write(f"{d}" + get_or_print_blank(testdict[expt],'global','NUM_ENS_MEMBERS') + "\n")
```
christinaholtNOAA: This bit is pretty clunky. Just to avoid repetition and inefficient calls to a function that does what a built-in does, you might consider:

```
key_pairs = [
 ('workflow', 'PREDEF_GRID_NAME'),
 ('workflow', 'CCPP_PHYS_SUITE'),
  ...
  ]

for key1, key2 in key_pairs:
    f.write(f"{d} {testdict[expt].get(key1, {}).get(key2, '')}"
```

tests/WE2E/utils.py
```diff
@@ -0,0 +1,584 @@
...
+def compare_rocotostat(expt_dict,name):
+    """Reads the dictionary showing the location of a given experiment, runs a `rocotostat` command
+    to get the full set of tasks for the experiment, and compares the two to see if there are any
+    unsubmitted tasks remaining.
+    """
+
+    # Call rocotostat and store output
+    rocoto_db = f"{expt_dict['expt_dir']}/FV3LAM_wflow.db"
+    rocotorun_cmd = ["rocotostat", f"-w {expt_dict['expt_dir']}/FV3LAM_wflow.xml",
+                     f"-d {rocoto_db}", "-v 10"]
```
christinaholtNOAA: Why not pull the xml path out for similar readability?

christinaholtNOAA (commented):
Sorry for the delayed review. I left some comments/questions below.

gsketefian:
@mkavulich I looked through the changes and they seem fine.  I have a few questions that I could find the answers to by carefully looking at the code, but I figure just asking is much faster:
1) With the current version of `run_WE2E_tests.py` and accompanying scripts, I've noticed that while the script is re-submitting jobs, if I go into an experiment directory and call `rocotorun` or `launch_FV3LAM_wflow.sh`, I get the following error:
```
$ ./launch_FV3LAM_wflow.sh 

Modules based on Lua: Version 8.5.2  2021-05-12 12:44 -05:00
    by Robert McLay mclay@tacc.utexas.edu

Currently Loaded Modules:
  1) rocoto/1.3.3   2) miniconda3/4.12.0   3) wflow_hera

FATAL ERROR: 
ERROR:
  From script:  "launch_FV3LAM_wflow.sh"
  Full path to script:  "/scratch2/BMC/det/Gerard.Ketefian/UFS_CAM/DTC_ensemble_task_pythonize/ufs-srweather-app/ush/launch_FV3LAM_wflow.sh"
Call to "rocotorun" failed with return code 1.
Exiting with nonzero status.
```
This doesn't always happen, and it may depend on the task the experiment is currently on.  Have you encountered this, and do you know if the changes in this PR get around this?
2) Is the `launch_FV3LAM_wflow.sh` script being kept, and if not, is there a replacement for it that can be called from within an experiment directory (maybe via a symlink) that will relaunch the workflow?  This is handy for me when I'm debugging a single task and want to relaunch it over and over again and get its status.  
3) Is `WE2E_summary.py` the replacement for `get_expts_status.sh`?
Thanks.

mkavulich:
@gsketefian Thanks for your comments and questions.

1. This is a limitation of the `launch_FV3LAM_wflow.sh` script and how it interacts with rocoto. Rocoto locks the `.db` file while rocotorun is being run, so if you try to run two rocotorun processes simultaneously, the second will fail with the message `WARNING: Workflow is locked by pid xxxxxx on host xx.xxx.xx.xx` and exit code 1. So if the `./run_WE2E_tests.py` or `./monitor_jobs.py` scripts are using rocotorun while you run that script, it will fail. The same thing would happen if you tried to run `launch_FV3LAM_wflow` twice simultaneously. Is there a use case I'm missing that would require running `launch_FV3LAM_wflow` while the `monitor_jobs` function is actively monitoring/submitting jobs?
2. `launch_FV3LAM_wflow.sh` is not being removed with this PR (I was only addressing the WE2E tests in this PR, and I think this script has uses outside of that testing framework), but if you want to get the status of a job `WE2E_summary.py` will give you the summary of the currently running jobs (you can give it a yaml file or a top-level experiment directory as an argument). 
If you need the ability to see the status of an active experiment while you're in that directory, is there a reason that simply calling `rocotostat` doesn't give you want you need?
3. `WE2E_summary.py` is not a 1-to-1 replacement of `get_expts_status.sh`, but it should give the same relevant information. It does not not have the `num_log_lines` argument (because it reads rocoto `.db` files directly rather than relying on parsing a log file) and it does not have the `launch_wflows` argument (`WE2E_summary.py` does not advance the workflow with rocotorun, it only reads and reports the current status). 
If the ability to manually advance the workflow one step at a time is needed it can be added. But I was not anticipating that use being necessary.

gsketefian:
@mkavulich:
1. A use case is when you launch a bunch of tests with `run_WE2E_tests.py`, and say one of them starts to fail (some tasks report back as failed), and you want to stop resubmitting tasks for that experiment and go to its directory and run one task at a time to debug (while allowing the other experiment to continue).  Do `run_WE2E_tests.py` and `monitor_jobs.py` lock the `.db` files of all experiments while they're running?  Or do they read it, free it for a bit, then read it again, etc?  I have a feeling it's the latter since I don't always get that error.
2. As I step through the tasks, I like to get the most updated status of the experiments, so I like to run `launch_FV3LAM_wflow.sh` because it calls `rocotorun` before `rocotostat` and stores the results in a log file, which is helpful if I want to see the progress (e.g. what tasks were called from one re-launch to the next).  I realize these can all be done from the command line, but it's nice to have one script do it all.  As long as `launch_FV3LAM_wflow.sh` is not being removed, it should be fine.
3. Sounds like `WE2E_summary.py ` will be work for my purposes.

Currently, if I want to step one task at a time with an experiment is to first create it with `run_WE2E_tests.py`, then `Ctr-C` out of it once it's done creating the experiment directories, then go to the experiment directory and use `launch_FV3LAM_wflow.sh` to go one or two tasks at a time while I debug the xml, the ex-script, j-job, METplus conf file, etc.  I'll just keep using that approach.
Thanks!

tests/WE2E/setup_WE2E_tests.sh
mkavulich: This was changed per @danielabdi-noaa's comments; since this script is re-called automatically with homedir and it is not a user-supplied argument, the usage statement shouldn't include it

tests/WE2E/test_configs/wflow_features/config.specify_template_filenames.yaml
mkavulich: This test will be removed in my next PR (related to Issue #587); a few test updates are needed like this, and I'd rather keep them compartmentalized for easier review. These changes were simply to let this test actually pass if run.

mkavulich:
@gsketefian To answer your first question, the new scripts cycle through each experiment serially, running rocotorun on one experiment at a time, with a 5 second delay between loops. It is only while rocotorun is being called that the rocoto database is locked.

As a side note, if one experiment fails, then the monitor_jobs function will not continue running rocotorun in that directory. So if that is your use case you shouldn't see conflicts when using the launch_FV3LAM_wflow.sh script unless something weird is happening.

tests/WE2E/utils.py
christinaholtNOAA: Oops. This is fine.

tests/WE2E/utils.py
mkavulich: This was a product of legacy code that I did not feel was in the scope of this PR to address. Ideally `calculate_cost` should be completely re-written (I already had to make significant changes to get it working again) but this would require a lot of effort that I feel is best placed elsewhere

tests/WE2E/utils.py
mkavulich: Trying to implement this resulted in breaking the parallel (starmap) calls to this function. I'd rather leave this how it is since it's technically correct.

mkavulich:
@christinaholtNOAA I believe I have addressed all your comments. Let me know if you have anything further.

tests/WE2E/WE2E_summary.py
christinaholtNOAA: This one doesn't seem to have been addressed.

christinaholtNOAA (commented):
LGTM. Just one comment below.

tests/WE2E/WE2E_summary.py
mkavulich: I re-pushed the change here: https://github.com/ufs-community/ufs-srweather-app/pull/637/commits/8f5525d152965511d65b17a575d973ac9d1d6626, it was un-intentionally reverted.

christinaholtNOAA: After a quick in person chat, I think I'm caught up. This suggestion was an organization suggestion for ease of future testing, and is out of scope for this PR given that many other scripts with user interfaces need similar care.

christinaholtNOAA (approved):
LGTM.

MichaelLueken:
@mkavulich All Jenkins tests successfully passed. The Orion tests ultimately sat in queue for several hours (over eleven hours), leading to a time-out for the Build and Test check.  Will now move forward with merging this PR.

PR #645, opened 2023-03-07
[develop] Add upgrade option to check_for_preexist_dir.

### LINKED ISSUE #655: Upgrade EXPTDIR directory option.
## Description

 
`RRFS_dev1` added a new option "upgrade", besides "delete","rename","quit" when checking for preexisting experiment or other types of directories. Essentially, what it does is the following
 - If it is an EXPTDIR, then it does an `rsync` on the EXPTDIR also keeping status of rocoto jobs, which will not be possible with the other options.
 - For other directories e.g. `orog`,`grid` directories, the same directory is used. So if those tasks are run again, existing files will be overwritten

## Solution

Add the "upgrade" option to `check_for_preexist_dir.py` and `check_for_preexist_dir.sh`. Currently the former is exclusively used for renaming EXPTDIR, so it is modified to do an `rsync`, while the latter simply re-uses the existing directory.

## Requirements**

None

Add "upgrade" option to the python/bash scripts checking for preexisting directories.

## Dependencies (optional)

None

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 

Addresses issue #655. Add `upgrade` option besides `delete`, `rename`, `quit` when checking for existing experiment directory (or another folder). For EXPTDIR the behavior is different from other directories. The "upgrade" option behaves like `rsync` for `EXPTDIR`. The workflow generation is handled by python code in SRW app, therefore `check_for_preexist_dir.py` handles that. For other directories, the shell version of the script is used, and `upgrade` means reuse existing directory. Apparently this option is sometimes useful for RRFS_dev1 workflow runs. 

### Type of change

## TESTS CONDUCTED: 

- [X] hera.intel
- [ ] orion.intel
- [ ] cheyenne.intel
- [ ] cheyenne.gnu
- [ ] gaea.intel
- [ ] jet.intel
- [ ] wcoss2.intel
- [ ] Jenkins
- [ ] fundamental test suite

## ISSUE: 

- #655 

## CHECKLIST

## LABELS (optional): 

- [ ] Work In Progress
- [ ] bug
- [ ] enhancement
- [ ] documentation
- [ ] release
- [ ] high priority
- [ ] run_ci
- [ ] Needs Cheyenne test 
- [ ] Needs Jet test 
- [ ] Needs Hera test 
- [ ] Needs Orion test 
- [ ] help wanted


ush/config_defaults.yaml
```diff
@@ -842,6 +842,10 @@ workflow:
   #   the suffix "_oldNNN", where NNN is a 3-digit integer chosen to make
   #   the new name unique.
   #
+  # * "upgrade":
```
gsketefian: @danielabdi-noaa This description is somewhat unclear to me.  So if "upgrade" is specified, I see below that a copy is made using `rsync`, but then what is upgraded?  What does it mean to "upgrade a directory"?  Also, you mention `EXPTDIR`, but it could be any directory, right?  Can you clarify a bit?  Thanks.

danielabdi-noaa: @gsketefian Here is a better description of the "upgrade" feature. I should probably add the full comment from there in this PR.  
https://github.com/NOAA-GSL/regional_workflow/blob/feature/RRFS_dev1/ush/bash_utils/check_for_preexist_dir_file.sh#L112-L152

The implementation in `RRFS_dev1` checks whether the directory whose existence we are checking is EXPTDIR or not, and then behaves differently. If it is an EXPTDIR, its behaviour is equivalent to the "rsync". In other cases it re-uses the existing directoy. This feature is apparently useful to keep rocoto jobs running but one can choose to use "rename" in all cases.

Note in SRW we have two version of `check_preexisting_dir` (python and bash version). The python version is explicitly used for renaming EXPTDIR so we don't need to check it it is another kind of directory, so it effectively does `rsync`. The bash version is used in ex-scripts/j-jobs (where we don't rename EXPTDIRs) so it keeps the existing directory

**Edit:** I mixed up somethings in my previous explanation, hopefully, it is now clearer. Also linked an issue #655

gsketefian: @danielabdi-noaa Thanks for the explanation.  It's much clearer now.  Some additional questions:
1) It seems like in the python script, what the "upgrade" option does is to make a backup copy of the directory while keeping the original.  Is that correct?  If so:
  a) Wouldn't it suffice to just use `cp` instead of `rsync`?
  b) Why not call the option something like "backup" instead of "upgrade" since "upgrade" sounds like something is being added to/updated in the original directory, which it isn't afa I can tell?
2) Why keep both python and shell scripts?  Why not just keep the python (and create a shell wrapper if necessary), and call it with "upgrade" (or "backup") when operating on `EXPTDIR` and with "delete" when operating on "grid", "orog", etc (or not call the script at all for "grid", "orog", ...)?

danielabdi-noaa: @gsketefian 
1) I suppose it can be reduced to `cp` in most cases, but maybe they wanted to add syncing experiment directory from remote machine? Will have to ask the original author about that. About the naming, yes the same directory contents are kept as before in both EXPTDIR and other folder types, so the name "upgrade"/"update" was probably meant to reflect what happens after that. I personally like "reuse" as the name, let me know what you think about it?

2) I don't like keeping around two versions of the same scripts (bash and python) either. Maybe we can move to using only the python script in a future PR?

gsketefian: @danielabdi-noaa Thanks again for the explanations.  So:
1) "reuse" sounds good to me.
2) Do you mind creating an issue to remove the shell version of the script?  I just want to make sure we don't forget about it.  

Otherwise looks good, thanks.

gsketefian (approved):

ush/config_defaults.yaml
danielabdi-noaa: @gsketefian Thanks, I have updated the name from "upgrade" to "reuse", and also created an issue #696 to address the duplicate shell/python functionality.

MichaelLueken (approved):
@danielabdi-noaa These changes look good to me!  I was able to clone this work on Jet, submit the fundamental WE2E tests, then reran them using the new reuse method.  The new method worked as described in the PR.  All tests for both runs successfully passed.  Approving these changes now.

MichaelLueken:
@danielabdi-noaa The Jenkins tests on Cheyenne Intel have finally passed successfully.  Moving forward with merging this work now.

PR #540, opened 2023-01-16
[develop] Incorporate RRFS_dev1 workflow to SRW and import most tasks

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 

I am opening up this PR for feedback per our discussion today. Will keep on working on it or a second PR but it is probably time
to get feedback on it. The tasks for both spinup (cold and warm start) and production cycles pass on CONUS domain 3km resolution as shown below. Although I am pretty sure there is a lot more work to do with other test cases I haven't considered, the new code that is constantly coming from RRFS_dev1 etc. So this is just a start to get atleast a test case running partially and try to keep up with the changes from RRFS_dev1.

It should pass fundamental tests now -- tested on Hera successfully.

Addresses the main issue of incorporating RRFS_dev1 workflow
  - #538

And also adds the following tasks that have successfully passed for the specific test case I considered
  - #486
  - #489
  - #546 
  - #545 
  - #543 
  - #542 
  - #490
  - #491

The `run_archive` test although imported can only be tested with someone with write privilege on HPSS. 

Tagging RRFS_dev1 developers for feedback: @hu5970 @EricJames-NOAA @chunhuazhou @ShunLiu-NOAA @TrevorAlcott-NOAA @guoqing-noaa @hongli-wang @terraladwig @daviddowellNOAA
## Result
Result on the `config.rrfs.yaml` file for the CONUS domain on 3km resolution
```
       CYCLE                    TASK                       JOBID               STATE         EXIT STATUS     TRIES      DURATION
================================================================================================================================
202107240000          get_extrn_lbcs                    41445542           SUCCEEDED                   0         1          20.0
202107240000            make_lbcs_00                    41445663           SUCCEEDED                   0         1         154.0
202107240000            make_lbcs_01                    41445664           SUCCEEDED                   0         1         151.0
202107240000            make_lbcs_02                    41445665           SUCCEEDED                   0         1         154.0
202107240000            make_lbcs_03                    41445667           SUCCEEDED                   0         1         151.0
202107240000            make_lbcs_04                    41445668           SUCCEEDED                   0         1         152.0
202107240000            make_lbcs_05                    41445669           SUCCEEDED                   0         1         152.0
202107240000            make_lbcs_06                    41445670           SUCCEEDED                   0         1         155.0
202107240000            make_lbcs_07                    41445671           SUCCEEDED                   0         1         152.0
202107240000            make_lbcs_08                    41445673           SUCCEEDED                   0         1         151.0
[...]

PR #638, opened 2023-03-06
[develop] Add radar reflectivity task.

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 

Add `process_radarref` task from `RRFS_dev1`.
Note that there is no workflow so this PR can not be run independently until then.
However, it did successfully run for both spinup/production cycles in PR #540.

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## TESTS CONDUCTED: 

- [X] hera.intel
  The CONUS 3km test case run successfully with this task in PR #540 
- [ ] orion.intel
- [ ] cheyenne.intel
- [ ] cheyenne.gnu
- [ ] gaea.intel
- [ ] jet.intel
- [ ] wcoss2.intel
- [ ] NOAA Cloud (indicate which platform)
- [ ] Jenkins
- [ ] fundamental test suite
- [ ] comprehensive tests (specify *which* if a subset was used)

## ISSUE: 

- #489 

## CHECKLIST

- [X] My code follows the style guidelines in the Contributor's Guide
- [ ] I have performed a self-review of my own code using the Code Reviewer's Guide
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes need updates to the documentation. I have made corresponding changes to the documentation
- [ ] My changes do not require updates to the documentation (explain).
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] Any dependent changes have been merged and published

## LABELS (optional): 

A Code Manager needs to add the following labels to this PR: 
- [ ] Work In Progress
- [ ] bug
- [ ] enhancement
- [ ] documentation
- [ ] release
- [ ] high priority
- [ ] run_ci
- [ ] run_we2e_fundamental_tests
- [ ] run_we2e_comprehensive_tests
- [ ] Needs Cheyenne test 
- [ ] Needs Jet test 
- [ ] Needs Hera test 
- [ ] Needs Orion test 
- [ ] help wanted

PR #647, opened 2023-03-07
[develop] Add new RRFS variables such as NWGES and workflow control variables

### LINKED ISSUE #654: Define shared variables and predefined grids for all tasks
## Description

 
`RRFS_dev1` tasks share a number of variables and predefined grids, so I believe these should go in before any task specific PR. I don't think it is appropriate to add these common things with a PR meant specifically for a given task. Examples:
  - new NCO directory names `NWGES`, `RRFSE_NWGES`, `ENSCTRL_COMIN`
  - `TAG` variable that is prepended to all task names so that one can identify tasks in multiple workflows
  - Workflow control flags such as `DO_SPINUP`, `DO_DACYCLE` etc 
  - `RRFS_dev1` defines a bunch of new cycles besides `forecast`, `initial` etc,
    
    ```
    AT_START_CYCLEDEF: "00 01 01 01 2100 *"
    INITIAL_CYCLEDEF: "00 01 01 01 2100 *"
    BOUNDARY_CYCLEDEF: "00 01 01 01 2100 *"
    BOUNDARY_LONG_CYCLEDEF: "00 01 01 01 2100 *"
    SPINUP_CYCLEDEF: "00 01 01 01 2100 *"
    PROD_CYCLEDEF: "00 01 01 01 2100 *"
    PRODLONG_CYCLEDEF: "00 01 01 01 2100 *"
    RECENTER_CYCLEDEF: "00 01 01 01 2100 *"
    ARCHIVE_CYCLEDEF: "00 01 01 01 2100 *"
    SAVEINPUT_CYCLEDEF: "00 01 01 01 2100 *"
    ```
 - Add namelists for DA cycle 
 - Maybe also add the workflow and test case? That will make sure the PRs are tested as they are added sequentially.
 
I think this PR is high priority because of this since all other tasks will depend on it. It can also serve as a base to develop the new yaml workflow since the cycles, workflow controls, new directories etc are all required

## Solution

Modify `config_defaults.yaml` and `predef_grid_params.yaml` to include the above variables and new predefined grids respectively.

## Requirements**

None

Has the common variables for all tasks and new predefined grids.

## Dependencies (optional)

None

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 

Addresses issue #654. 
- General (non-task specific) variables are added to `config_defaults.yaml` - New directory names `NWGES`, `RRFSE_NWGES`, `ENSCTRL_COMIN` etc, 
- new variables such as `TAG`. 
- Workflow control flags such as `DO_SPINUP`, `DO_DACYCLE`, `DO_SOIL_ADJUST` 
- Also `RRFS_dev1` defines a bunch of new cycles besides `forecast`, `initial` etc, so those are added too.
- Modifcations in `generate_FV3LAM_wflow.py` needed for creating  input namelist files for DA cycles
- Modifications to machine files, mostly worfklow related, but on jet/hera paths to data are also added

I think this PR is priority number 1 because of this since all tasks will depend on it, and can serve as a central place to coordinate workflow variables, common task variables etc.

### Type of change

## TESTS CONDUCTED: 

- [X] hera.intel
RRFS conus 3km test case as run in PR #540.
- [ ] orion.intel
- [ ] cheyenne.intel
- [ ] cheyenne.gnu
- [ ] gaea.intel
- [ ] jet.intel
- [ ] wcoss2.intel
- [ ] Jenkins
- [ ] fundamental test suite

## ISSUE: 

- #654 

## CHECKLIST

## LABELS (optional): 

- [ ] Work In Progress
- [ ] bug
- [ ] enhancement
- [ ] documentation
- [ ] release
- [ ] high priority
- [ ] run_ci
- [ ] Needs Cheyenne test 
- [ ] Needs Jet test 
- [ ] Needs Hera test 
- [ ] Needs Orion test 
- [ ] help wanted


panll:
Tested on Hera and got running time errors for tasks getting ICS and LBC:
Exiting with nonzero status.
End exregional_get_extrn_mdl_files.sh at Mon Mar 13 04:08:12 UTC 2023 with error code 1 (time elapsed: 00:00:32)
FATAL ERROR:
ERROR:
  From script:  "JREGIONAL_GET_EXTRN_MDL_FILES"
  Full path to script:  "/scratch2/BMC/fv3lam/lpan/test/review/03102023/03122023/ufs-srweather-app/jobs/JREGIONAL_GET_EXTRN_MDL_FILES"
Call to ex-script corresponding to J-job "JREGIONAL_GET_EXTRN_MDL_FILES" failed.
Exiting with nonzero status.
End JREGIONAL_GET_EXTRN_MDL_FILES at Mon Mar 13 04:08:12 UTC 2023 with error code 1 (time elapsed: 00:00:32)

danielabdi-noaa:
@panll Thanks for testing. There was a bug introduced from NOAA-GSL fork that uses DEFAULT queue instead of HPSS queue for get_extrn_ics/lbcs. [See here.](https://github.com/NOAA-GSL/regional_workflow/blob/feature/RRFS_dev1/ush/templates/FV3LAM_wflow.xml#L672-L675). I will file an issue there later.
I was able to run upto `run_fcst`, however, the change from 64 to 65 levels causes an issue so `External.cfg` need to be updated, which I haven't included in this PR, to complete the community test case.
**Edit:** I was able to complete community test case on PR #540

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1266328943/20230313175016/ufs-srweather-app
Build was Successful
Rocoto jobs started
Long term tracking will be done on 10 experiments
ci-hera-intel-WE

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1266328943/20230314005021/ufs-srweather-app
Build was Successful
Rocoto jobs started
Long term tracking will be done on 10 experiments
ci-hera-intel-WE
Experiment Failed on hera: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
2023-03-14 01:32:14 +0000 :: hfe08 :: Task make_ics, jobid=42866017, in state DEAD (FAILED), ran for 17.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
2023-03-14 01:32:14 +0000 :: hfe08 :: Task make_lbcs_00, jobid=42866018, in state DEAD (FAILED), ran for 22.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: MET_ensemble_verification
2023-03-14 01:40:07 +0000 :: hfe08 :: Task make_sfc_climo, jobid=42866194, in state DEAD (FAILED), ran for 18.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_2017_gfdlmp_regional_plot
2023-03-14 01:40:08 +0000 :: hfe09 :: Task make_ics, jobid=42866185, in state DEAD (FAILED), ran for 19.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_2017_gfdlmp_regional_plot
2023-03-14 01:40:08 +0000 :: hfe09 :: Task make_lbcs_00, jobid=42866186, in state DEAD (FAILED), ran for 22.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
2023-03-14 01:40:10 +0000 :: hfe08 :: Task make_sfc_climo, jobid=42866187, in state DEAD (FAILED), ran for 17.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_HRRR
2023-03-14 01:40:07 +0000 :: hfe12 :: Task make_sfc_climo, jobid=42866188, in state DEAD (FAILED), ran for 18.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: pregen_grid_orog_sfc_climo
2023-03-14 01:40:12 +0000 :: hfe09 :: Task make_lbcs_00, jobid=42866033, in state DEAD (FAILED), ran for 16.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_RAP_suite_HRRR
2023-03-14 01:44:13 +0000 :: hfe05 :: Task make_ics, jobid=42866462, in state DEAD (FAILED), ran for 19.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_RAP_suite_HRRR
2023-03-14 01:44:13 +0000 :: hfe05 :: Task make_lbcs_00, jobid=42866463, in state DEAD (FAILED), ran for 20.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
2023-03-14 01:40:16 +0000 :: hfe03 :: Task make_ics, jobid=42866193, in state DEAD (FAILED), ran for 18.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v15p2
2023-03-14 01:44:05 +0000 :: hfe12 :: Task make_ics, jobid=42866416, in state DEAD (FAILED), ran for 18.0 seconds, exit status=256, try=2 (of 2)
Experiment Failed on hera: grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v15p2
2023-03-14 01:44:05 +0000 :: hfe12 :: Task make_lbcs_00, jobid=42866417, in state DEAD (FAILED), ran for 19.0 seconds, exit status=256, try=2 (of 2)
Experiment Succeeded on hera: community_ensemble_2mems_stoch
All experiments completed

danielabdi-noaa:
I believe the hera failures are due to issue #652. The `run_envir=nco` mode seems to be broken. Manually re-running the tests using community mode, all tests pass!

@panll Could you pleas try again to run the community test case with this PR? I have run it manually and it completes successfully.

danielabdi-noaa:
To help with the review of this PR I will leave some notes here. The purpose of the PR is described in issue #654 in detail. In short, what it does define common variables, common predef grids and capture the "initial phase" of `RRFS_dev1`. For example, different namelists are created with `./generate_workflow.py` as opposed to the single `input.nml` we had. The workflow is incorporated so you can generate `RRFS_dev1` workflow with the included `config.rrfs.yaml` sample file right right away even though there will not be no new tasks. The tasks to be added later are commented out currently, and can be activated one by one to test them. Some notes about how this affects current SRW workflow. For example, here is a successfully run of `custom_GFDLgrid` test case run in `nco` mode
```
$ rocotostat -v 10 -w FV3LAM_wflow.xml -d FV3LAM_wflow.db                                                                                                    CYCLE                    TASK                       JOBID               STATE         EXIT STATUS     TRIES      DURATION
================================================================================================================================
201907010000               make_grid                    42894589           SUCCEEDED                   0         1          27.0
201907010000               make_orog                    42894911           SUCCEEDED                   0         1          75.0
201907010000          make_sfc_climo                    42895144           SUCCEEDED                   0         1          55.0
201907010000           get_extrn_ics                    42894590           SUCCEEDED                   0         1          12.0
201907010000          get_extrn_lbcs                    42894591           SUCCEEDED                   0         1          12.0
201907010000                make_ics                    42895166           SUCCEEDED                   0         1          69.0
201907010000            make_lbcs_00                    42895167           SUCCEEDED                   0         1         102.0
201907010000           run_fcst_prod                    42895259           SUCCEEDED                   0         1         231.0
201907010000           run_post_f000                    42895806           SUCCEEDED                   0         1          25.0
201907010000           run_post_f001                    42895807           SUCCEEDED                   0         1          24.0
201907010000           run_post_f002                    42895808           SUCCEEDED                   0         1          24.0
201907010000           run_post_f003                    42895907           SUCCEEDED                   0         1          26.0
201907010000           run_post_f004                    42895908           SUCCEEDED                   0         1          26.0
201907010000           run_post_f005                    42895909           SUCCEEDED                   0         1          27.0
201907010000           run_post_f006                    42895910           SUCCEEDED                   0         1          24.0
```
Notice that the forecast name is now `run_fcst_prod` instead of the old `run_fcst`. This is because I made the "forecast" cycle of existing SRW to be `prodcyc`. @chan-hoo  and @gsketefian Please check if the way AQM and new verification workflows are to your liking. I have also prepended all task names with an optional TAG variable.
Also `make_lbcs` has a number attached to it `00` because it will be parallelized with 10 workers (NOT included with this PR but the name remains in the workflow).

fix/fix_am/global_hyblev_fcst_rrfsL65.txt
```diff
@@ -0,0 +1,68 @@
...
+    3229.725  0.00000000
+    2681.893  0.00000000
+    2193.544  0.00000000
+    1758.738  0.00000000
+    1372.014  0.00000000
+    1028.378  0.00000000
+     723.285  0.00000000
+     452.613  0.00000000
+     212.637  0.00000000
+       0.000  0.00000000
```
danielabdi-noaa: This file is added because that is what `RRFS_dev1` uses but currently not activated. It will be activated in later PRs as it changes levels from `64` to `65`

parm/FV3LAM_wflow.xml
```diff
@@ -141,19 +189,39 @@ Directories and files.
...
+<!ENTITY FG_ROOT                  "{{ nwges_basedir }}">
+{%- endif %}
+
+<!-- Temporarily turn off fcst_fv3lam subdirectory usage -->
+<!-- until run_fcst of rrfs_dev is merged                -->
+{%- if do_rrfs_dev %}
+<!ENTITY FCST_DIR                  "fcst_fv3lam/">
+{%- else %}
+<!ENTITY FCST_DIR                  "">
+{%- endif %}
```
danielabdi-noaa: This will be removed later once everything is merged. If we start using `fcst_fv3lam` and `fcst_fv3lam_spinup` directories for the forecast output right from the beginning it would require modifying forecast/post etc steps from the very beginning making reviews complex. Only minimal changes are made in those scripts to account for the name change from `run_fcst` to `run_fcst_prod` in this PR.

scripts/exregional_run_fcst.sh
```diff
@@ -450,6 +450,24 @@ cat > itag <<EOF
...
+if [ "${DO_ENSEMBLE}" = "TRUE" ] && ([ "${DO_SPP}" = "TRUE" ] || [ "${DO_SPPT}" = "TRUE" ] || [ "${DO_SHUM}" = "TRUE" ] || \
+   [ "${DO_SKEB}" = "TRUE" ] || [ "${DO_LSM_SPP}" =  "TRUE" ]); then
+     STOCH="TRUE"
+fi
+if [ "${STOCH}" == "TRUE" ]; then
+  ln_vrfy -sf ${FV3_NML_STOCH_FP} ${DATA}/${FV3_NML_FN}
+ else
+  ln_vrfy -sf ${FV3_NML_FP} ${DATA}/${FV3_NML_FN}
+fi
+
```
danielabdi-noaa, line 458: This is needed to pick the right namelist file for stochastic physics. We currently directly modify the input namelist in SRW. This was needed to get `community_ensemble_2memss_stoch` working @JeffBeck-NOAA

scripts/exregional_run_post.sh
```diff
@@ -167,7 +167,7 @@ fi
 # Set the names of the forecast model's write-component output files.
 #
 if [ "${RUN_ENVIR}" = "nco" ]; then
-    DATAFCST=$DATAROOT/run_fcst${dot_ensmem/./_}.${share_pid}
+    DATAFCST=$DATAROOT/run_fcst_prod${dot_ensmem/./_}.${share_pid}
```
danielabdi-noaa: The name change I mentioned earlier.

ush/config.rrfs.yaml
```diff
@@ -0,0 +1,137 @@
...
+#  ADDNL_OUTPUT_GRIDS: []
+#task_run_anal:
+#  NCORES_RUN_ANAL: 240
+#  NCORES_RUN_OBSERVER: 80
+#task_run_ref2tten:
+#  PPN_RUN_REF2TTEN: 1
+#task_run_nonvarcldanl:
+#  PPN_RUN_NONVARCLDANL: '{{ task_run_fcst.IO_LAYOUT_Y }}'
+#task_run_graphics:
+#  TILE_LABELS: "full"
```
danielabdi-noaa: All of these will be activated one by one as we go along.

ush/config_defaults.yaml
```diff
@@ -51,9 +51,14 @@ user:
...
   # The account under which to submit jobs to the queue.
   #
+  # SERVICE_ACCOUNT:
+  # The account under which to submit non-reservation jobs to the queue.
+  # Defaults to ACCOUNT if not set.
+  #
   #-----------------------------------------------------------------------
   MACHINE: "BIG_COMPUTER"
   ACCOUNT: ""
+  SERVICE_ACCOUNT: '{{ user.ACCOUNT }}'
```
danielabdi-noaa, line 65: This is new. I suppose there are systems where you have to use a different account for HPSS?

ush/config_defaults.yaml
```diff
@@ -416,6 +495,60 @@ platform:
...
+  RAPHRR_SOIL_ROOT: ""
+  FIRE_RAVE_DIR: ""
+  AIRCRAFT_REJECT: ""
+  SFCOBS_USELIST: ""
+
+  NCL_REGION: "conus"
+  SST_update_hour: 99
+  GVF_update_hour: 99
+  SNOWICE_update_hour: 99
+  SOIL_SURGERY_time: 9999999999
```
danielabdi-noaa: These directories are defined for Hera in the machine file. Will try to set the jet/wcoss2 ones later, but other systems do not set them.

ush/config_defaults.yaml
```diff
@@ -424,11 +557,16 @@ workflow:
...
+  # WORKFLOW_ID
   # Unique ID for workflow run that will be set in setup.py
   #
+  # TAG
+  # Prepend all task name with this name if specified
+  #
   #-----------------------------------------------------------------------
   #
   WORKFLOW_ID: ""
+  TAG: ""
```
danielabdi-noaa, line 594: This is besides the WORKFLOW_ID variable we have that is used for NCO mode. The new one prepends task names to easily identify tasks within the same workflow run.

ush/config_defaults.yaml
```diff
@@ -635,6 +773,10 @@ workflow:
   FIELD_TABLE_FP: '{{ [EXPTDIR, FIELD_TABLE_FN]|path_join }}'
   NEMS_CONFIG_FP: '{{ [EXPTDIR, NEMS_CONFIG_FN]|path_join }}'
   FV3_NML_FP: '{{ [EXPTDIR, FV3_NML_FN]|path_join }}'
+  FV3_NML_CYCSFC_FP: '{{ [EXPTDIR, [FV3_NML_FN, "_cycsfc"]|join ]|path_join }}'
+  FV3_NML_RESTART_FP: '{{ [EXPTDIR, [FV3_NML_FN, "_restart"]|join ]|path_join }}'
+  FV3_NML_STOCH_FP: '{{ [EXPTDIR, [FV3_NML_FN, "_stoch"]|join ]|path_join }}'
+  FV3_NML_RESTART_STOCH_FP: '{{ [EXPTDIR, [FV3_NML_FN, "_restart_stoch"]|join ]|path_join }}'
```
danielabdi-noaa, line 808: These are the four different types of namelists generated by `RRFS_dev1` workflow.

ush/config_defaults.yaml
```diff
@@ -825,6 +1039,31 @@ workflow:
...
     - 6
 
+  STARTYEAR: "YYYY"
+  STARTMONTH: "MM"
+  STARTDAY: "DD"
+  STARTHOUR: "HH"
+  ENDYEAR: "YYYY"
+  ENDMONTH: "MM"
+  ENDDAY: "DD"
+  ENDHOUR: "HH"
```
danielabdi-noaa: I haven't seen these being used, but will check again.

ush/config_defaults.yaml
```diff
@@ -825,6 +1039,31 @@ workflow:
...
+  CYCL_HRS: []
+  CYCL_HRS_SPINSTART: []
+  CYCL_HRS_PRODSTART: []
+  CYCL_HRS_PRODSTART_ENS: []
+  CYCL_HRS_RECENTER: []
+  CYCL_HRS_STOCH: []
```
danielabdi-noaa, line 1067: Different cycle hours list for spinup/prod etc

ush/config_defaults.yaml
```diff
@@ -825,6 +1039,31 @@ workflow:
...
+  BOUNDARY_LEN_HRS: 0
+  BOUNDARY_LONG_LEN_HRS: 0
+  BOUNDARY_PROC_GROUP_NUM: 1
+  POSTPROC_LEN_HRS: 1
+  POSTPROC_LONG_LEN_HRS: 1
+  FCST_LEN_HRS_SPINUP: 1
+  FCST_LEN_HRS_CYCLES: []
```
danielabdi-noaa, line 1074: The last one can probably be merged with the way AQM handles it @chan-hoo

ush/config_defaults.yaml
```diff
@@ -974,6 +1230,35 @@ nco:
...
+  RRFSE_NWGES: '{{ OPSROOT }}/nwges'
+  RRFSE_NWGES_BASEDIR: '{{ RRFSE_NWGES }}'
+  ENSCTRL_NWGES: '{{ OPSROOT }}/nwges'
+  ENSCTRL_NWGES_BASEDIR: '{{ ENSCTRL_NWGES}}'
+
+  ENSCTRL_COMIN: '{{ COMIN_BASEDIR}}'
+  ENSCTRL_COMOUT: '{{ COMOUT_BASEDIR}}'
+  ENSCTRL_COMROOT: '{{ ENSCTRL_COMOUT }}'
+  ENSCTRL_COMIN_BASEDIR: '{{ ENSCTRL_COMIN }}'
+  ENSCTRL_COMOUT_BASEDIR: '{{ ENSCTRL_COMROOT}}/{{ envir}}'
```
danielabdi-noaa, line 1262: New NCO variables.

ush/config_defaults.yaml
```diff
@@ -1136,6 +1421,258 @@ workflow_switches:
...
+  #     Flag turn on the radar reflectivity to temperature tendenecy.
+  # DO_NLDN_LGHT
+  #     Flag turn on processing NLDN NetCDF lightning data
+  #
+  #-----------------------------------------------------------------------
+  #
+  DO_NONVAR_CLDANAL: false
+  DO_REFL2TTEN: false
+  DO_NLDN_LGHT: false
+  DO_SMOKE_DUST: false
```
danielabdi-noaa, line 1457: The DO_ variables are used both in the workflow xml file and in the scripts themselves. I have put them under `rrfs:` section.

ush/config_defaults.yaml
```diff
@@ -1136,6 +1421,258 @@ workflow_switches:
...
+  AT_START_CYCLEDEF: "00 01 01 01 2100 *"
+  INITIAL_CYCLEDEF: "00 01 01 01 2100 *"
+  BOUNDARY_CYCLEDEF: "00 01 01 01 2100 *"
+  BOUNDARY_LONG_CYCLEDEF: "00 01 01 01 2100 *"
+  SPINUP_CYCLEDEF: "00 01 01 01 2100 *"
+  PROD_CYCLEDEF: "00 01 01 01 2100 *"
+  PRODLONG_CYCLEDEF: "00 01 01 01 2100 *"
+  RECENTER_CYCLEDEF: "00 01 01 01 2100 *"
+  ARCHIVE_CYCLEDEF: "00 01 01 01 2100 *"
+  SAVEINPUT_CYCLEDEF: "00 01 01 01 2100 *"
```
danielabdi-noaa: Default cycle definitions set to an impossible date.

ush/config_defaults.yaml
```diff
@@ -2709,6 +3246,15 @@ global:
   HALO_BLEND: 10
   #
   #-----------------------------------------------------------------------
+  # 
+  # PRINT_DIFF_PGR:
+  # Option to turn on/off pressure tendency diagnostic
+  #
+  #-----------------------------------------------------------------------
+  #
+  PRINT_DIFF_PGR: false
```
danielabdi-noaa, line 2680: Set during workflow generation so I included it here.

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1266328943/20230315000522/ufs-srweather-app
Build was Successful
Rocoto jobs started
Long term tracking will be done on 10 experiments
ci-hera-intel-WE
Experiment Succeeded on hera: pregen_grid_orog_sfc_climo
Experiment Succeeded on hera: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_2017_gfdlmp_regional_plot
Experiment Succeeded on hera: community_ensemble_2mems_stoch
Experiment Succeeded on hera: MET_ensemble_verification
All experiments completed

ush/fill_jinja_template.py
```diff
@@ -186,6 +186,12 @@ def parse_args(argv):
         action="store_true",
         help="Suppress all output",
     )
+    parser.add_argument(
+        "-i",
+        "--ignore_undefined",
+        action="store_true",
+        help="Ignore undefined jinja template variables",
+    )
```
danielabdi-noaa, line 194: Current SRW jinja2 renderer has a strict requirement that all variables be defined. However, to be able to include the workflow from the beginning this needs to be relaxed. Unactivated workflow sections will completely disappear after rendering so no problem there, however undefined variables that are not used will have `None` entry. The workflow still works in this case.

ush/generate_FV3LAM_wflow.py
```diff
@@ -213,8 +213,11 @@ def generate_FV3LAM_wflow(ushdir, logfile: str = "log.generate_FV3LAM_wflow", de
         # from the jinja template file.
         #
         try:
+            # Until all of RRFS is merged lets temporarily ignore undefined template variables
+            # This is needed to have all the workflow in at the beginning, then add testable tasks one
+            # one by one. Once everything is merged this should be removed.
```
danielabdi-noaa: This can be removed right before this PR is merged so that the strict variable checking in fill_jinja is maintained. But for testing you would need this.

ush/generate_FV3LAM_wflow.py
```diff
@@ -284,8 +287,12 @@ def generate_FV3LAM_wflow(ushdir, logfile: str = "log.generate_FV3LAM_wflow", de
     #
     # Copy or symlink fix files
     #
-    if SYMLINK_FIX_FILES:
+    mkdir_vrfy("-p", FIXam)
+    if RUN_TASK_MAKE_GRID and RUN_ENVIR != "nco":
+        ln_vrfy(f"""-fsn '{HOMEdir}/fix/fix_am/global_hyblev_fcst_rrfsL65.txt'"""
```
danielabdi-noaa: We symlink the `65` level upp file here.

ush/generate_FV3LAM_wflow.py
```diff
@@ -607,12 +608,211 @@ def generate_FV3LAM_wflow(ushdir, logfile: str = "log.generate_FV3LAM_wflow", de
...
+                      Full path to output namelist file:
+                        FV3_NML_FP = '{FV3_NML_FP}'
+                      Full path to output namelist file for DA:
+                        FV3_NML_RESTART_FP = '{FV3_NML_RESTART_FP}'
+                      Namelist settings specified on command line:\n
+                        settings =\n\n"""
+                )
+                + settings_str
+            )
+    #
```
danielabdi-noaa, line 676: These are the different namelists RRFS_dev1 generates.

ush/machine/hera.yaml
```diff
@@ -35,7 +43,21 @@ platform:
...
+  LIGHTNING_ROOT: "/scratch2/BMC/zrtrr/rli/data/lightning"
+  ENKF_FCST: "/scratch2/BMC/zrtrr/rli/data/enkf/atm"
+  FFG_DIR: ""
+  SST_ROOT: "/scratch2/BMC/zrtrr/rli/data/highres_sst"
+  GVF_ROOT: "/scratch2/BMC/zrtrr/rli/data/gvf/grib2"
+  IMSSNOW_ROOT: "/scratch2/BMC/zrtrr/rli/data/snow/ims96/grib2"
+  RAPHRR_SOIL_ROOT: ""
+  FIRE_RAVE_DIR: ""
+  AIRCRAFT_REJECT: "/scratch2/BMC/zrtrr/rli/data/amdar_reject_lists"
+  SFCOBS_USELIST: "/scratch2/BMC/zrtrr/rli/data/mesonet_uselists"
```
danielabdi-noaa: I figured this from the `set_rrfs_config.sh` file in GSL fork for the test case I was running. Depending on the type of test these can be different.

ush/predef_grid_params.yaml
```diff
@@ -222,6 +222,41 @@
...
+    WRTCMP_cen_lon: -97.5
+    WRTCMP_cen_lat: 38.5
+    WRTCMP_stdlat1: 38.5
+    WRTCMP_stdlat2: 38.5
+    WRTCMP_nx: 1746
+    WRTCMP_ny: 1014
+    WRTCMP_lon_lwr_left: -122.17364391
+    WRTCMP_lat_lwr_left: 21.88588562
+    WRTCMP_dx: 3000.0
+    WRTCMP_dy: 3000.0
```
danielabdi-noaa: Someone please check if we already have this predef grid with a different name.

ush/predef_grid_params.yaml
```diff
@@ -877,35 +912,36 @@
 #-----------------------------------------------------------------------
 #
 # Emulation of GSD's RAP domain with ~13km cell size.
+# Note: Close to RRFS_dev1's GSD_RAP13km. Differences are given in comments
 #
 #-----------------------------------------------------------------------
 #
 "RRFS_NA_13km":
```
danielabdi-noaa: This had a different name in GSL fork. I kept the values in SRW and put the ones in GSL in comments.

ush/set_FV3nml_ens_stoch_seeds.py
```diff
@@ -58,7 +58,7 @@ def set_FV3nml_ens_stoch_seeds(cdate):
     #
     # -----------------------------------------------------------------------
     #
-    fv3_nml_ensmem_fp = f"{os.getcwd()}{os.sep}{FV3_NML_FN}"
+    fv3_nml_ensmem_fp = f"{os.getcwd()}{os.sep}{FV3_NML_FN}_base"
```
danielabdi-noaa, line 61: The base namelist file has a `_base` extension and all others are derived from it, for example, appending stochastic physics parameters etc.

panll:
Test passed on Hera. @danielabdi-noaa

panll (approved):

danielabdi-noaa:
@MatthewPyle-NOAA @chan-hoo Could you take a look at the PR specifically the NCO aspect of it? Pinning others to inform  that this is the high-priority PR I was talking about @christinaholtNOAA @mkavulich @BenjaminBlake-NOAA @EdwardSnyder-NOAA

fix/fix_am/global_hyblev_fcst_rrfsL65.txt
```diff
@@ -0,0 +1,68 @@
+     2    66
```
EdwardSnyder-NOAA: The `fix` subdir of the regional_workflow was removed awhile back, and it is now stagged locally on each platform under `/path/to/UFS_SRW_App/develop/fix`. I suggest we move this file under the `fix_am` directory there. I'm moving some files from the dev1 branch to the `fix_gsi` directory, so I can help stage this file too.

ush/config_defaults.yaml
MatthewPyle-NOAA: `$NWGES/$NET/$envir/$RUN.$yyyymmdd/$hh` 
This item in the comments is an outdated form.  Now an NWGES space is under COM, which I think is how it is actually being defined ({{OPSROOT}}/nwges)

fix/fix_am/global_hyblev_fcst_rrfsL65.txt
danielabdi-noaa: @EdwardSnyder-NOAA  I agree. It makes life easier if such custom fix files are available on each system, so I've removed this file and will rely on fix_am instead.

ush/config_defaults.yaml
danielabdi-noaa: @MatthewPyle-NOAA Fixed. Hopefully this directory can be removed in the future.

fix/fix_am/global_hyblev_fcst_rrfsL65.txt
EdwardSnyder-NOAA: Added `global_hyblev_fcst_rrfsL65.txt` file to `fix_am` directory on RDHPCs. Still need to add it to odin, stampede, and wcoss2.

ush/machine/hera.yaml
EdwardSnyder-NOAA: I was curious how you were going to handle switching between a retro and non-retro run. As you pointed out, `set_rrfs_config.sh` can switch these paths. My thoughts where to either add jinja logic to switch paths or add a new yaml section like "platform" or "data" called "retro_run", which will reconfigure paths whenever a retro task will run.

danielabdi-noaa: @EdwardSnyder-NOAA I haven't thought that far ahead , just picked what I needed for the CONUS test case!
Providing multiple paths in the machine yaml files for retro/realtime ensemble/non-ensemble and other conditions, similar to how the `data:` section picks a path based on the format, could probably work. Now that jinja is supported in the machine file too, that is also a possibility, but I prefer the other option. I will take a look at the latest `set_rrfs_config/genera_config` files to see how best to handle it.

danielabdi-noaa: I have implemented the yaml solution by extending the `data:` section to contain both realtime and retrospective run paths. Each RRFS data path is still under `platform:` and will be set in `setup.py` during workflow generation. One can still override the values by setting them directly. Let me know what you think.

EdwardSnyder-NOAA: @danielabdi-noaa - that looks good to me!

ush/config_defaults.yaml
MichaelLueken: It looks like the NOAA-GSL `ush/config_defaults.yaml` file has both a `SERVICE_ACCOUNT` and `HPSS_ACCOUNT`, where `HPSS_ACCOUNT` defaults to `SERVICE_ACCOUNT` if not set.  @hu5970, are these entries required?  I'm seeing that they appeared to be added for Jet.  Any additional information would be greatly appreciated.  Thanks!

ush/predef_grid_params.yaml
MichaelLueken: It looks like the `RRFS_CONUS_3km_HRRRIC` grid is identical to the `RRFS_CONUScompact_3km` grid, with the exception of `DT_ATMOS` being set to 36.  This modification was merged at [fd3c778](https://github.com/ufs-community/ufs-srweather-app/commit/fd3c778fa444906970a83664b8bdcdd091ccff68) and was to use less wasteful values of `DT_ATMOS`.  It would probably be best to have the scripts use the `RRFS_CONUScompact_3km` grid, rather than add the new `RRFS_CONUS_3km_HRRRIC` grid with a different `DT_ATMOS` value.

MichaelLueken (commented):
@danielabdi-noaa Overall, these changes look good to me.  I agree that I'm not entirely sure what the purpose of the `SERVICE_ACCOUNT` is in the `ush/config_defaults.yaml` file.  I have reached out to @hu5970 to see if he can shine some light on it (or if it even needs to be included).  Also, the newly added `RRFS_CONUS_3km_HRRRIC` grid appears to be identical to the `RRFS_CONUScompact_3km` grid.  The only difference is that the `DT_ATMOS` value was decreased recently to use less wasteful values.  So, I would suggest using `RRFS_CONUScompact_3km` rather than `RRFS_CONUS_3km_HRRRIC`.

ush/config_defaults.yaml
danielabdi-noaa: Thanks, I missed `HPSS_ACCOUNT` which was recently added after this PR is opened. They both default to the regular `ACCOUNT`, but there maybe a usecase for them on some systems as you pointed out.

danielabdi-noaa:
@MichaelLueken Thanks for capturing the predefined grid duplicates. It looks like there are no new predef grids to add, so I have removed all changes to predef_grid file.

ush/config_defaults.yaml
MichaelLueken: From Ming -

The `SERVICE_ACCOUNT` and `HPSS_ACCOUNT` are related to the machine environment and the applications on Jet.  When we run a real-time system, we have a reservation under a certain account, but we do not have enough resources to run all the jobs under reserved resources. So, we usually run key components (GSI, model,...) under the reservation and run post/plotting/archive tasks under another account, which is outside the reservation.

When I run RRFS, I run key components under the account that have more resources (or high possibility to run) but other post jobs under account that may run slow.

Sometimes HPSS can only run under a certain account and we added `HPSS_ACCOUNT` recently.

We do not need this under operation environment but need this on Jet to control resource usage.

So, it certainly sounds like we should keep both `SERVICE_ACCOUNT` and `HPSS_ACCOUNT` in develop, but these can be removed in the release/production branch.

danielabdi-noaa: Thanks @MichaelLueken @hu5970! We will keep both of them then.

MichaelLueken:
@danielabdi-noaa Going through the latest changes in the `NOAA-GSL/regional_workflow/RRFS_dev1` branch associated with `ush/config_defaults.sh`, I'm noting the following missing or changed options:

- `OUTPUT_FH` was introduced in https://github.com/NOAA-GSL/regional_workflow/pull/502
- `DO_ENS_RADDA` was introduced in https://github.com/NOAA-GSL/regional_workflow/pull/496
- `DO_ENVAR_RADAR_REF_ONCE` was introduced in https://github.com/NOAA-GSL/regional_workflow/pull/451
- `SAVEINPUT_CYCLEDEF` was changed to `SAVEDA_CYCLEDEF` in https://github.com/NOAA-GSL/regional_workflow/pull/480
- `DO_SAVE_INPUT` was changed to `DO_SAVE_DA_OUTPUT` in https://github.com/NOAA-GSL/regional_workflow/pull/480
- The defaults for `DA_CYCLE_INTERV`, `RESTART_INTERVAL`, and `RESTART_INTERVAL_LONG` were changed to `1`, `1,2`, and `1,2`, respectively, in https://github.com/NOAA-GSL/regional_workflow/pull/477
- `FIRE_RRFS_ROOT`, `OBSPATH_PM`, and `DO_SDDACYCLE` were introduced in https://github.com/NOAA-GSL/regional_workflow/pull/448
- `RTMA_OBS_FEED` was introduced in https://github.com/NOAA-GSL/regional_workflow/pull/473
- `PREP_FVCOM` was introduced in https://github.com/NOAA-GSL/regional_workflow/pull/472

Until this PR has been merged, I will keep an eye out on further changes to the `ush/config_defaults.sh` file.

danielabdi-noaa:
@MichaelLueken Thanks a lot for going through the commits and summarizing the changes! I have updated this PR with most of the new variable changes, except for two, `OUTPUT_FH` goes under `task_run_fcst` and `PREP_FVCOM` goes under `task_make_ics` so those are put in future PRs for the respective tasks.

ush/config.rrfs.yaml
christinaholtNOAA, line 1: This file should not be added yet. We don't have the capabilities in place to run or test any of it, and many of the settings here will not ever be supported in SRW.

ush/config_defaults.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   description: >-
     Default configuration for an experiment. The valid values for most of the
     parameters are specified in valid_param_vals.yaml
-  version: !!str '1.0'
+  version: !!str '2.0'
```
christinaholtNOAA, line 8: It's not clear to me how we will want to bump versions, but it definitely doesn't seem like the time to do that now.

ush/config_defaults.yaml
```diff
@@ -159,10 +169,52 @@ platform:
...
+  # The queue or QOS to which the task that upp is submitted.  
+  # If this is not set or set to an empty string, it will be (re)set to a 
+  # machine-dependent value.
+  #
+  # RESERVATION:
+  # The reservation for major tasks.  
+  #
+  # RESERVATION_POST:
+  # The reservation for post tasks.  
+  #
```
christinaholtNOAA, line 212: Partitions and queues are now all set in the parm/wflow files. None of these are needed here. Reservations, when they are added, should also live in the parm/wflow files.

ush/config_defaults.yaml
```diff
@@ -179,6 +231,16 @@ platform:
...
+  PARTITION_GRAPHICS: ""
+  QUEUE_GRAPHICS: ""
+  PARTITION_ANALYSIS: ""
+  QUEUE_ANALYSIS: ""
+  PARTITION_PRDGEN: ""
+  QUEUE_PRDGEN: ""
+  PARTITION_POST: ""
+  QUEUE_POST: ""
+  RESERVATION: ""
+  RESERVATION_POST: ""
```
christinaholtNOAA, line 243: None of these will be needed here as they can be set explicitly in the parm/wflow files for the tasks when they are added to the workflow.

ush/config_defaults.yaml
```diff
@@ -435,6 +520,64 @@ platform:
...
+  # Setup default locations for FIRE_RRFS files and update time
+  #  FIRE_RAVE_DIR
+  #  FIRE_RRFS_ROOT
+  #  FIRE_RRFS_update_hour
+  #
+  #-----------------------------------------------------------------------
+  #
+  ARCHIVEDIR: ""
+  NCARG_ROOT: ""
+  NCL_HOME: ""
```
christinaholtNOAA: ARCHIVEDIR, NCARG_ROOT, and NCL_HOME should not be needed in SRW. These are variables for tasks that will likely be add-ons at GSL, and not supported in SRW.

ush/config_defaults.yaml
```diff
@@ -435,6 +520,64 @@ platform:
...
+  #   SNOWICE_update_hour: cycle time for updating snow/ice 
+  #
+  # Setup default resource data locations for soil surgery and time:
+  #   RAPHRR_SOIL_ROOT: locations of RAP/HRRR forecast netcdf files
+  #   SOIL_SURGERY_time: cycle time for soil surgery 
+  #
+  # Setup default locations for FIRE_RRFS files and update time
+  #  FIRE_RAVE_DIR
+  #  FIRE_RRFS_ROOT
+  #  FIRE_RRFS_update_hour
```
christinaholtNOAA, line 554: You shouldn't need these items included in config_defaults since they are captured in the data section of the machine files. I don't think the data section is checked against the config_default entries.

ush/config_defaults.yaml
```diff
@@ -849,16 +1001,88 @@ workflow:
...
   #
   DATE_FIRST_CYCL: "YYYYMMDDHH"
   DATE_LAST_CYCL: "YYYYMMDDHH"
   INCR_CYCL_FREQ: 24
   FCST_LEN_HRS: 24
   FCST_LEN_CYCL:
+    - 6
+    - 12
+    - 12
+    - 6
```
christinaholtNOAA: I removed these 4 lines purposefully in the YAML/XML PR since it's best to keep these values consistent with the user choices above.

ush/config_defaults.yaml
christinaholtNOAA: Let's hold off on how we set the cycle hrs for all of these until it can be tested appropriately. Having these placeholders seems like it's too early.

ush/config_defaults.yaml
```diff
@@ -849,16 +1001,88 @@ workflow:
...
-
+  CYCL_HRS: []
+  CYCL_HRS_SPINSTART: []
+  CYCL_HRS_PRODSTART: []
+  CYCL_HRS_PRODSTART_ENS: []
+  CYCL_HRS_RECENTER: []
+  CYCL_HRS_STOCH: []
+  BOUNDARY_LEN_HRS: 0
+  BOUNDARY_LONG_LEN_HRS: 0
+  BOUNDARY_PROC_GROUP_NUM: 1
```
christinaholtNOAA, line 1070: These should come with the mods to get extrnl tasks so that we can test their functionality.

ush/config_defaults.yaml
```diff
@@ -1014,6 +1253,35 @@ nco:
...
 
+  #
+  # New additions from RRFS_dev1
+  #
+  NWGES: '{{ OPSROOT }}/nwges'
+  NWGES_BASEDIR: '{{ NWGES }}'
+  RRFSE_NWGES: '{{ OPSROOT }}/nwges'
+  RRFSE_NWGES_BASEDIR: '{{ RRFSE_NWGES }}'
+  ENSCTRL_NWGES: '{{ OPSROOT }}/nwges'
+  ENSCTRL_NWGES_BASEDIR: '{{ ENSCTRL_NWGES}}'
```
christinaholtNOAA, line 1256: Why is there a need for these pairs? For example why both NWGES and  NWGES_BASEDIR? One just defaults to another.

ush/config_defaults.yaml
```diff
@@ -1031,6 +1299,265 @@ nco:
...
+  # Flag turn on the spin-up cycle.
+  #
+  # DO_POST_SPINUP:
+  # Flag turn on the UPP for spin-up cycle.
+  #
+  # DO_POST_PROD:
+  # Flag turn on the UPP for prod cycle.
+  #
+  #-----------------------------------------------------------------------
+  #
```
christinaholtNOAA, line 1318: These types of high level switches, at least in this quantity, is not likely how the workflow will be configured moving forward. We will want to define what tasks run in a workflow given the definition of those tasks.

ush/config_defaults.yaml
```diff
@@ -1031,6 +1299,265 @@ nco:
...
+#----------------------------------
+start_time:
+  START_TIME_SPINUP: 01:10:00
+  START_TIME_PROD: 02:20:00
+  START_TIME_CONVENTIONAL_SPINUP: 00:40:00
+  START_TIME_LATE_ANALYSIS: 01:40:00
+  START_TIME_CONVENTIONAL: 00:40:00
+  START_TIME_NSSLMOSIAC: 00:45:00
+  START_TIME_LIGHTNINGNC: 00:45:00
+  START_TIME_PROCSMOKE: 00:45:00
```
christinaholtNOAA, line 1479: Cycledefs will be defined in wflow/parm files, or similar. Not in config_defaults.

ush/config_defaults.yaml
```diff
@@ -1466,6 +1993,26 @@ task_get_extrn_lbcs:
...
+  #  When search boundary conditions from previous cycles in prep_start step,
+  #  the search will start at cycle before (this parameter) of current cycle.
+  #  For example: 0 means search start at the same cycle lbcs directory.
+  #               1 means search start at 1-h previous cycle  lbcs directory.
+  #               2 means search start at 2-h previous cycle  lbcs directory.
+  #
+  #-----------------------------------------------------------------------
+  #
+  LBCS_SEARCH_HRS: 6
+  EXTRN_MDL_LBCS_SEARCH_OFFSET_HRS: 0
```
christinaholtNOAA, line 2002: Should these not be added when the functionality is added to search for these LBCS?

ush/fill_jinja_template.py
christinaholtNOAA: Is this still needed given our current situation? It seems like a dangerous situation to allow the jinja template to come out incomplete.

ush/generate_FV3LAM_wflow.py
```diff
@@ -360,6 +359,8 @@ def generate_FV3LAM_wflow(ushdir, logfile: str = "log.generate_FV3LAM_wflow", de
         "target_lon": LON_CTR,
         "target_lat": LAT_CTR,
         "nrows_blend": HALO_BLEND,
+        "regional_bcs_from_gsi": False,
+        "write_restart_with_bcs": False,
```
christinaholtNOAA, line 363: Will this impact AQM?

ush/machine/hera.yaml
```diff
@@ -37,8 +45,36 @@ platform:
...
+      LIGHTNING_ROOT: "/scratch2/BMC/public/data/lightning"
+      ENKF_FCST: "/scratch1/NCEPDEV/rstprod/com/gfs/prod"
+      FFG_DIR: ""
+      SST_ROOT: "/scratch2/BMC/zrtrr/rli/data/highres_sst"
+      GVF_ROOT: "/scratch2/BMC/zrtrr/rli/data/gvf/grib2"
+      IMSSNOW_ROOT: "/scratch2/BMC/zrtrr/rli/data/snow/ims96/grib2"
+      RAPHRR_SOIL_ROOT: ""
+      FIRE_RAVE_DIR: ""
+      AIRCRAFT_REJECT: "/scratch2/BMC/zrtrr/rli/data/amdar_reject_lists"
+      SFCOBS_USELIST: "/scratch2/BMC/zrtrr/rli/data/mesonet_uselists"
```
christinaholtNOAA, line 77: I don't think this is the appropriate structure for retrieve_data.py to gather these. I believe there are other efforts to get the data paths incorporated into the workflow.

ush/set_FV3nml_ens_stoch_seeds.py
christinaholtNOAA: If the name of the file changes, it should change in config_defaults so that it's clear what it is from the beginning. One of the features of having configuration settings is that there is rarely a need to change these types of things in the scripts.

ush/set_FV3nml_ens_stoch_seeds.py
```diff
@@ -174,7 +174,7 @@ def setUp(self):
         mkdir_vrfy("-p", EXPTDIR)
         cp_vrfy(
             os.path.join(PARMdir, "input.nml.FV3"),
-            os.path.join(EXPTDIR, "input.nml"),
+            os.path.join(EXPTDIR, "input.nml_base"),
```
christinaholtNOAA, line 177: Here, it seems like we should be referencing the configuration setting instead of repeating this information, especially if we'd like to change it.

christinaholtNOAA (changes requested):
@danielabdi-noaa I appreciate your contribution here. 

I am very concerned about this approach to adding a bunch of variables that aren't used yet. We have no way to test them, and no way to know whether they will be relevant or needed when the tasks that currently use them in RRFS_dev1 are contributed. 

I am also concerned that it will be difficult to communicate what does and does not work to the rest of the community. What I mean is, if someone sees a flag that says "do_something" they expect "something" to work. Given that is not true, I do not think it's fair for us to include those sorts of flags until the feature is enabled.

ush/config_defaults.yaml
danielabdi-noaa: But the `parm/wflow/default_worklfow.yaml` files get the there settings from config files, infact, we aready have partitions/queues in them already, so I am just extending that for jobs RRFS adds
```
    PARTITION_DEFAULT: '{{ platform.get("PARTITION_DEFAULT") }}'
    PARTITION_FCST: '{{ platform.get("PARTITION_FCST") }}'
    PARTITION_HPSS: '{{ platform.get("PARTITION_HPSS") }}'
    QUEUE_DEFAULT: '{{ platform.get("QUEUE_DEFAULT") }}'
    QUEUE_FCST: '{{ platform.get("QUEUE_FCST") }}'
    QUEUE_HPSS: '{{ platform.get("QUEUE_HPSS") }}'
   ```

ush/config_defaults.yaml
danielabdi-noaa: See above! We are not setting them directly, but getting them from `platform` section of the config file.

ush/config_defaults.yaml
danielabdi-noaa: No, they are needed. The `data:` section is set for `retro:` and `real:` runs but one is picked and used to set these variables that are under `platform:` section. This is exactly the same as how we set `EXTRN_MDL_SYSBASEDIR_ICS/LBCS`, where we use values under `data: ics_lbcs:` section to set it and discard `data:` section.

ush/config_defaults.yaml
danielabdi-noaa: Ok I must have missed those were removed due to the frequent merges I need to make for this and other PRs.

ush/config_defaults.yaml
danielabdi-noaa: Why wouldn't we allow the user to change cycledefs and other things from the user config files? Editing workflow yaml files is like editing source code to get a test case running, unless I missed what you are saying here. The user should interact with config files only iMO.

ush/config_defaults.yaml
danielabdi-noaa: This can definitely wait until then, but I don't agree with your approach that everything should wait until the task is added. Incorporating variables in a central PR helps with coordination, is less error prone etc.

ush/config_defaults.yaml
danielabdi-noaa: Why is this not the right time? Infact, everytime a variable is added is time for a minor version update.

ush/machine/hera.yaml
danielabdi-noaa: Aren't these "fixed files" that do not change from test case to test case? `retrieve_data.py` has so far been used to get ICS/LBCS that change from one test case to test case. @EdwardSnyder-NOAA has put some of these files under epic role account and these paths can be changed as new files are added to locations that SRW can use.

ush/config.rrfs.yaml
danielabdi-noaa: Actually you can run it even now. It will generate a workflow file that doesn't have any of the rrfs tasks.

ush/fill_jinja_template.py
danielabdi-noaa: I haven't checked but note that it is not used from `generate_workflow.py` so I left the option just in case I need to use it later, not because it is currently used.

ush/generate_FV3LAM_wflow.py
danielabdi-noaa: No idea @chan-hoo probably knows.

ush/config_defaults.yaml
danielabdi-noaa: These are already tested to be working and they are used for example in `process_radarref`. Look at the description in the j-job for the task.

ush/config_defaults.yaml
danielabdi-noaa: The PR for ic/lbcs modifications #666 is now open and depends on this PR to be tested. I prefer to define all variables that will be needed for upcoming tasks in this central PR.

ush/config_defaults.yaml
danielabdi-noaa: That is how they are defined in RRFS_dev
https://github.com/NOAA-GSL/regional_workflow/blob/feature/RRFS_dev1/ush/setup.sh#L1394-L1404
If we find that they are really duplicates later, we can ofcourse remove them but for now I trust the original developers more than my own judgement.

ush/set_FV3nml_ens_stoch_seeds.py
danielabdi-noaa: I am not sure what your concern is here, but the way RRFS handles input namelist files is different from how SRW does it. RRFS generates four different input namelist  files and symlinks it to `input.nml` depending on the configuration, as opposed to in-place modification of one namelist file that SRW uses. This also applies for the stochastic physics options modification and I tested that this modification works for the WE2E test `community_ensemble_2mems_stoch`.

ush/config_defaults.yaml
danielabdi-noaa: Removed.

ush/config_defaults.yaml
danielabdi-noaa: Fixed.

ush/machine/hera.yaml
EdwardSnyder-NOAA: These paths are a combination of fixed and observation files. We were trying to replicate the functionality from the [set_rrfs_config.sh](https://github.com/NOAA-GSL/regional_workflow/blob/feature/RRFS_dev1/ush/set_rrfs_config.sh) found in the dev1 branch. Essentially this script sets paths based on machine and model run (retro/current) type. It made sense to add the logic here even though the `retrieve_data.py` script isn't used to grab these files. We are definitely open for suggestions on placing this logic elsewhere. 

In addition, how often should we utilize the `retrieve_data.py` script? As the process lightning task grabs lightning data in the jobs script without the `retrieve_data.py` script. Would we need to integrate all the rrfs obs data into the `retrieve_data.py` script?

ush/config_defaults.yaml
christinaholtNOAA: It is generally unclear to me whether this should be the SRW version, the release version, etc.

ush/config_defaults.yaml
christinaholtNOAA: I'm suggesting that we don't necessarily have to set them explicitly here in config defaults. Here, the default queue is the main queue on a platform. FCST is special because it might need bigger nodes with more memory, and HPSS has internet access -- essentially it's the front-end node on platforms where that makes sense. Because each of these three function in separate ways for the workflow, we can choose from those 3 which task runs under which queue in the task definition.

There is no need for a one-to-one correspondence between available config defaults variables and what values are set as entities.


So, in the definition of the post jobs, we can modify the entry to specifically be `debug` if we want to change it for a given user/platform.

ush/config_defaults.yaml
christinaholtNOAA: As I argue above, we do not need them as they generally don't serve any additional function on our platforms than the current functions allow in the XML generation.

ush/config_defaults.yaml
christinaholtNOAA: The `data:` section of the machine files in SRW currently is meant to work with retrieve_data.py script to set where to get the datas on disk and is a structured section with the structure strictly defined by that tool.  The section should remain consistent with the tool's existing behavior.

ush/config_defaults.yaml
christinaholtNOAA: This is why I'm suggesting that the necessary variables be added with the tool that requires them. They should be added when they are needed, not separately where they cannot be tested on their own, or recognized as to why they are useful.

They are included in PR #638 , and I suggest that be the only place they are added.

ush/config_defaults.yaml
christinaholtNOAA: I prefer, as stated several times now, that they be included where they can be tested. This is a prime example of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) (You aren't gonna need it). 

"Always implement things when you actually need them, never when you just foresee that you [will] need them"

I understand that the foresight is right around the corner, but so many things could change in the review process for the other PR that you may have caused yourself additional work by having to change this thing to work with mods to the later PR. Handle it there and it's totally internally consistent right away.

ush/config_defaults.yaml
christinaholtNOAA: As you suggest, this one should not be included. A proposed change for the current solution for AQM is being discussed in Issue #709

ush/config_defaults.yaml
christinaholtNOAA: Putting in a clean implementation now is preferred. A main goal of the RRFS_dev1 merge is to take care of the refactoring needed to address some of the accidental complexities that have snuck in over time.

ush/config_defaults.yaml
christinaholtNOAA: The rocoto section of the config files are 100% configurable. I am definitely not suggesting that the users edit any of the parm/wflow files. The sections defined in those parm/wflow files are completely configurable by the user in their own user config. There is no need to add the sections to the config_defaults.yaml file.

ush/config_defaults.yaml
christinaholtNOAA: Please see my YAGNI argument above. It is a standard, industry wide best practice, and does apply in this case.

ush/machine/hera.yaml
christinaholtNOAA: I believe we should be using the retrieve_data script any time we need to pull data from disk and/or external resources. 

One assumption made in RRFS_dev was that we'd always have data staged on disk for real-time or retro runs. That is _not_ an appropriate assumption for the community who likely have different goals. SRW users working outside the real-time paradigm, or on platforms where real-time data streams aren't available, or on retrospective case studies, will need a way to pull data from archives like HPSS or AWS.

The retrieve_data tool is meant to allow a user to define whether they have staged data on disk via their own doing, or a supported data stream, or whether the workflow should go get it for them from known external locations.

christinaholtNOAA (changes requested):

PR #656, opened 2023-03-08
[develop] Configs for compiling and running online-cmaq on cheyenne.

padhrigmccarthy:
## DESCRIPTION OF CHANGES: 
Configurations for Cheyenne HPC, to facilitate building and running online-cmaq on cheyenne.

### Type of change

- [ x] New feature (non-breaking change which adds functionality)

## TESTS CONDUCTED: 
Tested on cheyenne with intel.

## DEPENDENCIES:
Submitting additions to UFS_UTILS and AQM_utils as well.

### DIFF
--- a/modulefiles/tasks/cheyenne/get_obs.local.lua
+++ b/modulefiles/tasks/cheyenne/get_obs.local.lua
@@ -1,1 +1,1 @@
-load("miniconda_regional_workflow-cmaq")
+load("miniconda_regional_workflow")
--- a/modulefiles/tasks/cheyenne/miniconda_regional_workflow.lua
+++ b/modulefiles/tasks/cheyenne/miniconda_regional_workflow.lua
@@ -1,5 +1,4 @@
 unload("python")
-load("conda")
 prepend_path("MODULEPATH","/glade/work/epicufsrt/contrib/miniconda3/modulefiles")
 load(pathJoin("miniconda3", os.getenv("miniconda3_ver") or "4.12.0"))
 
--- a/ush/machine/cheyenne.yaml
+++ b/ush/machine/cheyenne.yaml
@@ -31,4 +31,4 @@ platform:
   FIXshp: /glade/p/ral/jntp/UFS_SRW_App/develop/NaturalEarth
 data:
   ics_lbcs:
-    FV3GFS: /glade/collections/rda/data/ds084.1/${yyyy}/${yyyymmdd}
+    FV3GFS: /glade/p/ral/jntp/UFS_CAM/COMGFS/gfs.${yyyymmdd}/${hh}
--- a/ush/machine/cheyenne.yaml
+++ b/ush/machine/cheyenne.yaml
@@ -16,6 +16,8 @@ platform:
   RUN_CMD_POST: mpirun -np $nprocs
   RUN_CMD_SERIAL: time
   RUN_CMD_UTILS: mpirun -np $nprocs
+  RUN_CMD_AQM: mpirun -np $nprocs
+  RUN_CMD_AQMLBC: mpirun -np ${NUMTS}
   PRE_TASK_CMDS: '{ ulimit -s unlimited; ulimit -a; }'
   TEST_EXTRN_MDL_SOURCE_BASEDIR: /glade/p/ral/jntp/UFS_SRW_App/develop/input_model_data
   TEST_PREGEN_BASEDIR: /glade/p/ral/jntp/UFS_SRW_App/develop/FV3LAM_pregen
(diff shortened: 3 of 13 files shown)


modulefiles/tasks/cheyenne/get_obs.local.lua
```diff
@@ -1 +1 @@
-load("miniconda_regional_workflow")
+load("miniconda_regional_workflow-cmaq")
```
chan-hoo: @padhrigmccarthy, can you remove this change? Online-CMAQ doesn't run the 'get_obs' task.

ush/machine/cheyenne.yaml
```diff
@@ -29,4 +31,4 @@ platform:
   FIXshp: /glade/p/ral/jntp/UFS_SRW_App/develop/NaturalEarth
 data:
   ics_lbcs:
-    FV3GFS: /glade/p/ral/jntp/UFS_CAM/COMGFS/gfs.${yyyymmdd}/${hh}
+    FV3GFS: /glade/collections/rda/data/ds084.1/${yyyy}/${yyyymmdd}
```
chan-hoo: @padhrigmccarthy, is this path for all users or your team? I wonder if this new path  works for the WE2E tests. Since I don't have access to Cheyenne, I'd like to hear from other code managers.

modulefiles/tasks/cheyenne/miniconda_regional_workflow.lua
```diff
@@ -1,4 +1,5 @@
 unload("python")
+load("conda")
```
chan-hoo: @padhrigmccarthy, is this necessary? I am asking this because this change will affect the FV3-LAM runs. I'd like to hear from other code managers before I approve this change.

MichaelLueken: I don't think this should be added.  The necessary `regional_workflow` conda environment is loaded below this point.  No python or conda should be loaded until `miniconda3/4.12.0` has been loaded.

@padhrigmccarthy Is there a reason that `load("conda")` was added before the load of `miniconda3/4.12.0`?

ush/machine/cheyenne.yaml
MichaelLueken: While the WE2E tests for both Cheyenne Intel and Cheyenne GNU run without issue with this change, I think that we should point to the directories maintained by `ufssrw`, until such time that they are transitioned to EPIC.  Further input would be appreciated.

MichaelLueken (commented):
@padhrigmccarthy 

For the change to `modulefiles/tasks/cheyenne/miniconda_regional_workflow.lua`, is there a reason that you needed to load conda?  Loading `miniconda3/4.12.0` and then activating with `conda activate regional_workflow` should be fine.

For `ush/machine/cheyenne.yaml`, is there a reason that you moved the data location away from the `ufssrw` location?  While the WE2E tests successfully ran and passed for both Intel and GNU, I believe that the data location should continue to point to the current location, until the time that the maintaining of data is transitioned to EPIC.

padhrigmccarthy:
Hi Mike-

These comments agree with Chan-Hoo's suggestions. It took a LOT of
trial-and-error to get these conda environments to load as expected, so the
'load("conda")' calls are remnants of those attempts. Thanks for the
feedback.

I will simplify and test, then push these suggested changes.

-Paddy.

ush/machine/cheyenne.yaml
padhrigmccarthy: Happy to use the default location.

padhrigmccarthy:
> @padhrigmccarthy
> For `ush/machine/cheyenne.yaml`, is there a reason that you moved the data location away from the `ufssrw` location? While the WE2E tests successfully ran and passed for both Intel and GNU, I believe that the data location should continue to point to the current location, until the time that the maintaining of data is transitioned to EPIC.

@MichaelLueken It seems we need time-dependent data for FV3GFS, rather than a fixed time. Here is what I heard back from Rajesh Kumar, the scientist leading our project at NCAR:

> Hi Paddy,
> The default location of the GFS forecasts should be /glade/collections/rda/data/ds084.1. This is automatically populated with the new GFS data. The directory /glade/p/ral/jntp/UFS_CAM/COMGFS looks like a test set up.
> Rajesh

I've pushed the change to revert this to the `ufssrw` location and made the use of time-dependent data part of our local configuration instructions.

modulefiles/tasks/cheyenne/get_obs.local.lua
chan-hoo: This was not resolved yet. Do you have any reason for this change?

padhrigmccarthy: Oops. Failed to push this file. Resolved now.

chan-hoo (approved):

MichaelLueken (approved):
@padhrigmccarthy Thank you very much for making the changes!  Just a brief note, the `ush/machine/${machine}.yaml` files point to the directories that contains the staged data for the WE2E tests on the various machines, which is why it should ultimately point to the `ufssrw` location (where the WE2E test data is contained).  The Jenkins tests passed without issue, so I am now approving this work.

PR #613, opened 2023-02-11
[develop] Integrate 'online-cmaq' into 'develop' branch: Part 3, workflow set-up and run scripts

chan-hoo:
## DESCRIPTION OF CHANGES: 
Currently, the Online-CMAQ for air quality modeling is separately managed in the 'online-cmaq' branch of the authoritative UFS SRW App repository. The main target of the current 'online-cmaq' branch is the implementation of AQM ver.7 and delivery to NCO in April. Therefore, it has been developed and tested on WCOSS2 and Hera. The Online-CMAQ checks out the same ufs weather model as the 'develop' branch. However, since it couples another component 'AQM', the compile process is different from that of the 'develop' branch.

This 'online-cmaq' branch has beeen integrating into the 'develop' branch with PR #536, PR #549, and this PR. For the convenience of the reviewers, we split the 'online-cmaq' branch into three parts as follows:

Part 1: j-job and ex- scripts (PR #536)
Part 2: build scripts and module files (PR #549)
Part 3: workflow set-up, generation, run scripts (this PR)

- To build and run Online-CMAQ (https://github.com/ufs-community/ufs-srweather-app/wiki/Air-Quality-Modeling):
```
git clone -b online-cmaq https://github.com/ufs-community/ufs-srweather-app
cd ufs-srweather-app
./manage_externals/checkout_externals
./devbuild.sh -p=hera (or wcoss2) -a=ATMAQ
cd ush
cp config.aqm.community.yaml config.yaml
vim config.yaml (update MACHINE and ACCOUNT)
(load the python environment for the workflow)
python3 generate_FV3LAM_wflow.py
```
- Note that the last four tasks (post_stat_o3, post_stat_pm25, bias_correction_o3, and bias_correction_pm25) only work on WCOSS2.

### Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## TESTS CONDUCTED: 
- WE2E tests:
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_RAP_suite_HRRR
MET_verification
community_ensemble_2mems_stoch

- [x] hera.intel
- [ ] orion.intel
- [ ] cheyenne.intel
- [ ] cheyenne.gnu
[...]

PR #666, opened 2023-03-10
[develop] Modify get_lbcs and make_lbcs.

### LINKED ISSUE #665: Download 0th hour LBCS and parallelize make_lbcs for RRFS
## Description

 
There is some minor difference between SRW app and RRFS in the way LBCS are downloaded. 
  - SRW currently does not download the 0th hour lbcs while RRFS does.
  - RRFS does a prefetch of lbcs longer than the forecast hour. This is controlled by a new variable `BOUNDARY_LEN_HRS`
  - `make_lbcs` is parallelized in RRFS and saves a bit of time
## Solution

Make the necessary changes described above. 

## Requirements**

None

Make sure 
- existing SRW workflow including AQM works properly.
- RRFS get_lbcs and make_lbcs feature works properly

## Dependencies (optional)

None

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 

Addresses issue #665. Get_lbcs need to download 0th hour LBCS and make_lbcs need to be parallelized.
Note that there is no workflow so this PR can not be run independently until then.
However, it did successfully run for both spinup/production cycles in PR #540 

### Type of change

## TESTS CONDUCTED: 

- [X] hera.intel
Successfully run these tasks modified for RRFS in PR #540
- [ ] orion.intel
- [ ] cheyenne.intel
- [ ] cheyenne.gnu
- [ ] gaea.intel
- [ ] jet.intel
- [ ] wcoss2.intel
- [ ] Jenkins
- [ ] fundamental test suite

## DEPENDENCIES:

- #647 

## ISSUE: 

- #665 

## CHECKLIST

## LABELS (optional): 

- [ ] Work In Progress
- [ ] bug
- [ ] enhancement
- [ ] documentation
- [ ] release
- [ ] high priority
- [ ] run_ci
- [ ] Needs Cheyenne test 
- [ ] Needs Jet test 
- [ ] Needs Hera test 
- [ ] Needs Orion test 
- [ ] help wanted


panll:
Tested on Hera. Running time error: 
.../ufs-srweather-app/jobs/JREGIONAL_MAKE_ICS: line 63: NWGES_DIR: unbound variable
End JREGIONAL_MAKE_ICS at Tue Apr 11 17:42:24 UTC 2023 with error code 1 (time elapsed: 00:00:00)

danielabdi-noaa:
@panll I have added the necessary NWGES variables but this PR requires a workflow update for the specific way RRFS downloads LBCS, so it can not be tested without that.

jobs/JREGIONAL_MAKE_ICS
```diff
@@ -57,10 +57,14 @@ for the FV3 (in NetCDF format).
 #
 if [ $RUN_ENVIR = "nco" ]; then
     export INPUT_DATA="${COMIN}"
+    export INPUT_DATA_NWGES="${NWGES_DIR}${SLASH_ENSMEM_SUBDIR}/ics"
 else
     export INPUT_DATA="${COMIN}${SLASH_ENSMEM_SUBDIR}/INPUT"
+    export INPUT_DATA_NWGES="${NWGES_DIR}${SLASH_ENSMEM_SUBDIR}/ics"
 fi
 mkdir_vrfy -p "${INPUT_DATA}"
+mkdir_vrfy -p "${INPUT_DATA_NWGES}"
```
danielabdi-noaa, line 66: The NCO implementation uses NWGES directories to store ics/lbcs. To save disk space, I decided to store only symlinks to actual lbcs files in this case.

parm/input.nml.FV3
```diff
@@ -95,14 +95,15 @@
...
+    n_sponge = 9
+    sg_cutoff = 10000.0
     n_zs_filter = 0
     na_init = 1
     ncep_ic = .false.
     nggps_ic = .true.
     no_dycore = .false.
     nord = 3
-    npz = 64
+    npz = 65
```
danielabdi-noaa: Increase number of levels to 65 in this PR.

scripts/exregional_get_extrn_mdl_files.sh
```diff
@@ -69,12 +69,26 @@ if [ "${ICS_OR_LBCS}" = "ICS" ]; then
...
   fi
-  last_time=$((TIME_OFFSET_HRS + FCST_LEN_HRS))
+  end_hr=$FCST_LEN_HRS
+  if [ $BOUNDARY_LEN_HRS -gt $end_hr ]; then
+     end_hr=$BOUNDARY_LEN_HRS
+  fi
+  # There is a difference between RRFS_dev & SRW regarding download of 0th
+  # hour lbcs. RRFS_dev downloads 0th hour lbcs while SRW does not.
+  # I tried to make them the same but RRFS_dev workflow skips make_ics for
+  # some cycles so it needs to download 0th hour lbcs.
```
danielabdi-noaa: Please read this comment in case we have a better way to handle this difference.

scripts/exregional_make_ics.sh
```diff
@@ -532,7 +533,7 @@ settings="
  'mosaic_file_target_grid': ${FIXlam}/${CRES}${DOT_OR_USCORE}mosaic.halo$((10#${NH4})).nc,
  'orog_dir_target_grid': ${FIXlam},
  'orog_files_target_grid': ${CRES}${DOT_OR_USCORE}oro_data.tile${TILE_RGNL}.halo$((10#${NH4})).nc,
- 'vcoord_file_target_grid': ${FIXam}/global_hyblev.l65.txt,
+ 'vcoord_file_target_grid': ${FIXam}/global_hyblev_fcst_rrfsL65.txt,
```
danielabdi-noaa: To match the increase to 65 levels used in RRFS_dev, this specific file needs to be used.

scripts/exregional_make_ics.sh
```diff
@@ -672,6 +673,20 @@ fi
...
+#
+#-----------------------------------------------------------------------
+#
+for i in ${INPUT_DATA}/*.nc; do
+    file=$(basename $i)
+    prefix="${NET}.${cycle}${dot_ensmem}."
+    file=${file#"$prefix"}
+    file=${file/f000/000}
+    ln_vrfy -sf $i ${INPUT_DATA_NWGES}/${file}
+done
```
danielabdi-noaa, line 690: Generating symlinks in NWGES directories.

scripts/exregional_make_lbcs.sh
```diff
@@ -99,15 +99,6 @@ fi
...
 cd_vrfy $DATA
-
-if [ "${FCST_LEN_HRS}" = "-1" ]; then
-  CYCLE_IDX=$(( ${cyc} / ${INCR_CYCL_FREQ} ))
-  FCST_LEN_HRS=${FCST_LEN_CYCL[$CYCLE_IDX]}
-fi
-LBC_SPEC_FCST_HRS=()
-for i_lbc in $(seq ${LBC_SPEC_INTVL_HRS} ${LBC_SPEC_INTVL_HRS} $(( FCST_LEN_HRS+LBC_SPEC_INTVL_HRS )) ); do
-  LBC_SPEC_FCST_HRS+=("$i_lbc")
-done
```
danielabdi-noaa, line 111: I have removed LBC_SPEC_FCST_HRS because RRFS figures out the forecast hour in a different way. Look lines 523-525

scripts/exregional_make_lbcs.sh
```diff
@@ -363,7 +355,12 @@ fi
 #-----------------------------------------------------------------------
 #
 num_fhrs="${#EXTRN_MDL_FHRS[@]}"
-for (( i=0; i<${num_fhrs}; i++ )); do
+bcgrp10=${bcgrp#0}
+bcgrpnum10=${bcgrpnum#0}
+for (( ii=0; ii<${num_fhrs}; ii=ii+bcgrpnum10 )); do
+  i=$(( ii + bcgrp10 ))
+  if [ ${i} -lt ${num_fhrs} ]; then
+    echo " group ${bcgrp10} processes member ${i}"
```
danielabdi-noaa, line 367: This is the part that necessitates workflow update. LBCS files are downloaded in parallel in RRFS.

scripts/exregional_get_extrn_mdl_files.sh
christinaholtNOAA: Does it makes sense to tie this to cycling instead of some other piece of information? It's not really about RRFS, but about whether the workflow is making initial conditions (running the make_ics task) that include boundary conditions or not, in the case of DA.

Something like this (sudo code) at the script level here:

```
if [ $NEED_ALL_LBCS is True ] ; then
  first_time=$((TIME_OFFSET_HRS))
else
  first_time=$((TIME_OFFSET_HRS + LBC_SPEC_INTVL_HRS ))
fi
```

Where the workflow task sends the information via environment variable for the task when it make sense in the context of the workflow.

Essentially this is a cycle-dependent variable, so should probably be treated as such.

scripts/exregional_get_extrn_mdl_files.sh
```diff
@@ -69,12 +69,26 @@ if [ "${ICS_OR_LBCS}" = "ICS" ]; then
...
-  first_time=$((TIME_OFFSET_HRS + LBC_SPEC_INTVL_HRS))
   if [ "${FCST_LEN_HRS}" = "-1" ]; then
     CYCLE_IDX=$(( ${cyc} / ${INCR_CYCL_FREQ} ))
     FCST_LEN_HRS=${FCST_LEN_CYCL[$CYCLE_IDX]}
   fi
-  last_time=$((TIME_OFFSET_HRS + FCST_LEN_HRS))
+  end_hr=$FCST_LEN_HRS
+  if [ $BOUNDARY_LEN_HRS -gt $end_hr ]; then
+     end_hr=$BOUNDARY_LEN_HRS
+  fi
```
christinaholtNOAA, line 82: I haven't seen `BOUNDARY_LEN_HRS` defined here. Is that meant to be a run-time argument from the XML or a config_default?

scripts/exregional_make_lbcs.sh
christinaholtNOAA: This should not be removed. I think that AQM and SRW both use this option to find the LBCS. 

`EXTRN_MDL_FHRS` is sourced from the summary file written out by the get_extrnl_lbcs task, and I have to assume that @chan-hoo had a reason for not using that value, choosing this one instead in PR #613.

scripts/exregional_make_lbcs.sh
christinaholtNOAA: Please explain what you mean by LBCS being downloaded in parallel in RRFS. 

Where are they downloaded from and in what task? Should this task be run as a metatask over each forecast lead time, or is it a group of lead times?

scripts/exregional_make_lbcs.sh
```diff
@@ -456,7 +453,7 @@ settings="
  'mosaic_file_target_grid': ${FIXlam}/${CRES}${DOT_OR_USCORE}mosaic.halo$((10#${NH4})).nc,
  'orog_dir_target_grid': ${FIXlam},
  'orog_files_target_grid': ${CRES}${DOT_OR_USCORE}oro_data.tile${TILE_RGNL}.halo$((10#${NH4})).nc,
- 'vcoord_file_target_grid': ${FIXam}/global_hyblev.l65.txt,
+ 'vcoord_file_target_grid': ${FIXam}/global_hyblev_fcst_rrfsL65.txt,
```
christinaholtNOAA: This file is going to have to be a configurable setting, and we will have to have it default to the original text file.

ush/config_defaults.yaml
```diff
@@ -1014,6 +1021,21 @@ nco:
...
+  RRFSE_NWGES: '{{ OPSROOT }}/nwges'
+  RRFSE_NWGES_BASEDIR: '{{ RRFSE_NWGES }}'
+  ENSCTRL_NWGES: '{{ OPSROOT }}/nwges'
+  ENSCTRL_NWGES_BASEDIR: '{{ ENSCTRL_NWGES}}'
+
+  ENSCTRL_COMIN: '{{ COMIN_BASEDIR}}'
+  ENSCTRL_COMOUT: '{{ COMOUT_BASEDIR}}'
+  ENSCTRL_COMROOT: '{{ ENSCTRL_COMOUT }}'
+  ENSCTRL_COMIN_BASEDIR: '{{ ENSCTRL_COMIN }}'
+  ENSCTRL_COMOUT_BASEDIR: '{{ ENSCTRL_COMROOT}}/{{ envir}}'
```
christinaholtNOAA: We shouldn't include any RRFSE or ensemble-related variables here as they are out of context of this PR.

ush/config_defaults.yaml
```diff
@@ -1466,6 +1488,26 @@ task_get_extrn_lbcs:
...
+  #  When search boundary conditions from previous cycles in prep_start step,
+  #  the search will start at cycle before (this parameter) of current cycle.
+  #  For example: 0 means search start at the same cycle lbcs directory.
+  #               1 means search start at 1-h previous cycle  lbcs directory.
+  #               2 means search start at 2-h previous cycle  lbcs directory.
+  #
+  #-----------------------------------------------------------------------
+  #
+  LBCS_SEARCH_HRS: 6
+  EXTRN_MDL_LBCS_SEARCH_OFFSET_HRS: 0
```
christinaholtNOAA, line 1562: As far as I can tell, the functional part that uses these variables is not included with this PR. It might be good to either include that, or remove these variables.

christinaholtNOAA (changes requested):
Daniel, thanks for preparing this PR. I've left some comments below that focus on making sure that we don't change existing behavior while capturing RRFS configuration settings, and that the mods stay in scope for the intent of this PR.

scripts/exregional_make_lbcs.sh
danielabdi-noaa: Sorry I meant to say that `make_lbcs` is now parallel not `get_lbcs`. This does seem to help it finish much quicker. The `make_lbcs` task is now a metatask with something like this:
```
  metatask_make_lbcs_mem#mem#:
    var:
      bcgrp: '{% for h in range(0, workflow.BOUNDARY_PROC_GROUP_NUM) %}{{ " %02d" % h  }}{% endfor %}'
    task_make_lbcs_#bcgrp#_mem#mem#:
      account: '&ACCOUNT;'
```
The default for `BOUNDARY_PROC_GROUP_NUM` is 1 but is set to 10 in the RRFS test case to make the task finish faster.

ush/config_defaults.yaml
danielabdi-noaa: `LBCS_SEARCH_HRS` is used only in the workflow:
```
           <taskdep task="&TN_MAKE_ICS;{{ uscore_ensmem_name }}"/>
           <or>
             {%- for h in range(0,lbcs_search_hrs) %} 
             <metataskdep metatask="&TN_MAKE_LBCS;{{ uscore_ensmem_name }}" cycle_offset="-{{ h }}:00:00"/>
             {%- endfor %}
           </or>
```
The other one is used in task `prepstart` ex-script, but since it relates to LBCS, I've put it here.

ush/config_defaults.yaml
danielabdi-noaa: Removed

scripts/exregional_get_extrn_mdl_files.sh
danielabdi-noaa: Added this and related variables in config_defaults.yaml.

scripts/exregional_make_lbcs.sh
danielabdi-noaa: @chan-hoo Please comment on this change. I was planning to run an AQM test for this particular change, but you maybe able to see an issue or lack thereof immediately. My thought was that if LBCS are downloaded for every forecast hour, `EXTRN_MDL_FHRS` maybe the same as your construction.

scripts/exregional_make_lbcs.sh
danielabdi-noaa: @christinaholtNOAA The problem with this change is that it comes with associated changes in `input.nml.FV3` so it is difficult to make this configurable unless we do the same for `input.nml.FV3`. I am thinking to postpone these two changes since they would also require update to `ufs_utils` verstion IIRC.

christinaholtNOAA: That would be fine, but we do want to make sure that this is an option for the next release, I think.

venitahagerty:
Machine: hera
Compiler: intel
Job: WE
Repo location: /scratch1/BMC/zrtrr/rrfs_ci/autoci/pr/1271094432/20230425193511/ufs-srweather-app
Build was Successful
Rocoto jobs started
Experiment done: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
2023-04-25 20:13:24 +0000 :: hfe07 :: This cycle is complete: Success
Experiment failed: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_RAP_suite_HRRR
2023-04-25 20:27:14 +0000 :: hfe07 :: Task run_MET_PcpCombine_fcst_APCP01h_mem000, jobid=44194882, in state DEAD (FAILED), ran for 13.0 seconds, exit status=256, try=2 (of 2)
Experiment done: grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta
2023-04-25 20:11:40 +0000 :: hfe07 :: This cycle is complete: Success
Experiment done: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2
2023-04-25 20:07:58 +0000 :: hfe07 :: This cycle is complete: Success
Experiment done: grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_2017_gfdlmp_regional_plot
2023-04-25 20:09:30 +0000 :: hfe07 :: This cycle is complete: Success
Experiment done: pregen_grid_orog_sfc_climo
2023-04-25 20:07:17 +0000 :: hfe07 :: This cycle is complete: Success
Experiment done: grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v15p2
2023-04-25 20:07:31 +0000 :: hfe07 :: This cycle is complete: Success
ci-hera-intel-WE

scripts/exregional_get_extrn_mdl_files.sh
danielabdi-noaa: I have made it so that NEED_ALL_LBCS is passed from the j-job. It is currently turned on for RRFS workflow but we could go a step further and put the flag in config_defaults.yaml.

jobs/JREGIONAL_MAKE_ICS
MichaelLueken: Moving forward, we will need to be careful with NCO implementations and the use of symlinks.  With the move to WCOSS2, NCO is no longer wanting symlinks to be used.  Notes from Steven Earle in global-workflow:

```
The issue described in the ticket goes much further than just the recent wave issue; I can update the bugzilla if that's desired.
DATA needs to be self contained for several reasons....
-- Portability
-- Contained IO, making management of the system/storage possible
-- Pristine place to save after failures for debug/troubleshooting later
```
While storing symlinks in develop is fine, storing symlinks for the official release will likely not work.

MichaelLueken (approved):
@danielabdi-noaa Overall, these changes look good to me.  I did want to point out a potential issue between NCO implementation and symlinks.  NCO created a Bugzilla ticket to remove symlinks from the global workflow.  From Steven Earle:

```
The issue described in the ticket goes much further than just the recent wave issue; I can update the bugzilla if that's desired.
DATA needs to be self contained for several reasons....
-- Portability
-- Contained IO, making management of the system/storage possible
-- Pristine place to save after failures for debug/troubleshooting later
```
While the use of symlinks are fine in develop, it is possible that NCO will require that they are replaced with copy/move/rsync in operations.  While this work can be done in the release branch for RRFS, it is still something to keep in mind moving forward.

danielabdi-noaa:
@MichaelLueken Thanks, I agree with your assessment of use of symlinks. The NCO implementation here that uses NWGES does not conform with latest NCO standards, however, RRFS_dev1 uses it throughout. It was decided a couple of months ago to use it as a temporary solution, but the plan was indeed to get rid of the NWGES directory and thus associated symlinks in the future.

panll (approved):
Tested on Hera, it works fine.

danielabdi-noaa:
@christinaholtNOAA Is there something I need to address other than the potential AQM issue? I did run the make_lbcs test with an AQM test case a while ago and it seems to work.

ush/config_defaults.yaml
```diff
@@ -886,6 +895,9 @@ workflow:
 
   CYCL_HRS_SPINSTART: []
   CYCL_HRS_PRODSTART: []
+  BOUNDARY_LEN_HRS: 0
```
christinaholtNOAA, line 898: Should this not default to `FCST_LEN_HRS`? Zero does not seem consistent with the description above.

ush/config_defaults.yaml
```diff
@@ -886,6 +895,9 @@ workflow:
...
+  BOUNDARY_LONG_LEN_HRS: 0
```
christinaholtNOAA, line 899: This one isn't used in the logic here at all as far as I can tell. Does it need to be?

ush/config_defaults.yaml
```diff
@@ -1041,6 +1060,12 @@ nco:
   COMIN_BASEDIR: '{{ COMROOT }}/{{ NET }}/{{ model_ver }}'
   COMOUT_BASEDIR: '{{ COMROOT }}/{{ NET }}/{{ model_ver }}'
 
+  #
+  # New additions from RRFS_dev1
+  #
+  NWGES: '{{ OPSROOT }}/nwges'
+  NWGES_BASEDIR: '{{ NWGES }}'
```
christinaholtNOAA, line 1067: I still am unclear why there needs to be both an NWGES and NWGES_BASEDIR.

ush/config_defaults.yaml
```diff
@@ -1066,11 +1091,18 @@ rrfs:
   #
   #-----------------------------------------------------------------------
   #
+  # Set parameters associated with running retrospective experiments.  Definitions:
+  #
+  # DO_RRFS_DEV:
+  # Flag to turn on RRFS dev workflow. Should be removed
+  # in the future once SRW and RRFS dev workflows are completely integrated
+  #
```
christinaholtNOAA, line 1099: I don't think this should be included at this time, especially with the caveat that it should be removed. Why not just set `NEED_ALL_LBCS` to what it should be?

christinaholtNOAA (changes requested):
I'm sorry for the delay in re-reviewing. I've left a few more comments below. It seems we're still waiting on @chan-hoo to weigh in, as well.

ush/config_defaults.yaml
danielabdi-noaa: The 0th boundary condition is set only for RRFS not for SRW or AQM, so we need to tell it when it is RRFS that is running. The `DO_RRFS_DEV` is also used in other places where things are differently for RRFS so I think it is good to have it. Otherwise I will have to define NEED_ALL_LBCS and other similar flags in `config_defaults.yaml` when one DO_RRFS_DEV will suffice for all things that are done differently in RRFS.

ush/config_defaults.yaml
danielabdi-noaa: It is used in the workflow for LBCS task just like LBCS_SEARCH_HRS below.

ush/config_defaults.yaml
danielabdi-noaa: BOUNDARY_LEN_HRS can be longer than FCST_LEN_HRS because RRFS pre-fetches boundary conditions in advance not only for the current cycle forecast lenght. I think I asked this question a while ago and MIng clarified it like that. Here is the relevant code in make_lbcs
```
  end_hr=$FCST_LEN_HRS
  if [ $BOUNDARY_LEN_HRS -gt $end_hr ]; then
     end_hr=$BOUNDARY_LEN_HRS
  fi
```

ush/config_defaults.yaml
danielabdi-noaa: That is the way it is in RRFS_dev1. `NWGES_BASEDIR` can be set to `CYCLE_DIR` in community mode and to `NWGES` in NCO mode. 
https://github.com/NOAA-GSL/regional_workflow/blob/feature/RRFS_dev1/ush/setup.sh#L1402-L1413
I am not saying this can not be removed, but that the implementation mirrors it.

ush/config_defaults.yaml
danielabdi-noaa: I see now you are talking about the default value, not that it should be set to `FCST_LEN_HRS`.
RRFS_dev1 sets both variables to default to 0
https://github.com/NOAA-GSL/regional_workflow/blob/feature/RRFS_dev1/ush/config_defaults.sh#L581-L582
The loops in the workflow that use these variables will behave differently if defaulted to `FCST_LEN_HRS`.
Can we just leave things as they are as long as it is consistent with RRFS_dev1 atleast for now?  We can
thinker with things later once we have a setup to test things with.

scripts/exregional_make_lbcs.sh
christinaholtNOAA: Please leave this chunk of code for now and adopt the way that SRW does it. I believe this logic replaced the old logic that was in RRFS_dev1, and should not be reverted. Once that is done, I'm fine with this being merged.

danielabdi-noaa: @christinaholtNOAA Note that if I leave that code as is, it will not work for RRFS (only SRW and AQM) because we are fetching more LBCS than the forecast length hour. I would have to either modify it to mirror that in `get_lbcs` which is:
```
  last_time=$((TIME_OFFSET_HRS + FCST_LEN_HRS))
  end_hr=$FCST_LEN_HRS
  if [ $BOUNDARY_LEN_HRS -gt $end_hr ]; then
     end_hr=$BOUNDARY_LEN_HRS
  fi
  # Download 0th hour lbcs if requested for it, mostly for DA
  if [ ${NEED_ALL_LBCS} = "TRUE" ]; then
    first_time=$((TIME_OFFSET_HRS))
  else
    first_time=$((TIME_OFFSET_HRS + LBC_SPEC_INTVL_HRS ))
  fi
  last_time=$((TIME_OFFSET_HRS + end_hr))


  fcst_hrs="${first_time} ${last_time} ${LBC_SPEC_INTVL_HRS}"
  ```
  Or use the hours  witten in the summary file, which should correspond to the above hours. I think this is a better strategy.  I have just run the make_lbcs step of AQM but did not re-build from scratch. It would be good if @chan-hoo can comment on this though.

christinaholtNOAA: Maybe you can reach out to ChanHoo directly, instead of through comments here?

chan-hoo: @danielabdi-noaa @christinaholtNOAA, I apologize for my late response again. I was able to complete the NRT sample run of AQM successfully. I confirm that this change works for AQM correctly too.

danielabdi-noaa: @chan-hoo Thanks for testing. Glad that the change works in AQM.

christinaholtNOAA (approved):

MichaelLueken:
@danielabdi-noaa There were a few failures in the Jenkins tests, but they were expected failures:

On Cheyenne GNU - the `nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15_thompson_mynn_lam3km` test failed intermittently in the run_post tasks.  This has been an issue since the test has been introduced.

Following the PrgEnv update on Gaea, it is currently impossible to build and run the SRW and weather model on the machine.  Working with Natalie to get this corrected as soon as possible.

On Hera Intel - the `grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta` test failed in run_fcst (issue #731).

Since the failures are expected, I will now move forward with merging this PR to develop.

PR #701, opened 2023-03-28
[develop] Add jinja templating to all config file formats.

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 

Mainly addresses issue #700
- Add jinja templating for a "block of lines" for all config file formats. This is different from what `extend_yaml` does, namely, fill template values of dictionary entries. Given a second file containing the context (dictionary) for rendering the jinja template, you can now do this:
```
{% if DO_REAL_TIME %}
WALL_LIMIT_ANAL: 16:00:00
WALL_LIMIT_FCST: 24:00:00
{% else %}
WALL_LIMIT_ANAL: ""
WALL_LIMIT_FCST: ""
{% endif %}
```
instead of
```
WALL_LIMIT_ANAL: '{% if DO_REAL_TIME %}"16:00:00"{% else %}""{% endif %}'
WALL_LIMIT_FCST: '{% if DO_REAL_TIME %}"16:00:00"{% else %}""{% endif %}'
```
If these variables are defined only for realtime runs, the first option without the `{% else %}` will not define them, while the second option still has to define the variables to null string. 

I think this feature is required for the upcoming rrfs workflow with yaml template files.

Other changes include:
- Parse xml attributes as dictionary entries
- Parse repeated entries into a list
- Bug fix with matching single/double quotes

With this change you should be able to parse a rendered xml file (not the one in parm) and convert it to any of the five formats.
```
./config_utils.py -c $EXPTDIR/FV3LAM_wflow.xml -o yaml
```
If the context for rendering is available in file `context.yaml`, we can render and convert the templated file instead like this
```
./config_utils.py -c $PARMdir/FV3LAM_wflow.xml -x $PWD/context.yaml -o yaml
```
---
**Edit: Work done after this PR's approval**
- Added F90 namelist as one of the supported config formats. You should be able to convert to/from nml files now including workflow xml files.
- Made shell/ini formats work with deeply nested dictionaries. This means workflow files in these formats should work now, but the format does not look pretty, so this is just for completeness i.e. enable all six formats for workflow configuration.
- Some bug fixes to make sure that a `yaml/json` workflow file work as intended. Tested with:
```
custom_GFDLgrid
community_ensemble_2mems
```
The test is conducted by converting the generated `FV3LAM_wflow.xml` file say to yaml using this tool, and then back to xml.
The resulting xml file replaces the one in EXPTDIR and the test re-run again. The above tests succeeded using this verification method.

---

### Type of change

[...]

PR #721, opened 2023-04-10
[develop] Integrate jinja-templated workflow and add some rrfs workflow files.

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 

This PR extends upon work done in PR #701 to integrate jinja support to config and workflow files. Enabling jinja in config/workflow files turned out to be not as straight forward as I expected. Some things that complicated things include

- We relied on `extend_yaml` that only renders jinja-templates in "values" of dictionaries. While this works just fine, we
may need jinja-templating throughout the config files e.g. `if-conditions`
- Some newly added filters like `include` load a yaml config file and do not work when it is templated. The solution I used for this is to just copy/paste the files to be included. The jinja `{% include %}` construct could potentially used but I haven't investigated further
- Rendering for "values" in `extend_yaml` has a dependency chain because it is referring to the same dictionary to update values. To render other templates, you will have to provide a secondary dictionary.
- Anchors used in the new workflow files need to be unique because they may not be expanded and eliminated before filling up jinja templates
- We need to construct the config files (default/machine/user) before constructing workflow config files, since the latter use them. So `setup.py` is modified to construct config files before workfow files.

To test this functionality, I have added the RRFS workflow jinja-templated throughout. Where I think this may come in handly is in dependencies where a lot of jinja-template d code is used. The jinja-templated workflow files mirror that of the xml file.
I tested this setup in PR #647 where none of the RRFS tasks are added, so while it will generate the workflow, it needs to be tested/fixed as each task goes in. 

This PR has modifications from PR#701 and PR#647 since it needs both for a successful RRFS_Dev workflow generation.

### Type of change

## TESTS CONDUCTED: 

- [x] hera.intel
- [ ] orion.intel
- [ ] cheyenne.intel
- [ ] cheyenne.gnu
- [ ] gaea.intel
- [ ] jet.intel
- [ ] wcoss2.intel
- [ ] Jenkins
- [ ] fundamental test suite

## CHECKLIST

## LABELS (optional): 

- [ ] Work In Progress
- [ ] bug
- [ ] enhancement
- [ ] documentation
- [ ] release
- [ ] high priority
- [ ] run_ci
- [ ] Needs Cheyenne test 
- [ ] Needs Jet test 
- [ ] Needs Hera test 
- [ ] Needs Orion test 
- [ ] help wanted


christinaholtNOAA (commented):
I do not think this will, in general work. We should work offline to discuss the design pattern and how new workflows should and can be added with the existing tools. The behavior of the existing tool will get us the full flexibility we need for RRFS.

Specifically, putting these switches in convolutes the process of building a workflow based on a configuration from component pieces. For example, if you want to modify cycles, it should be done by config file, not a switch defined in the workflow. Let me find some time to have a more thorough discussion about the options and why I don't think we want to go this direction.

danielabdi-noaa:
@christinaholtNOAA Ok I will cose this PR.

PR #676, opened 2023-03-15
[develop] Adds a YAML interface for creating a Rocoto XML.

christinaholtNOAA:
## DESCRIPTION OF CHANGES: 

Refactors the creation of a Rocoto XML to use a very generic Jinja2 template that is flexible enough to meet the needs of various workflow configurations supported by SRW. Specifically, it allows for a completely arbitrary workflow to be created under SRW, which includes the addition of completely arbitrary tasks on top of the predefined ones. 

The workflow has been refactored to allow for the definition a workflow given specific entries in the rocoto section of the configuration files. The paradigm shifts to telling the configuration which sets of tasks to run, and removing a section when certain pre-configured tasks should not run.

The high level overview of changes include:

- Moved batch job related config variables out of config_default.yaml sections, and into the Rocoto config files for the given job. This resulted in the removal of entire sections of config_default.yaml. References to those removed sections were removed throughout SRW.
- The transition of the FV3LAM_wflow.xml template to a complete generic interpretation of YAML configs.
- The addition of parm/wflow files in groups of tasks that are commonly run together. All the Rocoto-required information for each task now lives in these yaml files.
- A modification to specifying varying forecast lengths to remove a dependence on writing all cycles in a workflow to the config file. The FCST_LEN_CYCL array is expected to correspond to hours in the day associated with a cycling freqency. For example, a list of 4 forecast lengths with a 6 hour cycling freqency indicates that the forecast lengths correspond to 0, 6, 12, and 18 UTC.
- The removal of the workflow_switches section. Whether tasks run is now based on their being defined as part of the rocoto section of the config file.
- The number of processors is now included as an environment variable passed via Rocoto instead of computed in the scripts since Rocoto holds the source of that information.
- Edits to the test config files to reflect how the test should have originally been run
- Writes a new file to the experiment directory -- rocoto_defns.yaml. It contains the rocoto section of the config that has been removed from the var_defns.yaml.

A follow-on PR will be needed to create the documentation necessary to support this change, although some documentation will be included with this PR. 

I hope to prioritize this PR as it has been, and will continue to be very difficult to support as changes are made to the existing workflow and scripts. I will work with others impacted by the change to ensure their workflow changes make it into the new workflow config files.

A demo of this tool is scheduled for this Thursday, March 16. It will be recorded.

[...]

Issue #688, opened 2023-03-21
PCP Combine tasks fail in NCO mode


mkavulich:
## Expected behavior
PCP Combine tasks should succeed.

## Current behavior
Running a PCP Combine task results in failure when running in NCO mode. The problem appears to be due to the `exregional_run_met_pcpcombine.sh` script not looking for forecast files in the correct location:

```
========================================================================
Entering script:  "exregional_run_met_pcpcombine.sh"
In directory:     "/scratch2/BMC/fv3lam/kavulich/UFS/workdir/test_develop/ufs-srweather-app/scripts"

This is the ex-script for the task that runs the METplus PcpCombine
that combines hourly accumulated precipitation (APCP) data to generate
files containing multi-hour accumulated precipitation (e.g. 3-hour, 6-
hour, 24-hour).  The input files can come from either observations or
a forecast.
========================================================================
Initial (i.e. before filtering for missing files) set of forecast hours
is:
  fhr_array = ( "3" "6" )

The file (fp) for the current forecast hour (fhr; relative to the cycle
date cdate) is missing:
  fhr = "1"
  cdate = "2019061500"
  fp = "/scratch2/BMC/fv3lam/kavulich/UFS/workdir/test_develop/expt_dirs/fundamental/MET_ensemble_verification/2019061500/mem001/postprd/rrfs.t00z.prslev.f001.rrfs_conus_25km.grib2"
Excluding the current forecast hour from the list of hours passed to the
METplus configuration file.

The file (fp) for the current forecast hour (fhr; relative to the cycle
date cdate) is missing:
  fhr = "4"
  cdate = "2019061500"
  fp = "/scratch2/BMC/fv3lam/kavulich/UFS/workdir/test_develop/expt_dirs/fundamental/MET_ensemble_verification/2019061500/mem001/postprd/rrfs.t00z.prslev.f004.rrfs_conus_25km.grib2"
Excluding the current forecast hour from the list of hours passed to the
METplus configuration file.

Final (i.e. after filtering for missing files) set of foreast hours is
(written as a single string):
  fhr_list = ""

FATAL ERROR:
ERROR:
  From script:  "exregional_run_met_pcpcombine.sh"
  Full path to script:  "/scratch2/BMC/fv3lam/kavulich/UFS/workdir/test_develop/ufs-srweather-app/scripts/exregional_run_met_pcpcombine.sh"
The list of forecast hours for which to run METplus is empty:
  FHR_LIST = []
[...]

PR #722, opened 2023-04-10
[develop] Fix issues on AQM and NCO mode caused by new YAML interface PR #676

### LINKED ISSUE #709: AQM configuration files do not work after YAML change PR #676
## Expected behavior
- The AQM configuration files `config.aqm.community.yaml` and `config.aqm.nco.realtime.yaml` generate XML files.

## Current behavior
- `config.aqm.community.yaml`: XML file is generated, but the workflow dependencies are not set correctly.
- `config.aqm.nco.realtime.yaml`: XML file is not generated.

## Steps To Reproduce
1. cp config.aqm.nco.realtime.yaml config.yaml
2. python3 generated_FV3LAM_wflow.py

chan-hoo:
## DESCRIPTION OF CHANGES: 
- Fix AQM configuration issues on both `community` and `nco` modes.
- Fix workflow entity issues on `nco` mode.

### Type of change

## TESTS CONDUCTED: 
Fundamental WE2E tests on WCOSS2

- [x] hera.intel
- [ ] orion.intel
- [ ] cheyenne.intel
- [ ] cheyenne.gnu
- [ ] gaea.intel
- [ ] jet.intel
- [x] wcoss2.intel
- [ ] Jenkins
- [ ] fundamental test suite

## ISSUE: 
Fixes issue mentioned in #709 

## CHECKLIST


jobs/JREGIONAL_RUN_POST
```diff
@@ -138,16 +138,14 @@ Call to ex-script corresponding to J-job \"${scrfunc_fn}\" failed."
...
 #-----------------------------------------------------------------------
 #
-if [ "${FCST_LEN_HRS}" = "-1" ]; then
-  for i_cdate in "${!ALL_CDATES[@]}"; do
-    if [ "${ALL_CDATES[$i_cdate]}" = "${PDY}${cyc}" ]; then
-      FCST_LEN_HRS="${FCST_LEN_CYCL_ALL[$i_cdate]}"
-      break
-    fi
-  done
+if [ "${NUM_FCST_LEN_CYCL}" -gt "1" ]; then
```
christinaholtNOAA: This new variable does not need to be added because the same can be accomplished by bash using the existing variables, which reduces unnecessary complexity. For example:

```
if [ ${#FCST_LEN_CYCL[@]} -ne 1 ]; then
```

jobs/JREGIONAL_RUN_POST
```diff
@@ -138,16 +138,14 @@ Call to ex-script corresponding to J-job \"${scrfunc_fn}\" failed."
...
+  cyc_mod=$(( ${cyc} - ${DATE_FIRST_CYCL:8:2} ))
+  CYCLE_IDX=$(( ${cyc_mod} / ${INCR_CYCL_FREQ} ))
+  FCST_LEN_HRS=${FCST_LEN_CYCL[$CYCLE_IDX]}
```
christinaholtNOAA, line 144: I don't think this logic works just right. 

If DATE_FIRST_CYCL starts at 18Z, and we're looking at the next 00Z cycle, we get:

```
cyc_mod=$(( 00 - 18 ))
CYCLE_IDX=$(( -18 / 6 ))
```

So we're left with a negative `CYCLE_IDX` and can't get the right information in the list. I'd suggest that we add the requirement on `FCST_LEN_CYCL` to define a list that starts from 00Z and increments by the `FCST_LEN_CYCL` if you need to define variable forecasts lengths.

ush/config_defaults.yaml
```diff
@@ -857,6 +857,7 @@ workflow:
   FCST_LEN_HRS: 24
   FCST_LEN_CYCL:
     - '{{ FCST_LEN_HRS }}'
+  NUM_FCST_LEN_CYCL: 1
```
christinaholtNOAA: As suggested above, this may be able to be removed given that the information is already available by interrogating `FCST_LEN_CYCL`.

christinaholtNOAA (commented):
@chan-hoo Again, I am sorry that I did not see your comments on the Issue earlier.

I have just a concerns about these changes below.

jobs/JREGIONAL_RUN_POST
chan-hoo: @christinaholtNOAA, you're correct. I'll modify it soon.

jobs/JREGIONAL_RUN_POST
chan-hoo: @christinaholtNOAA, I don't agree with you. {cyc} starts from a lower value. if {cyc} includes "06" and "18", its order will be ["06", "18"] (not ["18", "06"]). I think the 'cyc_mod' will not have a negative value in any cases because these variable forecast length hours are only set per day as you modified it in your former PR:  https://github.com/ufs-community/ufs-srweather-app/blob/662ff2ae4679e09305140deadaf84f30d036597d/ush/setup.py#L681  
What do you think about this? In the above example of ["06", "18"], the current condition will fail because their indexes will be [ "06/12", "18/12"]. With a new condition, they will be ["0", "1"] (["(6-6)/12", "(18-6)/12"]).

ush/config_defaults.yaml
chan-hoo: Agreed. I'll remove it soon.

etc/lmod-setup.sh
```diff
@@ -14,13 +14,15 @@ else
...
-$has_me && set -e
+if [ "$L_MACHINE" != wcoss2 ]; then
+  [[ ${SHELLOPTS} =~ nounset ]] && has_mu=true || has_mu=false
+  [[ ${SHELLOPTS} =~ errexit ]] && has_me=true || has_me=false
+  $has_mu && set +u
+  $has_me && set +e
+  source /etc/profile
+  $has_mu && set -u
+  $has_me && set -e
+fi
```
danielabdi-noaa, line 25: I wonder why wcoss2 no longer needs sourcing of /etc/profile?

jobs/JREGIONAL_RUN_POST
```diff
@@ -8,7 +8,7 @@
 #
 #-----------------------------------------------------------------------
 #
-
+set -x
```
danielabdi-noaa: I noticed you adding "set -x" in AQM scripts sometimes. Note that DEBUG=TRUE should be able to do that for you so no need to add it in the scripts.

scripts/exregional_aqm_lbcs.sh
```diff
@@ -50,6 +50,7 @@ lateral boundary conditions.
 #
 #-----------------------------------------------------------------------
 #
+set -x
```
danielabdi-noaa: Here too.

scripts/exregional_bias_correction_o3.sh
```diff
@@ -50,6 +50,7 @@ This is the ex-script for the task that runs BIAS-CORRECTION-O3.
 #
 #-----------------------------------------------------------------------
 #
+set -x
```
danielabdi-noaa: Here

ush/setup.py
```diff
@@ -728,6 +727,24 @@ def get_location(xcs, fmt, expt_cfg):
...
+            cycl_next = date_to_str(date_first_cycl + timedelta(hours=incr_cycl_freq), format="%Y%m%d%H")            
+
+        first = date_first_cycl.strftime("%Y%m%d%H")
+        last = date_last_cycl.strftime("%Y%m%d%H")
+
+        cycled_from_second = []
+        if cycl_next != first:
+            cycled_from_second.append(f'{cycl_next}00 {last}00 {incr_cycl_freq}:00:00')
+
+        rocoto_config['cycledefs']['cycled_from_second'] = cycled_from_second
```
danielabdi-noaa: Is it possible to define the cycle `cycled_from_second` from the workflow yaml file intstead?

ush/setup.py
```diff
@@ -1109,8 +1126,9 @@ def get_location(xcs, fmt, expt_cfg):
         for nco_var in nco_vars:
             nco_config[nco_var.upper()] = exptdir
 
-        # Set the rocoto string for the fcst output location
+        # Set the rocoto string for the fcst output location and comin directory
         rocoto_config["entities"]["FCST_DIR"] = "{{ nco.COMOUT_BASEDIR }}/@Y@m@d@H"
+        rocoto_config["entities"]["COMIN_DIR"] = "{{ nco.COMIN_BASEDIR }}/@Y@m@d@H"
```
danielabdi-noaa: How about this one?

scripts/exregional_run_post.sh
```diff
@@ -166,7 +166,7 @@ fi
 # Set the names of the forecast model's write-component output files.
 #
 if [ "${RUN_ENVIR}" = "nco" ]; then
-    DATAFCST=$DATAROOT/run_fcst${dot_ensmem/./_}.${share_pid}
+    DATAFCST=$DATAROOT/run_fcst_mem${ENSMEM_INDX}.${share_pid}
```
danielabdi-noaa, line 169: Does the directory name also have _mem000 for non-ensemble runs now? I thought this addition affected only task names.

danielabdi-noaa (approved):
Looks good to me. Some minor comments.

etc/lmod-setup.sh
chan-hoo: @danielabdi-noaa, I was told that this should not be run in the scripts on wcoss2. Otherwise, a user will lose access to wcoss2.

jobs/JREGIONAL_RUN_POST
chan-hoo: I didn't know this option. Thanks for the info. I'll remove "set -x" from the scripts.

ush/setup.py
chan-hoo: I have no idea about this. Could you let me know how I can define it in the yaml fie?

scripts/exregional_run_post.sh
chan-hoo: Yes, it does. Without this change, the output files of the run_post are not generated. The issue is that the log file of `run_post` says that the dir/file is not found, but the task is completed successfully and generates "SUCCESS" message at the end.

danielabdi-noaa: Ok thanks! I am going to have to make similar modifications in rrfs dev where `dot_ensmem` is used.
Can we modify "dot_ensmem" defintion in `job_preamble.sh` instead and keep the old way of setting DATAFCST?
I think we just have to remove " [ "${DO_ENSEMBLE}" = "TRUE" ] " in the if condition
```
if [ "${RUN_ENVIR}" = "nco" ] && [ ! -z $ENSMEM_INDX ]; then
    export dot_ensmem=".mem${ENSMEM_INDX}"
else
    export dot_ensmem=
fi
```

danielabdi-noaa: But we have to be careful though, since this will also modify the names of output files.
```
${COMOUT}/${NET}.${cycle}${dot_ensmem}.prslev.f${fhr}.${POST_OUTPUT_DOMAIN_NAME}.grib2
```
Modifying `dot_ensmem` will add a ".mem000" to the name which we probably don't want.
I am not sure it is a good idea so forget that I suggested it.
Non-ensemble runs should not have been modified to have a mem000!

ush/setup.py
danielabdi-noaa: Possible but too long and hard to read i think -- yet another reason why jinja2 support is needed.
```
    cycled_from_second:
      - !startstopfreq ['{%- if workflow.DATE_FIRST_CYCL != workflow.DATE_LAST_CYCL %}{{ [workflow.DATE_FIRST_CYCL[0:8], workflow.INCR_CYCL_FREQ]|join }}{%- else %}{{workflow.DATE_FIRST_CYCL}}{%- endif %}', '{{workflow.DATE_LAST_CYCL}}', '{{workflow.INCR_CYCL_FREQ}}']
```
It is also not possible to define this only for AQM.

ush/setup.py
chan-hoo: @danielabdi-noaa, do you mean defining these from the workflow yaml file?

danielabdi-noaa: Yes, these should also be easy enough to support from there. I think we should define workflow related stuff from the workflow yaml files alone.

jobs/JREGIONAL_RUN_POST
christinaholtNOAA: I am still concerned about the use case given that there would be a strong coupling between the `FCST_LEN_CYCL` and the `DATE_FIRST_CYCL` and `DATE_LAST_CYCL` variables.

In the above example, I was suggesting that these might be the user-defined settings for two cycles, where the first is meant to be a short forecast and the second a long one:
```
DATE_FIRST_CYCL=2023030518
DATE_LAST_CYCL=2023030600
FCST_LEN_CYCL=["06", "18"]
```

This is a specific example of where I'm saying it might be better to require users to define `FCST_LEN_CYCL` as if it were all the possible daily cycles like this:
```
FCST_LEN_CYCL=["18", "06",  "06", "06"]
```

This means that we don't have to handle the "special" cases and aren't so heavily coupling the user-defined dates to the forecast lengths.

ush/setup.py
chan-hoo: @danielabdi-noaa, I set as follows, but it doesn't work:
```
    COMIN_DIR: '{% if user.RUN_ENVIR == "nco" %}{{ nco.COMIN_BASEDIR }}/{{ nco.RUN }}.@Y@m@d/@H{% else %}{{ nco.COMIN_BASEDIR }}/@Y@m@d@H{% endif %}'
```
How can I fix this?

chan-hoo: @danielabdi-noaa, I found a solution:
```
    COMIN_DIR: '{% if user.RUN_ENVIR == "nco" %}{{"{}/{}.@Y@m@d/@H".format(nco.COMIN_BASEDIR,nco.RUN)}}{% else %}{{"{
}/@Y@m@d@H".format(workflow.EXPTDIR)}}{% endif %}'
```

danielabdi-noaa: @chan-hoo I wonder why you needed to use format. Does it not render the jinja template correctly?

chan-hoo: @danielabdi-noaa, I just followed the existing one: 
'''
    LOGDIR: '{% if user.RUN_ENVIR == "nco" %}{{"{}/@Y@m@d".format(nco.LOGBASEDIR)}}{% else %}{{nco.LOGBASEDIR }}
'''
The issue was "nco.COMIN_BASEDIR". I had to change it "EXPTDIR".
If you don't prefer format, I'll change it. However, I found another issue:
'''
       - !startstopfreq ['{%- if workflow.DATE_FIRST_CYCL != workflow.DATE_LAST_CYCL %}{{ [workflow.DATE_FIRST_CYCL[0:8], workflow.INCR_CYCL_FREQ]|join }}{%- else %}{{workflow.DATE_FIRST_CYCL}}{%- endif %}', '{{workflow.DATE_LAST_CYCL}}', '{{workflow.INCR_CYCL_FREQ}}']
'''
When I run the above def, the result of `[workflow.DATE_FIRST_CYCL[0:8], workflow.INCR_CYCL_FREQ]|join }}` becomes "20230428600" but it should "202304280600". How can I set workflow.INCR_CYCL_FREQ to 2-digits in the expression?

danielabdi-noaa: @chan-hoo Formatting the cyle frequency probably works
```
"{:02d}".format(workflow.INCR_CYCL_FREQ)
```

danielabdi-noaa: Btw, you don't have to do this in this PR :) It is just my wish list to have workflow stuff defined entirely from the workflow yaml files instead of using setup.py

chan-hoo: Thanks, and your suggestion works well!

jobs/JREGIONAL_RUN_POST
chan-hoo: @christinaholtNOAA, I understand your concern. However, the change in this PR is better than the current one you made because it at least works for the special case. The current version will not work for both cases (my case as well as your case). In my opinion, the best solution is to use my original version:
```
  for i_cdate in "${!ALL_CDATES[@]}"; do
    if [ "${ALL_CDATES[$i_cdate]}" = "${PDY}${cyc}" ]; then
      FCST_LEN_HRS="${FCST_LEN_CYCL_ALL[$i_cdate]}"
      break
    fi
```
I understand that you didn't want to use `ALL_CDATES`, but this one will not have any issues. What do you think about that?

christinaholtNOAA: Leaning on ALL_CDATES is not a valid solution for when we want to run a real-time run indefinitely. Do you mind sharing the use case that you are trying to get running? The start and end date, and the forecast lengths you'd like to support?

chan-hoo: @christinaholtNOAA: 
1) Official NRT (Near Real Time) run: 4 cycles per day ("00" "06" "12" "18") and varying forecast length hours = ("06" "72" "72" "06").
2) Official Retro run: same 4 cycles per day and lengths as 1); 3 month period
3) Non-official test run: 2 cycles per day ("06" "18") for 3 days ("INCR_CYCL_FREQ: 12). This case causes the error in the current status. We don't support this case officially. However, I remember someone asked me if this case was available in the workflow. This is the reason why I am trying to update this part in my PR.

chan-hoo: @christinaholtNOAA, if you don't agree with my change, please provide us a reasonable solution. This PR takes so long for review now. The AQM users are not able to use the develop branch now and they are waiting for this PR to be merged. @MichaelLueken, could you please ask other reviewers to review this PR?

christinaholtNOAA (approved):
After chatting with @chan-hoo, we decided to open an Issue on the limitations of this formulation, and go ahead with this PR as-is.

MichaelLueken (approved):
@chan-hoo Thank you for working with @christinaholtNOAA and @danielabdi-noaa to address their concerns!  These changes look good to me, so approving and submitting Jenkins tests now.

MichaelLueken:
@chan-hoo The verification WE2E tests on Cheyenne and Hera failed (these are noted in issue #688).  All other tests successfully passed.  Will now move forward with merging this work.

PR #745, opened 2023-04-26
[develop] Bugfix find task function in setup.py

danielabdi-noaa:
## DESCRIPTION OF CHANGES: 

There seems to be a bug in `dict_find`. I am not sure how it worked so far.
- When a task is found but is deeper in the dictionary it returns `False`
- In some cases it returns `None` instead of `False`
- Logic for `task_make_lbcs` is wrong

### Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## TESTS CONDUCTED: 

- [ ] hera.intel
- [ ] orion.intel
- [ ] cheyenne.intel
- [ ] cheyenne.gnu
- [ ] gaea.intel
- [ ] jet.intel
- [ ] wcoss2.intel
- [ ] NOAA Cloud (indicate which platform)
- [ ] Jenkins
- [ ] fundamental test suite
- [ ] comprehensive tests (specify *which* if a subset was used)

## CHECKLIST

- [ ] My code follows the style guidelines in the Contributor's Guide
- [ ] I have performed a self-review of my own code using the Code Reviewer's Guide
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes need updates to the documentation. I have made corresponding changes to the documentation
- [ ] My changes do not require updates to the documentation (explain).
- [ ] My changes generate no new warnings
- [ ] New and existing tests pass with my changes
- [ ] Any dependent changes have been merged and published

## LABELS (optional): 

A Code Manager needs to add the following labels to this PR: 
- [ ] Work In Progress
- [ ] bug
- [ ] enhancement
- [ ] documentation
- [ ] release
- [ ] high priority
- [ ] run_ci
- [ ] run_we2e_fundamental_tests
- [ ] run_we2e_comprehensive_tests
- [ ] Needs Cheyenne test 
- [ ] Needs Jet test 
- [ ] Needs Hera test 
- [ ] Needs Orion test 
- [ ] help wanted

PR #732, opened 2023-04-17
[develop] Round 2 of overhaul to WE2E test suites (and other test improvements!)

mkavulich:
**Note to code manager: The label `run_we2e_fundamental_tests` should be renamed to `run_we2e_coverage_tests` after this PR is merged.**

## DESCRIPTION OF CHANGES: 
This test continues the overhaul of WE2E test suites as described in Issue #587 (specifically stage 3 and parts of stage 4 in [this comment](https://github.com/ufs-community/ufs-srweather-app/issues/587#issuecomment-1461049391)). The changes are summarized below, roughly in order of importance.

 - "fundamental" tests are replaced by "coverage" test suites. "fundamental" tests are returned to their original purpose: a lightweight set of tests to be run the same on all platforms. "coverage" tests now evenly distribute all comprehensive tests across all platform/compiler combinations for use in Jenkins testing.
 - "comprehensive" test list is updated to include all tests (except current known failures). For platforms that have known failures (for example, HPSS tests on on platforms without HPSS access), `comprehensive.<platform>[.<compiler>]` files are included to automatically run only the tests expected to succeed
 - Fix several existing failures
   - Use correct date format in grid_SUBCONUS_Ind_3km_ics_FV3GFS_lbcs_FV3GFS_suite_WoFS_v0
   - In `config_parser.py`, when populating a jinja template, keep dates in string format rather than converting to a `datetime` object (this fixes problem with `get_from_NOMADS_ics_FV3GFS_lbcs_FV3GFS`)
   - Fix unit tests for `retrieve_data.py`, there was a bug causing all tests to be run in nested subdirectories that eventually leads to failure when running all tests including HPSS retrieval
 - Remove several "get_from_HPSS" tests in favor of new unit tests for HPSS data in `test_retrieve_data.yaml`
 - Add several more dates and data sources to unit tests in `test_retrieve_data.yaml`
 - The example config files in the `ush/` directory (`config.community.yaml` and `config.nco.yaml`) are now included as WE2E tests (symbolically linked in the `tests/WE2E/test_configs/default_configs/` directory)
 - Remove long-known failing test `grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v16` (#359). This is now an old capability with only legacy support (global spectral model was retired in 2019) and there are no immediate plans to fix the bug.
 - `WE2E_summary*.txt` files are now written to the experiment directory rather than `tests/WE2E`
[...]

Issue #731, opened 2023-04-17
WE2E test `grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta` fails at the run_fcst step


mkavulich:
## Expected behavior
The `grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta` WE2E test should pass in the develop branch.

## Current behavior
The model segfaults.

## Machines affected
Jet and Hera, likely others.

## Steps To Reproduce
1. Compile latest develop branch
2. Run WE2E test `grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta`
3. Observe segmentation fault at run_fcst step

## Additional Information (optional)
This error was introduced with commit https://github.com/ufs-community/ufs-srweather-app/commit/ff9eddd0a93d04a3a937f6544cab0df7343f5f6c (PR #682) which made several changes to how the forecast model is run. Specifically the changes to the `model_configure` file and the default values of `OMP_NUM_THREADS_RUN_FCST` and `OMP_STACKSIZE_RUN_FCST` stand out to me as candidates for the source of the failure; these changes should be either reverted or properly justified and the test file modified appropriately.

## Output (optional)
And example of the failure can be found on Hera here: `/scratch2/BMC/fv3lam/kavulich/UFS/workdir/issue_587_round_2/expt_dirs/comprehensive/grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta`

MichaelLueken:
@mkavulich Following the merge of PR #745, this test now appears to be passing.  The Jenkins test for PR #475 had this test successfully pass and merging develop (updated with PR #475) into a local clone of PR #729 allowed the test to pass as well.

gspetro-NOAA:
This test was removed from the SRW App in PR #1201 . Closing the issue.

PR #729, opened 2023-04-13
[develop] Added environment variable LDFLAGS on macOS as requested by the corresponding module file

ywangwof:
## DESCRIPTION OF CHANGES: 

On macOS with GNU compiler, once the module file _build_macos_gnu.lua_ is loaded, it outputs a message to request for setting environment variable **LDFLAGS**:

>    Please export env. variable LDFLAGS after the module is successfully loaded:
>        > export LDFLAGS=\"-L\$MPI_ROOT/lib \" "
>   %

When using script _devbuild.sh_ however, it does not give users a chance to read/set this environment variable. I added three lines in the script _devbuild.sh_, which will set this environment variable automatically when the module **build_macos_gnu** is detected.

>     if [[ "${MODULE_FILE}" == "build_macos_gnu" ]]; then
>         export LDFLAGS="-L$MPI_ROOT/lib "
>     fi

This will make script _devbuild.sh_ working on macOS as it is on other platforms and will also avoid a _library not found_ error on macOS.

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## TESTS CONDUCTED: 

It is tested on Mac OS Monterey using GNU compiler and the latest HPC-stack installation (from develop branch). It should not affect the performance of this script on other platforms and it is also confirmed on Jet using Intel compiler.

- [ ] hera.intel
- [ ] orion.intel
- [ ] cheyenne.intel
- [ ] cheyenne.gnu
- [ ] gaea.intel
- [x] jet.intel
- [ ] wcoss2.intel
- [ ] NOAA Cloud (indicate which platform)
- [ ] Jenkins
- [ ] fundamental test suite
- [ ] comprehensive tests (specify *which* if a subset was used)

## DEPENDENCIES:

None

## DOCUMENTATION:
N/A

## CHECKLIST

- [x] My code follows the style guidelines in the Contributor's Guide
- [x] I have performed a self-review of my own code using the Code Reviewer's Guide
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes need updates to the documentation. I have made corresponding changes to the documentation
- [x] My changes do not require updates to the documentation (explain).
- [x] My changes generate no new warnings
- [x] New and existing tests pass with my changes
- [ ] Any dependent changes have been merged and published

## LABELS (optional): 

[...]

PR #799, opened 2023-05-15
[develop] Update WM and UPP hashes and minor rearrangement of WE2E coverage tests that fail on certain platforms

MichaelLueken:
## DESCRIPTION OF CHANGES: 
* The WM and UPP hashes in the `Externals.cfg` file were updated to current versions.
* The `nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15_thompson_mynn_lam3km` test that is constantly failing on Cheyenne GNU and the `grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta` test on Hera Intel have been moved to platforms where the tests successfully run.
* Minor changes to the `DT_ATMOS` for the `RRFS_CONUS_25km` grid were made to allow the WE2E tests run after updating to the current UFS-WM hash.  Please note that there is still an issue with the `grid_RRFS_AK_3km_ics_FV3GFS_lbcs_FV3GFS_suite_HRRR` test on Jet.

### Type of change
- [X] Bug fix (non-breaking change which fixes an issue)

## TESTS CONDUCTED: 
- [X] hera.intel
- [X] hera.gnu
- [X] orion.intel
- [X] cheyenne.intel
- [X] cheyenne.gnu
- [ ] gaea.intel
- [X] jet.intel
- [ ] wcoss2.intel
- [ ] Jenkins
- [X] fundamental (coverage) test suite

## CHECKLIST
- [X] I have performed a self-review of my own code using the Code Reviewer's Guide
- [X] My changes do not require updates to the documentation (explain).
- [X] My changes generate no new warnings
- [X] New and existing tests pass with my changes


danielabdi-noaa (approved):
LGTM

chan-hoo (approved):

mkavulich (changes requested):
@MichaelLueken Sorry I am not very up-to-date on current development due to other obligations: what commit(s) resulted in the failure for the `nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15_thompson_mynn_lam3km ` and `grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta` tests? We shouldn't be allowing existing failures unless there is a very good reason for this. 

This also reminds me of the issue around the known failure of `grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta` (https://github.com/ufs-community/ufs-srweather-app/issues/731). I didn't have time to attempt a fix of this, but I was at least able to point to the change that resulted in this failure. We need to have someone assigned to fix these failures prior to the release, it shouldn't be something we just sweep under the rug.

MichaelLueken:
> @MichaelLueken Sorry I am not very up-to-date on current development due to other obligations: what commit(s) resulted in the failure for the `nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15_thompson_mynn_lam3km ` and `grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta` tests? We shouldn't be allowing existing failures unless there is a very good reason for this.
> 
> This also reminds me of the issue around the known failure of `grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta` (#731). I didn't have time to attempt a fix of this, but I was at least able to point to the change that resulted in this failure. We need to have someone assigned to fix these failures prior to the release, it shouldn't be something we just sweep under the rug.

@mkavulich No worries.  The failures in `nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15_thompson_mynn_lam3km` and `grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta` were first seen for the PRs that came in directly after PR #732 was merged - PRs #745 and #729, respectively.

The `nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15_thompson_mynn_lam3km` test passes on any machine that isn't Cheyenne, so it might be a similar issue to what was encountered in PR #732 with the `grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2`test.  The failure is also very reminiscent - only failing intermittently in the run_post tasks.

The `grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta` test is failing in the run_fcst step, similar to what you noted in issue #731 for the `grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta` test.

Since this PR is fairly open-ended, I can certainly attempt to work through these failures as part of this PR, rather than comment them out in the coverage tests.

For instance, it's possible that the reverting the modifications made to the `model_configure` file will allow both the `grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta` and `grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta` tests to pass.

mkavulich:
Thanks for the reply @MichaelLueken. I think it is okay to re-allocate tests to different machines if they are failing only on one or two (likely machine-specific memory/node problems), but these all-platform run_fcst failures are more serious. I apologize for bringing this up as a roadblock on your PR since you didn't introduce these failures, but I feel like it is very important to avoid setting the precedent of simply kicking the can down the road on test failures. At the very least, any known failure should be documented as a high-priority issue.

I'm curious to hear your thoughts on the best way forward here. I am willing to help investigate and open issues for existing failures if that would be helpful, but I do not have time to try to debug model segfaults and related mysterious errors caused by namelist/configure/openmp changes that I'm not very familiar with. If there is a solid plan with people assigned to solve the problem, I would be okay with temporarily removing these failing tests and merging this PR, but there must be a specific plan to resolve these failures soon, *prior* to the release. Especially since the `grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta` domain is specifically used for the RRFS.

Externals.cfg
```diff
@@ -21,7 +21,7 @@ protocol = git
 repo_url = https://github.com/NOAA-EMC/UPP
 # Specify either a branch name or a hash but not both.
 #branch = develop
-hash = 22cfb88
+hash = dccb321
```
mkavulich, line 24: [`dccb321`](https://github.com/NOAA-EMC/UPP/commit/dccb32176930676ef2a258eb65571ab4e3f7e7a4) is an old hash from early april, the latest is [`1edfb3c`](https://github.com/NOAA-EMC/UPP/commit/1edfb3c71894ed0740f75692bd362a014b9c66aa); is there a reason not to update here?

```suggestion
hash = 1edfb3c
```

MichaelLueken: During the preparation for the SRW v2.1.0 release, I had to drop the version of UPP to mirror the version that the WM used.  I was told that this was to allow for the inline-post to work properly.

The WM at hash [e403bb4](https://github.com/ufs-community/ufs-weather-model/commit/e403bb43ce5f1e0b5a10a37d136cc84a205da7ce) uses the UPP at [dccb321](https://github.com/NOAA-EMC/UPP/commit/dccb32176930676ef2a258eb65571ab4e3f7e7a4).

I don't know if there is any truth to this, but that's the way that I've been updating the WM and UPP hashes since the v2.1.0 release.  I bring the WM hash to the HEAD of develop, then update UPP to the version used by the WM.  If it is safe to use the current HEAD of both repositories, then I don't mind updating them (though I will likely be asked to drop the UPP version in the release branch).

MichaelLueken:
@mkavulich For the two tests that I commented out, I think that the issue with `grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta` might be better described by issue #652, rather than issue #731.  This test is part of the Hera Intel test suite, so it is being run in NCO-mode.  It's also a case that this test doesn't _always_ fail in the coverage tests.  I just wish that this failure wasn't so random and gave enough information to properly debug it.

I'll go ahead and move the `nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15_thompson_mynn_lam3km` test to another machine.

I'll do some testing on the `grid_RRFS_NA_3km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta` test to see if I can't correct the issue here.  If it is a case that the changes from PR #682 need to be reverted, then it shouldn't be an issue, since those changes are in the current `production/AQM.v7` branch.

Externals.cfg
mkavulich: I don't think there's any reason the UPP hashes have to match, but I guess an abundance of caution is called for this close to a release. I'm okay with an older hash with that reason given.

MichaelLueken: Wen usually opens PRs in the WM to bring in the latest UPP changes on a monthly cadence.  Wen has been out for the last three weeks, which is why the UPP in the WM is so far behind.  I will likely update the WM one more time before the release, which should bring in a much more recent version of UPP.

modulefiles/build_gaea_intel.lua
natalie-perlin, line 1: All the regression tests pass on Gaea with the following updated stack:
`/lustre/f2/dev/role.epic/contrib/hpc-stack/intel-classic-2022.0.2/modulefiles/stack/`

Please see comments and logs for the WM regression tests:
https://github.com/ufs-community/ufs-weather-model/issues/1755#issuecomment-1553042387
https://github.com/ufs-community/ufs-weather-model/issues/1755#issuecomment-1553253952

modulefiles/build_gaea_intel.lua
```diff
@@ -7,12 +7,11 @@ whatis([===[Loads libraries needed for building the UFS SRW App on Gaea ]===])
 
 load(pathJoin("cmake", os.getenv("cmake_ver") or "3.20.1"))
 
-prepend_path("MODULEPATH","/lustre/f2/dev/role.epic/contrib/hpc-stack/intel-2021.3.0_noarch/modulefiles/stack")
+prepend_path("MODULEPATH","/lustre/f2/dev/role.epic/contrib/hpc-stack/intel-2022.0.2/modulefiles/stack")
```
natalie-perlin: Suggesting changes following the recent updates and tests:

prepend_path("MODULEPATH","/lustre/f2/dev/role.epic/contrib/hpc-stack/intel-classic-2022.0.2/modulefiles/stack")

modulefiles/build_gaea_intel.lua
```diff
@@ -7,12 +7,11 @@ whatis([===[Loads libraries needed for building the UFS SRW App on Gaea ]===])
...
 load(pathJoin("hpc", os.getenv("hpc_ver") or "1.2.0"))
-load(pathJoin("intel", os.getenv("intel_ver") or "2021.3.0"))
-load(pathJoin("hpc-intel", os.getenv("hpc_intel_ver") or "2021.3.0"))
+load(pathJoin("intel-classic", os.getenv("intel_ver") or "2022.0.2"))
+load(pathJoin("hpc-intel", os.getenv("hpc_intel_ver") or "2022.0.2"))
 load(pathJoin("hpc-cray-mpich", os.getenv("hpc_cray_mpich_ver") or "7.7.11"))
```
natalie-perlin: Suggesting changes following the recent updates for Gaea C3, C4, hpc-stack, and WM regression tests: 
load(pathJoin("hpc-intel-classic", os.getenv("hpc_intel_classic_ver") or "2022.0.2"))
load(pathJoin("hpc-cray-mpich", os.getenv("hpc_cray_mpich_ver") or "7.7.20"))

natalie-perlin (changes requested):
Suggesting changes following the recent updates for Gaea C3, C4, hpc-stack, and WM regression tests, in build_gaea.intel.lua

MichaelLueken:
@natalie-perlin Is there an ETA on when the updated hpc-stack location on Gaea will be merged into the ufs-weather-model repo?  If it is soon, then I'm fine with making the modifications that you suggest.  Otherwise, I would like to move forward with this update that should allow the SRW to build and run again using the same modifications as the weather model.

natalie-perlin:
@MichaelLueken @jkbk2004 
The code management team was notified yesterday on the successful regression tests with the new stack, and I opened a weather-model PR: https://github.com/ufs-community/ufs-weather-model/pull/1758.

mkavulich (approved):
Thanks for addressing my concerns 👍 Let me know if you need any help investigating/testing the known failures.

mkavulich:
@MichaelLueken when I updated my own branch to the latest WM hash, I saw some new failures on Jet (segfaults in the run_fcst step for `GST_release_public_v1`, `grid_RRFS_AK_3km_ics_FV3GFS_lbcs_FV3GFS_suite_HRRR`). I am curious if you have noticed this, and if anyone is looking into it? It may be that the latest weather model hash requires some change to the namelist settings or test config.

MichaelLueken:
@mkavulich Yes, I've noticed that there are several tests across the Tier-1 platforms (I haven't checked Jet yet, but I know there are failures on Cheyenne and Gaea) that are failing with the latest hash of the WM.  I'm working on trying to identify the hash that causes the failures and see if the PR required further changes to be made to allow the regression tests to run properly.

MichaelLueken:
On Cheyenne Intel - the `specify_template_filenames` test is failing in `run_fcst` with the following in the log file:
`FATAL from PE     1: compute_qs: saturation vapor pressure table overflow, nbad=      1`

On Hera Intel - the `grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2` test is failing in `run_fcst` with the following in the log file:
`FATAL from PE     1: compute_qs: saturation vapor pressure table overflow, nbad=      1`

On Hera GNU - the `GST_release_public_v1` test is failing in `run_fcst` with the following in the log file:
`FATAL from PE     2: compute_qs: saturation vapor pressure table overflow, nbad=      1`

The PR that caused the failure is [#1731](https://github.com/ufs-community/ufs-weather-model/pull/1731).  This is a PBL, convection, and microphysics update for HR2.  No changes were made to the WM's RTs in order to make them run.

On Jet - the `grid_RRFS_AK_3km_ics_FV3GFS_lbcs_FV3GFS_suite_HRRR` test is failing in run_fcst with the following in the log file:
`NOTE from PE     0: ==> Error from range_check: data out of bound`

mkavulich:
Thanks for looking into that @MichaelLueken.  It looks like all the failures are due to CFL violations or similar errors; the tweaking of physics must have caused a change in sensitivity that causes the failures we're seeing now (many of our tests push the limit of acceptable timesteps in order to make them cheaper). When I re-ran my failed tests with shorter timesteps the GST_release_public_v1 test (180 --> 150) succeeded. However, the failing grid_RRFS_AK_3km_ics_FV3GFS_lbcs_FV3GFS_suite_HRRR test was already at 10s, and reducing to 8 still resulted in failure at roughly the same time (vertical velocity too large). I wonder if any tests were run on these changes in complex terrain at high resolution before they were committed (@Qingfu-Liu?). The regression tests only contain low resolution domains.

MichaelLueken:
@mkavulich It looks like the failure with the `grid_RRFS_AK_3km_ics_FV3GFS_lbcs_FV3GFS_suite_HRRR` is due to changes from a separate PR in the ufs-weather-model.  I'm currently testing the hash from [#1720](https://github.com/ufs-community/ufs-weather-model/pull/1720), and this test is still failing.  This might be something that we can correct in the SRW, but I need to still find which hash is causing the failure first.

MichaelLueken:
The `modulefiles/build_gaea_intel.lua` modulefile has been reverted to the what is currently in develop.  Once the weather model has merged @natalie-perlin's PR to use the latest hpc-stack location, a PR will be opened in the SRW App to make the same update.  In the interim, Gaea has been deactivated in the `Jenkinsfile` (the SRW App is unable to build without an updated modulefile).  Applied @mkavulich's fix to DT_ATMOS (reducing the value from 180 to 150) for the RRFS_CONUS_25km grid.  Currently, the only test that is failing is the Jet test - `grid_RRFS_AK_3km_ics_FV3GFS_lbcs_FV3GFS_suite_HRRR`.  Will continue to try and find the hash that causes this test to fail and will hopefully be able to correct it at that time.

modulefiles/build_gaea_intel.lua
MichaelLueken: @natalie-perlin I have reverted my modifications to `modulefiles/build_gaea_intel.lua` for now.  Once your PR has been merged in the UFS-WM, we can open a PR in the SRW App to update the modulefile here too.

modulefiles/build_gaea_intel.lua
MichaelLueken: @natalie-perlin I have reverted my modifications to `modulefiles/build_gaea_intel.lua` for now. Once your PR has been merged in the UFS-WM, we can open a PR in the SRW App to update the modulefile here too.

modulefiles/build_gaea_intel.lua
natalie-perlin: Sounds good. I've added a new issue to keep track of the changes to be made for Gaea:
https://github.com/ufs-community/ufs-weather-model/issues/1772

natalie-perlin (approved):

MichaelLueken:
@mkavulich I have gone through all of the preceding hashes and it looks like the issue with the `grid_RRFS_AK_3km_ics_FV3GFS_lbcs_FV3GFS_suite_HRRR` test was brought about by the changes in UFS-WM hash [889254a](https://github.com/ufs-community/ufs-weather-model/commit/889254a63e38b5318cefb6d07a75ea99837f7bf4) and is associated with PR [#1658](https://github.com/ufs-community/ufs-weather-model/pull/1658).  I will now look into what changes might be causing this test to fail.  If you see anything in this PR that looks like the culprit, please let me know.  Thank you very much!

mkavulich:
Thanks @MichaelLueken  for your continued work on this. I have some welcome (if weird) news! I made a run of the `grid_RRFS_AK_3km_ics_FV3GFS_lbcs_FV3GFS_suite_HRRR ` case on Jet with an increased timestep of DT_ATMOS=20 to see if I could get the failure to occur closer to the first output forecast hour (to see if I could directly observe some of the developing numerical instability). And yet when I did so, the case ran to completion! I also tried DT_ATMOS=30 and DT_ATMOS=45 and they both succeeded as well...so I think we can be comfortable increasing the default DT_ATMOS for that domain to either of those values! 30 is probably safer.

My strong suspicion is that the changes in https://github.com/ufs-community/ccpp-physics/pull/43 by Joe Olson are responsible for the failures, since HRRR includes the MYNN-EDMF scheme that was heavily modified in that PR. It is very curious that *increasing* the time step would resolve this failure, I may drop Joe a note to see if this behavior makes any sense to him.

MichaelLueken:
@mkavulich Very welcome news indeed!  Thank you so much for assisting with trying to figure out what was causing issues with the updated WM hash!  I've set the DT_ATMOS to 30 for this grid and have resubmitted the test.  Once it is complete, I will push the change.  Have a great weekend!

MichaelLueken:
@mkavulich The changes made to the DT_ATMOS value to make the `specify_template_filenames`, `grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2`, and `GST_release_public_v1` tests pass using the updated WM hash, decreasing the value from 180 to 150, is now causing the `grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v17_p8_plot` test to fail on Cheyenne Intel.  Values of 160 and 170 were also tested for this case, but ultimately resulted in failure.  Just wanted to let you know why this work hasn't been merged as of this time.

mkavulich:
@MichaelLueken It doesn't make any sense at all that changing the time step would cause plotting to fail (even less sense than increasing time steps to solve a CFL violation; at least that is all related to the dynamical core!). If valid output is produced, then the plotting scripts should be able to plot it. 

Does the test succeed if you manually change DT_ATMOS back to 180 for that plotting test?

MichaelLueken:
> @MichaelLueken It doesn't make any sense at all that changing the time step would cause plotting to fail (even less sense than increasing time steps to solve a CFL violation; at least that is all related to the dynamical core!). If valid output is produced, then the plotting scripts should be able to plot it.
> 
> Does the test succeed if you manually change DT_ATMOS back to 180 for that plotting test?

@mkavulich I completely agree.  Using DT_ATMOS of 150, everything successfully passes, except for the plotting task.  If the rest of the tasks successfully complete, then the plotting script should be able to successfully plot.  However, this isn't the case.  Yes, changing DT_ATMOS back to 180 allows the plotting test to successfully run.  When DT_ATMOS was set to 160 and 170, there were failures in the post tasks.  Further decreasing DT_ATMOS to 140 causes the run_fcst job to fail.  The failure in the plotting script using DT_ATMOS 150 appears to be related to a self-intersection point:

```
ERROR: TopologyException: side location conflict at -97.608377902731846 34.009777791998118
INFO: Self-intersection at or near point -97.878776042255282 34.00918460186481
```

Again, setting DT_ATMOS to 180 solves this failure.  Of course, using this value also causes three tests to fail.

mkavulich:
>Further decreasing DT_ATMOS to 140 causes the run_fcst job to fail. 

*This* is problematic. Similar to the AK domain problem, but in this case it's *not* using the MYNN-EDMF scheme....so your guess is as good as mine.

The hacky solution would be to change back to DT_ATMOS=180 just for the plotting test. But I am worried that we are just masking potential problems here that users will start to run into if they make any modifications. Do you know if there is anyone on the weather model team more familiar with the dynamical core that can offer us any assistance here?

MichaelLueken:
@mkavulich I've reached out to Jong to see if he can recommend someone that can hopefully assist with this issue.

MichaelLueken:
Following the discussion during the SRW CM meeting, I will update the `grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v17_p8_plot` test config file to use DT_ATMOS = 180 and then create an issue on the issue.

MichaelLueken:
The Jenkins test on Cheyenne GNU successfully passed.  Moving forward with merging this PR now.  For more details on the `grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v17_p8_plot` test filaure, please see issue #827.

Issue #652, opened 2023-03-08
NCO mode (`run_envir="nco"`) results in random failures for WE2E tests


mkavulich:
## Expected behavior
The `run_envir` capability was included in the `run_WE2E_tests.sh` script (and its replacement `run_WE2E_tests.py`) in order to be able to force tests to run with either `run_envir=community` or `run_envir=nco`, regardless of what setting was included in the test config yaml file. Calling the WE2E run script with `run_envir=nco` will force all tests to run in NCO mode. Ideally this should not be a problem, as even though the `nco_dirs` directory is shared among the various tests, conflicts *should* be avoided by running each task in its own subdirectory.

## Current behavior
Currently, running several experiments in parallel in nco mode reveals some problems with the system. Tasks seem to fail randomly -- often without a descriptive error message -- and will work upon re-running. 

I confirmed that this behavior is random by running the same set of experiments twice, and seeing a completely different set of failures in each run. Running this same set of tests *without* the `run_envir="nco"` option, or running each task serially so that no two tests were running at the same time, resulted in all successes.

Examples of these failures can be found on Hera in /scratch2/BMC/fv3lam/kavulich/UFS/workdir/nco_tests/expt_dirs

## Machines affected
All that I have tested so far (Hera and Jet). I assume this will affect all platforms.

## Steps To Reproduce
Run the fundamental suite of WE2E tests using either the shell- or python-based run script:

```
./run_WE2E_tests.sh test_type=fundamental machine=hera account=ACCOUNT run_envir=nco
```
or
```
./run_WE2E_tests.py -m=hera -a=ACCOUNT --tests=fundamental --run_envir=nco
```

**Note** Due to the error described in #571, the above sets of tests will not be the same on Hera. The error should still occur regardless (though due to its random/inconsistent nature, it may take a few tries to replicate).

mkavulich:
Update: I was able to replicate this error without explicitly passing the `run_envir=nco` option, simply by specifying 4 nco-mode tests in my test list:

 - nco_grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16
 - nco_grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_timeoffset_suite_GFS_v16
 - nco_grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15_thompson_mynn_lam3km
[...]

Issue #805, opened 2023-05-19
WE2E test grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta fails on Hera Intel in the run_fcst step


MichaelLueken:
## Expected behavior
The WE2E coverage tests should successfully pass on any given machine.

## Current behavior
The `grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta` test is failing in the run_fcst step with the following error message on Hera Intel (NCO-mode test):

`FATAL from PE    25: NetCDF: Index exceeds dimension bound: netcdf_read_data_2d: file:INPUT/gfs_data.nc- variable:ps`

Please note that, if this test is run by itself, it will successfully pass.  Also, this test will pass on other machines.

## Machines affected
Hera Intel WE2E coverage tests

## Steps To Reproduce
1. Clone the current develop branch on Hera
2. Run the coverage.hera.intel.nco suite
3. See that the `grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta` test fails in run_fcst step with the above error message.

MichaelLueken:
This test successfully runs on Hera Intel as part of the comprehensive test suite and also on Hercules as part of the coverage tests for that machine.  Closing obsolete issue.

PR #810, opened 2023-05-23
[develop] Fix retrieve data.

christinaholtNOAA:
## DESCRIPTION OF CHANGES: 

Fix a whitespace issue that leads to pre-mature exit when looking through the possible archive internal directories.

Also addresses an issue uncovered when a user has a pre-existing PYTHONPATH set. We should be prepending the path instead of appending it so that we override any other clones a user might have in their PYTHONPATH.

This fixes an issue with the `nco_grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_timeoffset_suite_GFS_v16` test.

### Type of change

## TESTS CONDUCTED: 

- [x] hera.intel
- [ ] orion.intel
- [ ] cheyenne.intel
- [ ] cheyenne.gnu
- [ ] gaea.intel
- [ ] jet.intel
- [ ] wcoss2.intel
- [ ] Jenkins
- [x] fundamental test suite

Several manual tests were run in conjunction with @MichaelLueken and @mkavulich to isolate the sort of behavior that led to the failure for the NCO test. The Intel fundamental tests were run in a clean clone in a different directory structure than my usual (see below) to ensure that the NCO directories were indeed "fresh".

After the failure was identified a few different tests were run with the user PYTHONPATH set to other clones' ush dirs. This type of failure is addressed by the "prepend" instead of the "append" in the "run tests" script.

## DEPENDENCIES:

n/a

## DOCUMENTATION:

n/a

## ISSUE: 

Related to bug in Issue #652 

The intermittent nature of that failure could be explained by the location of the tester's clone, and how the NCO directories are set. For example, I personally tend to cloning the repo with a different name under the same directory, and have forever. This means that all my tests for all repos forever have had the same NCO directories by default. So, way back when this test wasn't broken, I pulled the data and it has lived there on disk since, so all my fundamental tests pass.

When I clone the repo anywhere else and run the fundamental tests, I observe the data failure associated with not finding the data in the correct internal archive directory structure. The get_data_* test do not complain, and the make_ics and make_lbcs tasks fail with no data.

I believe this is why some folks may have seen intermittent or random failures, as outlined in Issue #652.

## CHECKLIST

## CONTRIBUTORS (optional): 

Thanks @mkavulich @MichaelLueken for running tests to help isolate the problem!

### DIFF
--- a/tests/WE2E/run_WE2E_tests.py
+++ b/tests/WE2E/run_WE2E_tests.py
@@ -8,7 +8,7 @@ import logging
 from textwrap import dedent
 from datetime import datetime
 
-sys.path.append("../../ush")
+sys.path.insert(0, "../../ush")
 
 from generate_FV3LAM_wflow import generate_FV3LAM_wflow
 from python_utils import (
--- a/tests/WE2E/run_WE2E_tests.py
+++ b/tests/WE2E/run_WE2E_tests.py
@@ -8,7 +8,7 @@ import logging
 from textwrap import dedent
 from datetime import datetime
 
-sys.path.insert(0, "../../ush")
+sys.path.insert(1, "../../ush")
 
 from generate_FV3LAM_wflow import generate_FV3LAM_wflow
 from python_utils import (
--- a/ush/retrieve_data.py
+++ b/ush/retrieve_data.py
@@ -609,11 +609,11 @@ def hpss_requested_files(cla, file_names, store_specs, members=-1, ens_group=-1)
             # something has gone wrong.
             unavailable = set.union(*unavailable.values())
         
-        # Break loop if unexpected files were found or if files were found
-        # A successful file found does not equal the expected file list and 
-        # returns an empty set function.
-        if not expected == unavailable:
-            return unavailable - expected
+    # Break loop if unexpected files were found or if files were found
+    # A successful file found does not equal the expected file list and 
+    # returns an empty set function.
+    if not expected == unavailable:
+        return unavailable - expected
     
     # If this loop has completed successfully without returning early, then all files have been found
     return {}


MichaelLueken (approved):
@christinaholtNOAA These changes look good to me!  I was able to run the fundamental tests and the coverage suite on Hera Intel (with the expected failure of `grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta`  from issue #805) without issue.  Approving now.

tests/WE2E/run_WE2E_tests.py
```diff
@@ -8,7 +8,7 @@
 from textwrap import dedent
 from datetime import datetime
 
-sys.path.append("../../ush")
+sys.path.insert(0, "../../ush")
```
mkavulich: Per [this StackOverflow comment](https://stackoverflow.com/a/10097543), it seems like it's better practice to use 1 rather than 0 here (sys.path[0] is the current working directory).

mkavulich (approved):
Thanks for the fix, looks good

MichaelLueken:
@christinaholtNOAA The Jenkins automated tests failed on Gaea (known issue), Hera functional tests, and Orion due to the inability to clone the Externals.  The tests on Jet, Cheyenne Intel, and the expected tests on Cheyenne GNU passed without issue.  The WE2E coverage tests were manually ran on Hera and Cheyenne, where they passed without issue.  Additionally, manually running the functional tests script on Hera showed that the data was successfully found and pulled (I don't know why reruns of the Jenkins tests on Hera still show failures).  Moving forward with merging this update now.

PR #828, opened 2023-06-13
[develop] Bring change of production/AQM.v7 back to develop

### LINKED ISSUE #786: Data structure, error handling, and COM definition should be updated to meet NCO standards
## Description
- Based on the EE2 review for AQM.v7, the structure of the data directories/files and error/warning handling should be updated for the `nco` mode. 
- The utility `compath.py` should be used to define COMIN/COMOUT for the `nco` mode.

chan-hoo:
## DESCRIPTION OF CHANGES: 
Multiple scripts relevant to AQM are updated to meet the NCO standards and EE2 reviewer's comments based on the `production/AQM.v7` branch.
- The NCO variables such as NET, RUN, envir are renamed with the suffix `_default` meaning the default value. This is because they should be defined in `job cards` as can be seen in the NCO standards (pp.4, Table 1). In the current rocoto-based workflow, the job cards do not exist (instead, it uses the XML file). Therefore, they can be set in the configuration file and `ush/machine/` file. However, for ecFlow, these variables should be replaced with those in the job cards. (The ecFlow option will be added in a separate PR later). This is done in the `ush/job_preamble.sh` script which is sourced in the J-job scripts.
- NCO wants to define COMIN/COMOUT, COMINmodel, DATAROOT using the `compath` call available in `prod_util`.
- NCO does not want to add a new directory (COMINext) to COM. The use of COMINext is removed.
- NCO wants to use the error handling calls such as 'err_chk' and 'err_exit' in 'prod_util' which is available only on WCOSS2.
- NCO uses the `YES/NO` flags while SRW App uses `TRUE/FALSE`. To resolve this mismatch, the `boolify` calls for the NCO variables are added to `job_preamble.sh`. 
- The 'pgmerr' function is added to `POST_STEP`.
- The forecast hour in some output file names is changed to 3 digit. 
- The working directory `DATA` is defined and created in `job_preamble.sh` for the NCO mode. Therefore, if-statement for the 'community' mode is added to the J-job scripts.
- The `KEEPDATA` capability does not work correctly for the shared tasks. This is fixed in this PR.
- The input data directories are renamed to `DCOMINdata`.

### Type of change

## TESTS CONDUCTED: 
- WE2E fundamental on Hera
- AQM community sample run on Hera
- AQM nco realtime sample run on WCOSS2

- [x] hera.intel
- [ ] orion.intel
- [ ] cheyenne.intel
- [ ] cheyenne.gnu
- [ ] gaea.intel
- [ ] jet.intel
- [x] wcoss2.intel
- [ ] Jenkins
- [ ] fundamental test suite

## ISSUE: 
Fixes issue mentioned in #786

## CHECKLIST


jobs/JREGIONAL_GET_EXTRN_MDL_FILES
```diff
@@ -30,7 +30,7 @@
 #
 . $USHdir/source_util_funcs.sh
 source_config_for_task "task_get_extrn_ics|task_get_extrn_lbcs" ${GLOBAL_VAR_DEFNS_FP}
-. $USHdir/job_preamble.sh
+. $USHdir/job_preamble.sh "TRUE"
```
christinaholtNOAA, line 33: It doesn't look like `job_preamble.sh` should accept an argument. If it does and I missed it, please add documentation about it in that script.

ush/config_defaults.yaml
```diff
@@ -1004,7 +987,11 @@ nco:
   #-----------------------------------------------------------------------
   #
   # Set variables that are only used in NCO mode (i.e. when RUN_ENVIR is 
-  # set to "nco").  Definitions:
+  # set to "nco"). All variables have the suffix [_dfv] meaning the default value.
+  # This is bacuase they are supposed to be defined in job cards for the
```
MichaelLueken: because rather than bacuase.

```suggestion
  # This is because they are supposed to be defined in job cards for the 
```

MichaelLueken (approved):
@chan-hoo These changes look good to me!  I noted a minor spelling issue in my review, but that is the only issue I noted.  I also ran the fundamental tests on Orion and they all passed successfully.  Approving now.

ush/config_defaults.yaml
chan-hoo: @MichaelLueken, Oops! silly mistake! Thank you for letting me know. Fixed.

chan-hoo:
@MichaelLueken, thanks for your quick review!

BenjaminBlake-NOAA (changes requested):
@chan-hoo  Thank you for putting together this PR!  I have a couple of questions/comments, but otherwise everything looks good to me.

1. Why is the $DATA working directory now defined in job_preamble.sh for NCO mode only?  Doing this adds more logic to each J-job (if RUN_ENVIR = community) so I was just wondering why this change was made.

2. I'm not sure why, but dfv looks strange to me as an abbreviation for default value.  I'd prefer either df or def instead.  I'm interested to see if @christinaholtNOAA has a preference because she had mentioned she was in the process of reviewing this PR too.

chan-hoo:
@BenjaminBlake-NOAA 
1) The working directory ($DATA) has been defined in `job_preamble.sh` for the `nco` mode (I think Daniel designed this structure. I guess he put everything for NCO into `job_preamble.sh`). I didn't change it here. However, $DATA was set again in the J-job script for the `community` mode. From the `nco` perspective, the second is not necessary. To prevent this duplicate, this PR adds if-statement to the second definition in J-job scripts.
2) I agree, but 'def' is used for 'definition'. For example, the directory name of the ecFlow definition file is 'def'. I just want to use three characters. I feel two characters look too short. I'd like to hear what you think :)

BenjaminBlake-NOAA:
@chan-hoo 
1. Understood!  Yes NCO won't want duplicates of the directory definition.

2. Agreed about "def" which could be confused with definition, and I'm fine with something longer than 2 characters.  What about "dflt"?

chan-hoo:
@BenjaminBlake-NOAA, I googled it. 'DV' is used for 'default value' in computing.

BenjaminBlake-NOAA (approved):
@chan-hoo  We could go with dv, but I also agree with you that two letters seems too short.  Maybe it'd be best to spell out default to eliminate any possible confusion?  Submitting my approval now regardless of what you decide to do :)

RatkoVasic-NOAA:
@chan-hoo @BenjaminBlake-NOAA 
From default value:
DVL, DVE, DFV?

jobs/JREGIONAL_GET_EXTRN_MDL_FILES
```diff
@@ -245,7 +244,7 @@ Call to ex-script corresponding to J-job \"${scrfunc_fn}\" failed."
 #
 #-----------------------------------------------------------------------
 #
-job_postamble
+job_postamble "FALSE"
```
christinaholtNOAA, line 247: Similar comment as above. I see that this function does accept TRUE as an action here, but this seems to be an undefined keyword that could be literally anything.

ush/config_defaults.yaml
```diff
@@ -1042,43 +1029,34 @@ nco:
...
   #-----------------------------------------------------------------------
   #
-  envir: "para"
-  NET: "rrfs"
-  RUN: "rrfs"
-  model_ver: "v1.0.0"
+  envir_dfv: "para"
+  NET_dfv: "rrfs"
+  RUN_dfv: "rrfs"
+  model_ver_dfv: "v1.0.0"
```
christinaholtNOAA: I'm confused about the need for this change of variable name in the first place. I only briefly saw traffic about the naming convention used.

A given variable's name should not be connected to its value. If a given variable has a default value, then we can use it. I don't think it makes sense to decouple the default value from the variable it's meant to be a default for. That adds a level of misdirection and (perhaps accidental) complexity that I don't see a clear, concise use for here.

ush/job_preamble.sh
```diff
@@ -12,6 +12,77 @@ if [ $# -ne 0 ]; then
...
 fi
+
+#
+#-----------------------------------------------------------------------
+#
+# Set NCO standard environment variables
+#
+#-----------------------------------------------------------------------
+#
+export envir="${envir:-${envir_dfv}}"
```
christinaholtNOAA: I think here is where my previous comment comes in. Where else could `envir` have come from if not the `var_defns.sh` file? If that is something that is happening, we really should aim to resolve that duplication of information in the design of how the NCO mode is configured.

scripts/exregional_aqm_lbcs.sh
```diff
@@ -214,11 +226,17 @@ Please ensure that you've built this executable."
 #----------------------------------------------------------------------
 #
   PREP_STEP
-  eval ${RUN_CMD_AQMLBC} ${exec_fp} ${REDIRECT_OUT_ERR} || \
-    print_err_msg_exit "\
-Call to executable (exec_fp) to generate chemical and GEFS LBCs
-file for RRFS-CMAQ failed:
+  eval ${RUN_CMD_AQMLBC} ${exec_fp} ${REDIRECT_OUT_ERR}
+  export err=$?
+  if [ "${RUN_ENVIR}" = "nco" ] && [ "${MACHINE}" = "WCOSS2" ]; then
```
christinaholtNOAA, line 231: Machine dependent logic really should not go into run scripts like this. It would be great if we could make err_chk work for all cases, if that's desirable as a way to reliably check for errors.

scripts/exregional_bias_correction_o3.sh
```diff
@@ -125,15 +113,50 @@ if [ -d "${DATA}/data/bcdata.${yyyymm}" ]; then
   cp_vrfy -rL "${AQM_AIRNOW_HIST_DIR}/bcdata.${yyyymm}/interpolated" "${DATA}/data/bcdata.${yyyymm}" 
 fi
 
-# Retrieve real-time airnow data for the last three days
+# Retrieve real-time airnow data for the last three days and convert them into netcdf
 if [ "${DO_REAL_TIME}" = "TRUE" ]; then
-  mkdir -p ${DATA}/data/bcdata.${yyyymm_m1}/airnow/${yyyy_m1}/${PDYm1}/b008
-  mkdir -p ${DATA}/data/bcdata.${yyyymm_m2}/airnow/${yyyy_m2}/${PDYm2}/b008
-  mkdir -p ${DATA}/data/bcdata.${yyyymm_m3}/airnow/${yyyy_m3}/${PDYm3}/b008
+  for ipdym in {1..3}; do
```
christinaholtNOAA: The variable name ipdym is not very descriptive. Would you mind spelling out what this variable is looping over?

scripts/exregional_bias_correction_o3.sh
```diff
@@ -188,14 +219,23 @@ if [ "${DO_AQM_SAVE_AIRNOW_HIST}" = "TRUE" ]; then
...
+  cp_vrfy ${DCOMINairnow}/${PDYm3}/airnow/HourlyAQObs_${PDYm3}*.dat ${AQM_AIRNOW_HIST_DIR}/bcdata.${yyyymm_m3}/airnow/csv/${yyyy_m3}/${PDYm3}
+
+  # NetCDF files
+  mkdir_vrfy -p ${AQM_AIRNOW_HIST_DIR}/bcdata.${yyyymm}/airnow/netcdf/${yyyy}/${PDY}
+  mkdir_vrfy -p ${AQM_AIRNOW_HIST_DIR}/bcdata.${yyyymm_m1}/airnow/netcdf/${yyyy_m1}/${PDYm1}
+  mkdir_vrfy -p ${AQM_AIRNOW_HIST_DIR}/bcdata.${yyyymm_m2}/airnow/netcdf/${yyyy_m2}/${PDYm2}
+  mkdir_vrfy -p ${AQM_AIRNOW_HIST_DIR}/bcdata.${yyyymm_m3}/airnow/netcdf/${yyyy_m3}/${PDYm3}
+  cp_vrfy ${DATA}/data/bcdata.${yyyymm_m1}/airnow/netcdf/${yyyy_m1}/${PDYm1}/HourlyAQObs.${PDYm1}.nc ${AQM_AIRNOW_HIST_DIR}/bcdata.${yyyymm_m1}/airnow/netcdf/${yyyy_m1}/${PDYm1}  
+  cp_vrfy ${DATA}/data/bcdata.${yyyymm_m2}/airnow/netcdf/${yyyy_m2}/${PDYm2}/HourlyAQObs.${PDYm2}.nc ${AQM_AIRNOW_HIST_DIR}/bcdata.${yyyymm_m2}/airnow/netcdf/${yyyy_m2}/${PDYm2}  
+  cp_vrfy ${DATA}/data/bcdata.${yyyymm_m3}/airnow/netcdf/${yyyy_m3}/${PDYm3}/HourlyAQObs.${PDYm3}.nc ${AQM_AIRNOW_HIST_DIR}/bcdata.${yyyymm_m3}/airnow/netcdf/${yyyy_m3}/${PDYm3}
```
christinaholtNOAA: This is a lot of duplicate information that is, in general, difficult to support over time. Could it be condensed into a loop over the cycles?

scripts/exregional_bias_correction_pm25.sh
```diff
@@ -174,48 +197,59 @@ mkdir_vrfy -p ${DATA}/data/site-lists.interp
 mkdir_vrfy -p ${DATA}/out/pm25/${yyyy}
 mkdir_vrfy -p ${DATA}/data/bcdata.${yyyymm}/interpolated/pm25/${yyyy}
 
-cp_vrfy ${PARMaqm_utils}/bias_correction/sites.valid.pm25.20220724.12z.list ${DATA}/data/site-lists.interp
+cp_vrfy ${PARMaqm_utils}/bias_correction/sites.valid.pm25.20230331.12z.list ${DATA}/data/site-lists.interp
```
christinaholtNOAA, line 192: Should this date string be a parameter?

scripts/exregional_bias_correction_pm25.sh
```diff
@@ -348,7 +396,7 @@ wgrib2 tmpfile_pm25_bc -set_grib_type c3b -new_grid_winds earth -new_grid ${grid
 
 cp_vrfy tmpfile_pm25_bc ${COMOUT}/${NET}.${cycle}.ave_1hr_pm25_bc.${id_domain}.grib2
 cp_vrfy ${NET}.${cycle}.grib2_pm25_bc.227 ${COMOUT}/${NET}.${cycle}.ave_1hr_pm25_bc.227.grib2
-if [ "${SENDDBN}" = "YES" ]; then
+if [ "${SENDDBN}" = "TRUE" ]; then
```
christinaholtNOAA, line 391: These changes to the standard boolean value are great! :)

christinaholtNOAA (changes requested):
@chan-hoo This is a huge effort, and a great step toward getting develop NCO compliant. 

In most the jobs/ scripts mods, I'm seeing that the mkdir command for the working directory has been removed, which means it should have been relegated to a different part of the workflow. Which level is now responsible for that in NCO mode? I thought NCO requirements mandated responsibility for that task to the j-job script. It would be helpful if that responsibility is in the same place for both NCO and community mode.

In most of the scripts, error handling has become platform dependent. That presents a problem for both supportability and transportability. 

Let's talk if it doesn't make sense what I'm trying to say here writing.

jobs/JREGIONAL_GET_EXTRN_MDL_FILES
chan-hoo: @christinaholtNOAA, it accepts an argument for the shared job id. If an argument exists, a shared pid is used.
```
if [ $# -ne 0 ]; then
    export pid=$share_pid
    export jobid=${job}.${pid}
fi
```
I'll add a comment to the script.

jobs/JREGIONAL_GET_EXTRN_MDL_FILES
chan-hoo: I add a new function to `job_postamble` for the shared tasks. When an argument exists, the working directory will not be removed even when KEEPDATA=false. Only when an argument is `true`, all existing working directories will be removed from the `tmp` directory.

ush/config_defaults.yaml
chan-hoo: If we don't use `ecFlow`, you are correct because Rocoto does not use job cards and these NCO variables don't have to be replaced with the new values in the job cards. However, as you can see in Table 1 (pp.4) of the NCO standards, the NCO variables should be defined in the job cards. When these NCO variables are defined in the job cards, they should replace these variables ({var}_dfv) defined in config.yaml and ush/machine/.

ush/job_preamble.sh
chan-hoo: As explained above, `ecFlow` uses job cards to submit tasks (same as what rocoto XML file does). The NCO variables should be defined in the job cards. As a rocoto user, you think that this is a duplicate. However, from the `ecFlow` perspective, they don't have be defined in the configuration file (in var_defns.sh). The reason why I renamed them with "_dfv" is to use both Rocoto and ecFlow.

scripts/exregional_aqm_lbcs.sh
chan-hoo: I understand what your concern is. The problem is that these functions such as 'err_exit', 'err_chk', 'compath' are included in the module `prod_util`. However, this module is only available on WCOSS2 at this moment. Matt is trying to make it work on other platforms.

scripts/exregional_bias_correction_pm25.sh
chan-hoo: No, it is not. As far as I know, it is just a base file name.

scripts/exregional_bias_correction_o3.sh
chan-hoo: It is very straightforward. PDYm1, PDYm2, PDYm3 are the NCO variables standing for the present day (PDY) minus (m) 1, 2, 3 (1, 2, 3 days before the present day). `i` means integer. I've changed it to `i_pdym` to be clear a little bit more.

scripts/exregional_bias_correction_o3.sh
chan-hoo: Modified.

chan-hoo:
@BenjaminBlake-NOAA, I've changed the suffix `_dfv` to `_default`.

chan-hoo:
@christinaholtNOAA, it (mkdir for a working directory) is not removed. For the `nco` mode, the working directory is defined and created in the `job_preamble.sh` script. This script is sourced in J-job scripts. Therefore, a working directory is created in J-job scripts for both `community` and `nco` modes. However, in the current J-job scripts, after the `job_preamble.sh` script is sourced, the `mkdir` command runs again. This means a working directory is created twice for the `nco` mode. In this PR, I add the if-statement to avoid this duplicate.

ush/config_defaults.yaml
christinaholtNOAA: Can that requirement not be interpreted as sourcing the `var_defns.sh` file in the job card? It doesn't explicitly state that the variables are written out there and set. 

I also am worried that this requirement leads to unnecessary duplication of information and general misdirection. For example, where do we point users for setting these values? Typically it's been a one-stop shop for controlling this information in the config_defaults.sh. To override those defaults, putting the information in the config.yaml is the one way to specify a different value.

Even if we do need to adopt this approach for this handful of variables, we need to be consistent with defining them in Rocoto and in ecFlow as environment variables.

It's unclear on whether the requirement here is meant to reflect that _environment_ has these variables set in the job card or if the job card itself is responsible for setting the single source of information to be used for all other scripts. Can you speak to that distinction?

scripts/exregional_aqm_lbcs.sh
christinaholtNOAA: @chan-hoo would you or @MatthewPyle-NOAA be willing to open an issue in SRW (if it doesn't already exist) to reflect that this a bit of logic that might be nice to clean up once `prod_util` is available everywhere?

scripts/exregional_bias_correction_o3.sh
christinaholtNOAA: A comment in the code to just describe that would be helpful for users who do not know the pdy lingo.

scripts/exregional_bias_correction_pm25.sh
christinaholtNOAA: What I mean is that it changes over time. Is this the only place in the code where this changes?

christinaholtNOAA:
For the directory creation, the NCO standards doc seems fairly clear that the j-job script is responsible for directory creation. Delegating that to another script seems antithetical to the standard (pp 3 in v11.0.0)

> The purpose of the J-job is fourfold: to set up location (application/data directory) variables, to set up temporal (date/cycle) variables, to initialize the data and working directories, and to call the ex-script.

chan-hoo:
@christinaholtNOAA, All working directories are created in the J-job scripts for both `nco` and `community`. The difference is that a working directory is created inside the `job_preamble.sh` for the `nco` mode while it is created in the J-job script for the `community` mode (outside the `job_preamble.sh` script). The `job_preamble.sh` script is sourced inside the J-job script. I just remove a duplicate in this PR. So this change does not violate the NCO standards.

christinaholtNOAA:
Okay, so why does sourcing the appropriate `var_defns` section in the job card not work for setting the variables?

chan-hoo:
I think I didn't explain it well. The job cards also read `var_defns`. However, NCO wants to set (modify) the NCO variables in the head/environment files separately for the ecFlow (head.h/env-p1.h) which are called on the top of the job cards. Then the `var_defns` can be sourced in the next line of the job cards. Therefore, the NCO variables defined in `var_defns` should be replaced with those in the `head.h/env-p1.h` files in the job cards for ecFlow.

scripts/exregional_bias_correction_pm25.sh
chan-hoo: This will not be changed case by case. This is a kind of fix file.

scripts/exregional_bias_correction_o3.sh
chan-hoo: I'll add a comment there.

scripts/exregional_aqm_lbcs.sh
chan-hoo: I'll open an issue.

christinaholtNOAA:
@chan-hoo Can you share more about the use case of how NCO folks will want to modify these variables? Do they do it because of a change in the system when the operational model is running? Some other time? Is it not constant through the entirety of the lifecycle of an operational deployment?

What benefit do they have by changing it in the head/environment files instead of var_defns? A single source of information seems like a safer way to manage this information.

scripts/exregional_aqm_lbcs.sh
chan-hoo: Opened: https://github.com/ufs-community/ufs-srweather-app/issues/839

scripts/exregional_bias_correction_o3.sh
chan-hoo: Added.

chan-hoo:
@christinaholtNOAA, their tradition (???). I think it is how they maintain a variety of the applications. They want to set and modify the NCO variables in `ecf/include/head.h` and `ecf/include/env-p1.h` files. The header file `head.h` can be modified only by NCO itself. Even EMC is not allowed it. The developers can add some modification to the environment file `env-p1.h`. I don't have much experience. @MatthewPyle-NOAA, can you explain this in more detail for @christinaholtNOAA to understand?

chan-hoo:
@christinaholtNOAA, Matt is on vacation. He'll come back next Wednesday. Regarding your last two sentences `What benefit do they have by changing it in the head/environment files instead of var_defns? A single source of information seems like a safer way to manage this information.`, NCO has not used `var_defns` (separate file for global environment variables) for their applications. When an EMC folk who has much experience in NCO production applications took a look at production/AQMv.7, his first question was 'why does AQM.v7 store all environment variables in `var_defns` and source it from all scripts?'. In the UFS SRW App, we create a workflow experiment (EXPTDIR) case by case using the python generation script. However, I was told that NCO doesn't use this approach. They adjust all environment variables (cycle, date, nco variables) directly in the ecflow scripts (job cards). The UFS SRW App is a totally new system to NCO and it is totally based on Rocoto. From the rocoto perspective, using a single `var_defn` is safe and efficient. However, from the ecFlow perspective, it is unsafe and inefficient to store all environment variables in a separate file and to modify the file as needed. I think you're asking NCO to change their entire systems only for the UFS SRW App. :) :)

christopherwharrop-noaa:
@chan-hoo - What your narrative about NCO's perspective brings to light is the fact that there is a design conflict between the extremely narrow interests of the Operational environment and what is needed for the community to do research and development.  NCO's approach is not portable or usable outside of operations.  And so, their requirements are causing design problems for the system because now two methods must be accommodated.  This is bad software design.  I don't know that there is a great way out of this problem, but it is worth noting that sometimes Operational requirements appear to dictate approaches that are clearly inferior from the larger perspective of having a single system usable for both research and operations.

chan-hoo:
@christopherwharrop-noaa, as you may know, I don't have any power to change NCO's standards :)  One of the EMC's main roles is to support NCO to run applications operationally. In order to do that, we (EMC) should follow the NCO standards and NCO's approach. Without satisfy them, NCO will not accept any new applications. You can think it is a bad software design. However, to resolve this, NCO should agree with that. It is very hard (especially from my side. I have never met NCO guys since I joined EMC :) ). Since the UFS SRW App is designed to support operation, I think we should give full consideration about this. As you know, they (NCO) are very conservative.

christopherwharrop-noaa:
@chan-hoo - I am in no way suggesting we try to change or ignore NCO standards here.  That is not possible. Everything you said is true.  My point is that this is an ongoing problem that creates conflict between the interests of Operations and the community.  Requiring an implementation that can only work in one place/system is objectively bad software design.  I don't think there is any serious argument to be had on that point. But, we're stuck with it. A solution to this problem will require engagement from up the management chain, and that is outside the scope/purview of this work.

chan-hoo:
@christinaholtNOAA, do you still have any concerns? To add the ecFlow option to develop, I hope this pr is merged as early as possible. If you have concerns relevant to ecFlow, I think we can discuss them in the PR for ecFlow later.

chan-hoo:
@christinaholtNOAA, below is the link to my technical document. On Page 102, you can find how the NCO environment variables are updated through the workflow with ecFlow. I hope this helps you understand why `[nco var]_default` is necessary:
https://drive.google.com/file/d/1zyCvHmuL_LWx0Yto6B1GXF04m-n6Kbs2/view?usp=drive_link

christinaholtNOAA (approved):
@chan-hoo Thanks for addressing my comments. I'm sorry I haven't been available to respond for several days.

MichaelLueken:
@chan-hoo All of the automated Jenkins tests failed.  The issue is with `tests/WE2E/run_WE2E_tests.py`.  On line 161, there is the following:

`test_cfg['nco'].update({"OPSROOT": args.opsroot})`

Since `OPSROOT` has been changed to `OPSROOT_default` in this PR, the automated testing is failing because `OPSROOT` no longer exists in config_defaults.  Please change this `OPSROOT` to `OPSROOT_default`, then I can resubmit the Jenkins tests.

scripts/exregional_make_ics.sh
```diff
@@ -84,7 +84,7 @@ fi
 #-----------------------------------------------------------------------
 #
 if [ $RUN_ENVIR = "nco" ]; then
-    extrn_mdl_staging_dir="${COMINext}${SLASH_ENSMEM_SUBDIR}"
+    extrn_mdl_staging_dir="${DATAROOT}/get_extrn_ics.${share_pid}"
```
MichaelLueken: An ensemble WE2E test that is run in NCO mode is currently failing on Hera.  The issue is that the data is being placed in `${DATAROOT}/get_extrn_ics.${share_pid}${SLASH_ENSEM_SUBDIR}` and not `${DATAROOT}/get_extrn_ics.${share_pid}`.  It looks like we will need to add back in the `${SLASH_ENSMEM_SUBDIR}` to the path for those cases with ensembles.

scripts/exregional_make_lbcs.sh
```diff
@@ -82,7 +82,7 @@ fi
 #-----------------------------------------------------------------------
 #
 if [ $RUN_ENVIR = "nco" ]; then
-    extrn_mdl_staging_dir="${COMINext}${SLASH_ENSMEM_SUBDIR}"
+    extrn_mdl_staging_dir="${DATAROOT}/get_extrn_lbcs.${share_pid}"
```
MichaelLueken: An ensemble WE2E test that is run in NCO mode is currently failing on Hera.  The issue is that the data is being placed in `${DATAROOT}/get_extrn_lbcs.${share_pid}${SLASH_ENSEM_SUBDIR}` and not `${DATAROOT}/get_extrn_lbcs.${share_pid}`.  It looks like we will need to add back in the `${SLASH_ENSMEM_SUBDIR}` to the path for those cases with ensembles.

MichaelLueken (commented):
@chan-hoo The Jenkins tests are now passing for Cheyenne, Jet, and Hera GNU.  However, there is a WE2E test that is still failing on Hera Intel.  The test that is failing is:

`get_from_HPSS_ics_GDAS_lbcs_GDAS_fmt_netcdf_2022040400_ensemble_2mems`

All coverage tests that are run on Hera Intel are run in NCO mode.  The failure is in the make_ics/lbcs tasks:

`/scratch1/NCEPDEV/stmp2/role.epic/jenkins/workspace/fs-srweather-app_pipeline_PR-828__2/scripts/exregional_make_ics.sh: line 93: /scratch1/NCEPDEV/stmp2/role.epic/jenkins/workspace/fs-srweather-app_pipeline_PR-828__2/nco_dirs/tmp/get_extrn_ics.id_1687963892_2022040400/rrfs.t00z.GDAS.ICS.extrn_mdl_var_defns.sh: No such file or directory`

The ICs and LBCs are found in:

`/scratch1/NCEPDEV/stmp2/role.epic/jenkins/workspace/fs-srweather-app_pipeline_PR-828__2/nco_dirs/tmp/get_extrn_ics.id_1687963892_2022040400/mem*`

The location on Hera with the failed tests is:

`/scratch1/NCEPDEV/stmp2/role.epic/jenkins/workspace/fs-srweather-app_pipeline_PR-828__2/expt_dirs/get_from_HPSS_ics_GDAS_lbcs_GDAS_fmt_netcdf_2022040400_ensemble_2mems`

In order to make this test work, I had to add `${SLASH_ENSMEM_SUBDIR}` back into `exregional_make_ics.sh` and `exregional_make_lbcs.sh`.  I tried adding this variable to `job_preamble.sh`, but the jobs still failed with the above error messages.

Is it okay to add ${SLASH_ENSMEM_SUBDIR} back in for ensembles, or would that go against NCO standards?  If it goes against NCO standards to reintroduce this variable in these two scripts, can you think of another way to ensure that ensembles can properly run in NCO mode?  Thanks!

chan-hoo:
@MichaelLueken, thanks for testing. It should be fine. I've added SLASH_ENSEM_SUBDIR to the paths. Please run the tests again.

scripts/exregional_make_ics.sh
chan-hoo: Added.

scripts/exregional_make_lbcs.sh
chan-hoo: Added.

MichaelLueken:
Thanks, @chan-hoo!  I have resubmitted the automated Jenkins testing.

scripts/exregional_make_ics.sh
```diff
@@ -84,7 +84,7 @@ fi
 #-----------------------------------------------------------------------
 #
 if [ $RUN_ENVIR = "nco" ]; then
-    extrn_mdl_staging_dir="${COMINext}${SLASH_ENSMEM_SUBDIR}"
+    extrn_mdl_staging_dir="${DATAROOT}/get_extrn_ics.${share_pid}${SLASH_ENSEM_SUBDIR}"
```
MichaelLueken: The modification should have been `${SLASH_ENSMEM_SUBDIR}`, not `${SLASH_ENSEM_SUBDIR}`.  I'm very sorry about that.

```suggestion
    extrn_mdl_staging_dir="${DATAROOT}/get_extrn_ics.${share_pid}${SLASH_ENSMEM_SUBDIR}"
```

scripts/exregional_make_lbcs.sh
```diff
@@ -82,7 +82,7 @@ fi
 #-----------------------------------------------------------------------
 #
 if [ $RUN_ENVIR = "nco" ]; then
-    extrn_mdl_staging_dir="${COMINext}${SLASH_ENSMEM_SUBDIR}"
+    extrn_mdl_staging_dir="${DATAROOT}/get_extrn_lbcs.${share_pid}${SLASH_ENSEM_SUBDIR}"
```
MichaelLueken: The modification should have been `${SLASH_ENSMEM_SUBDIR}`, not `${SLASH_ENSEM_SUBDIR}`.  I'm very sorry about that.

```suggestion
    extrn_mdl_staging_dir="${DATAROOT}/get_extrn_lbcs.${share_pid}${SLASH_ENSMEM_SUBDIR}"
```

MichaelLueken (commented):
@chan-hoo The same test failed, but due to:

`/scratch1/NCEPDEV/stmp2/role.epic/jenkins/workspace/fs-srweather-app_pipeline_PR-828__2/scripts/exregional_make_ics.sh: line 87: SLASH_ENSEM_SUBDIR: unbound variable`

The added variable should have been `${SLASH_ENSMEM_SUBDIR}`, rather than `${SLASH_ENSEM_SUBDIR}`.  I'm very sorry about this.

I'll resubmit the tests one last time after this has been updated, then move forward with the merge.

scripts/exregional_make_ics.sh
chan-hoo: Updated!

scripts/exregional_make_lbcs.sh
chan-hoo: Updated!

chan-hoo:
@MichaelLueken, I am sorry. I didn't double-check it. I've updated it. Thanks again for the testing.

MichaelLueken:
Now that Orion is back from yesterday's maintenance, the WE2E tests were manually ran and all passed successfully:

```
----------------------------------------------------------------------------------------------------
Experiment name                                                  | Status    | Core hours used 
----------------------------------------------------------------------------------------------------
deactivate_tasks                                                   COMPLETE               1.04
get_from_AWS_ics_GEFS_lbcs_GEFS_fmt_grib2_2022040400_ensemble_2me  COMPLETE             751.05
grid_CONUS_3km_GFDLgrid_ics_FV3GFS_lbcs_FV3GFS_suite_RRFS_v1beta   COMPLETE             252.27
grid_RRFS_AK_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16_plot        COMPLETE             141.12
grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_RRFS_v1beta            COMPLETE              14.19
grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_2017_gfdlmp  COMPLETE              10.21
grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_HRRR              COMPLETE             380.56
grid_RRFS_CONUScompact_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16   COMPLETE              28.19
grid_RRFS_CONUScompact_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16    COMPLETE             279.65
grid_SUBCONUS_Ind_3km_ics_FV3GFS_lbcs_FV3GFS_suite_WoFS_v0         COMPLETE              13.12
nco                                                                COMPLETE               7.70
2020_CAD                                                           COMPLETE              31.47
----------------------------------------------------------------------------------------------------
Total                                                              COMPLETE            1910.57
```

Awaiting completion of automated Jenkins testing before moving forward with the merge.

MichaelLueken:
@chan-hoo The automated tests on Cheyenne, Jet, and Hera Intel successfully passed.  The `get_from_NOMADS_ics_FV3GFS_lbcs_FV3GFS` failed on Hera GNU.  The `get_extrn_ics` task failed.  Using `rocotorewind/rocotoboot`, the test successfully passed.  I will now move forward with merging this work.

```
[Michael.Lueken@hfe09 get_from_NOMADS_ics_FV3GFS_lbcs_FV3GFS]$ rocotostat -w FV3LAM_wflow.xml -d FV3LAM_wflow.db -v 10 | more
       CYCLE                    TASK                       JOBID               STATE         EXIT STATUS     TRIES      DURATION
================================================================================================================================
202306270000               make_grid                    46515760           SUCCEEDED                   0         1           9.0
202306270000               make_orog                    46515788           SUCCEEDED                   0         1          58.0
202306270000          make_sfc_climo                    46515993           SUCCEEDED                   0         1          74.0
202306270000           get_extrn_ics                    46516729           SUCCEEDED                   0         1         290.0
202306270000          get_extrn_lbcs                    46516768           SUCCEEDED                   0         1         483.0
202306270000         make_ics_mem000                    46517185           SUCCEEDED                   0         1         660.0
202306270000        make_lbcs_mem000                    46517186           SUCCEEDED                   0         1        1214.0
202306270000         run_fcst_mem000                    46517792           SUCCEEDED                   0         1        1458.0
202306270000    run_post_mem000_f000                    46518001           SUCCEEDED                   0         1          13.0
202306270000    run_post_mem000_f001                    46518233           SUCCEEDED                   0         1          12.0
202306270000    run_post_mem000_f002                    46518234           SUCCEEDED                   0         1          13.0
202306270000    run_post_mem000_f003                    46518280           SUCCEEDED                   0         1          13.0
202306270000    run_post_mem000_f004                    46518448           SUCCEEDED                   0         1          13.0
202306270000    run_post_mem000_f005                    46518449           SUCCEEDED                   0         1          14.0
202306270000    run_post_mem000_f006                    46518450           SUCCEEDED                   0         1          13.0
```

PR #938, opened 2023-10-10
[develop] Build conda and environments in SRW.

christinaholtNOAA:
## DESCRIPTION OF CHANGES: 

Modifies `devbuild.sh` to add the option to install miniforge (a version of miniconda that manages channels more strictly) in a user's specified location and defaults to inside the user's clone. It also installs two environments needed for SRW -- `srw_app`, which is similar to the old `workflow_tools` environment, and `srw_graphics`, which is sufficient to support the plotting scripts in SRW. 

A few additional details:

* Does the conda installation right away so that MacOS users have the necessary bash utilities like readlink to use the devbuild.sh script.
* Requires the user to provide the `conda` target to build conda -- it doesn't build by default.
* Adds a conda module file that points to the user-installed location of miniconda.
* Modifies the GitHub Actions workflows to use the same environments that are built by devbuild.sh.

### Type of change

## TESTS CONDUCTED: 

- [x] hera.intel
- [ ] orion.intel
- [ ] hercules.intel
- [ ] cheyenne.intel
- [ ] cheyenne.gnu
- [ ] derecho.intel
- [ ] gaea.intel
- [ ] gaeac5.intel
- [ ] jet.intel
- [ ] wcoss2.intel
- [ ] Jenkins
- [x] fundamental test suite

~~Test suite is still pending on Hera.~~ Edit: The fundamental test suite has passed on Hera.

I also tested the conda installation bits on MacOS, but did not carry out an entire build. I did confirm that the environments were installed and could support the readlink utility.

This needs tests on all platforms, most likely.

## DEPENDENCIES:

None.

## DOCUMENTATION:

I update the docs with this PR.

## ISSUE: 

N/A

## CHECKLIST


devbuild.sh
```diff
@@ -187,11 +188,48 @@ while :; do
...
+  fi
+  if ! conda env list | grep -q "^srw_graphics\s" ; then
+    mamba env create -n srw_graphics --file graphics_environment.yml
+  fi
+else
+  source ${CONDA_BUILD_DIR}/etc/profile.d/conda.sh
+  conda activate
+fi
+
+# Conda environment should have linux utilties to perform these tasks on macos.
```
MichaelLueken: utilities, instead of utilties.

```suggestion
# Conda environment should have linux utilities to perform these tasks on macos.
```

docs/UsersGuide/source/BuildingRunningTesting/RunSRW.rst
```diff
@@ -111,23 +111,37 @@ The first two steps depend on the platform being used and are described here for
...
 
-.. attention:: 
+In addition to building the SRW App components, the ``devbuild.sh`` script will install miniconda,
+specifically miniforge, and install the SRW-required environments if the ``conda`` target is
+provided. For example, the following command will build all the default SRW components on the hera
+platform using the Intel compiler, and will also install conda and environments:
+
+```
+./devbuild.sh --platform=hera --compiler=intel default conda
+```
```
gspetro-NOAA:
```suggestion
.. code-block:: console

   ./devbuild.sh --platform=hera --compiler=intel default conda
```

docs/UsersGuide/source/Reference/FAQ.rst
```diff
@@ -234,7 +234,7 @@ To run a new experiment at a later time, users need to rerun the commands in :nu
    module use /path/to/modulefiles
    module load wflow_<platform>
 
-Follow any instructions output by the console (e.g., ``conda activate workflow_tools``). 
+Follow any instructions output by the console (e.g., ``conda activate srw_app``).
```
gspetro-NOAA, line 237: In line 229, `*workflow_tools*` should be changed to `*srw_app*`.

devbuild.sh
```diff
@@ -187,11 +188,48 @@ while :; do
...
+SRW_DIR=$(cd "$(dirname "$(readlink -f -n "${BASH_SOURCE[0]}" )" )" && pwd -P)
+MACHINE_SETUP=${SRW_DIR}/src/UFS_UTILS/sorc/machine-setup.sh
+BUILD_DIR="${SRW_DIR}/build"
+INSTALL_DIR=${SRW_DIR}
```
MichaelLueken: Is there a reason that these two default settings had to be moved for MacOS?  In this position, the default values for `BUILD_DIR` and `INSTALL_DIR` are being used, rather than what is being passed via `tests/build.sh`.  This is causing the `test/build.sh` script to fail to find the executables, since `install_$COMPILER/exec` doesn't exist.

gspetro-NOAA:
@christinaholtNOAA It occurs to me that you probably need to add some documentation to BuildSRW that mentions the `conda` (and `--conda-dir`) argument if that's what we prefer users to use from here on.

.github/workflows/python_tests.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - 'release/*'
+  pull_request:
+    branches:
+      - develop
+      - 'release/*'
+  workflow_dispatch:
+
+defaults:
+  run:
+    shell: bash -leo pipefail {0}
```
gspetro-NOAA, line 15: Curious what the "l" in -leo is for?

.github/workflows/python_tests.yaml
```diff
@@ -0,0 +1,58 @@
...
+          export UNIT_TEST=True
+          export PYTHONPATH=$(pwd)/ush:$(pwd)/ush/python_utils/workflow-tools:$(pwd)/ush/python_utils/workflow-tools/src
+          python -m unittest -b tests/test_python/*.py
+
+      - name: Run python functional tests
+        run: |
+          micromamba activate srw_app
+          export CI=true
+          export PYTHONPATH=${PWD}/ush
+          python3 -m unittest -b tests/test_python/test_retrieve_data.py
```
gspetro-NOAA, line 58: This single file seems like a much more concise way to write the tests (rather than 3 yamls with repeated info). I know @christinaholtNOAA mentioned that the one downside here is that it serializes the tests and am curious whether anyone thinks this would be an issue? Seems like they're short tests, so I personally prefer the concision of one file but am curious what other reviewers think and what the likelihood is that these tests might be expanded in a way that makes serialization unwieldy.

.github/workflows/python_tests.yaml
```diff
@@ -0,0 +1,58 @@
...
+          cache-environment: true
+
+      - name: Checkout externals
+        run: |
+          ./manage_externals/checkout_externals ufs-weather-model workflow-tools
+
+      - name: Lint the python code
+        run: |
+          micromamba activate srw_app
+          export PYTHONPATH=$(pwd)/ush:$(pwd)/ush/python_utils/workflow-tools:$(pwd)/ush/python_utils/workflow-tools/src
```
gspetro-NOAA, line 40: @christinaholtNOAA Just wanted to check that the repeated env activation and PYTHONPATH exports are needed for each task. GHA doesn't carry over that info? Or is there a reason we don't want it to carry over?

devbuild.sh
```diff
@@ -187,11 +188,48 @@ while :; do
...
+    installer=Miniforge3-${os}-${hardware}.sh
+    curl -L -O "https://github.com/conda-forge/miniforge/releases/latest/download/${installer}"
+    bash ./${installer} -bfp "${CONDA_BUILD_DIR}"
+    rm ${installer}
+  fi
+
+  source ${CONDA_BUILD_DIR}/etc/profile.d/conda.sh
+  # Put some additional packages in the base environment on MacOS systems
+  if [ "${os}" == "MacOSX" ] ; then
+    mamba install -y bash coreutils sed
```
gspetro-NOAA, line 231: @christinaholtNOAA I see you're installing coreutils and sed instead of gsed. Previously @natalie-perlin was installing gsed, but I think there was some question is to whether this was necessary. Just making sure its omission/substitution of sed instead of gsed isn't an oversight.

devbuild.sh
```diff
@@ -187,11 +188,48 @@ while :; do
...
+  conda activate
+  if ! conda env list | grep -q "^srw_app\s" ; then
+    mamba env create -n srw_app --file environment.yml
+  fi
+  if ! conda env list | grep -q "^srw_graphics\s" ; then
+    mamba env create -n srw_graphics --file graphics_environment.yml
+  fi
+else
+  source ${CONDA_BUILD_DIR}/etc/profile.d/conda.sh
+  conda activate
```
gspetro-NOAA, line 248: @christinaholtNOAA Just wondering what `conda activate` does (as opposed to `conda activate env_name`).

.github/workflows/python_tests.yaml
christinaholtNOAA: These three arguments are:

```
  -l        Make bash act as if it had been invoked as a login shell
  -e        Exit immediately if a command exits with a non-zero status.
  -o pipefail    the return value of a pipeline is the status of
                           the last command to exit with a non-zero status,
                           or zero if no command exited with a non-zero status
```

.github/workflows/python_tests.yaml
christinaholtNOAA: We still need that for this PR, but I will remove it in a subsequent one. Here, we are still using the workflow-tools repository cloned as part of manage_externals.

Once I can replace that logic, I'll also remove this PYTHONPATH stuff.

devbuild.sh
christinaholtNOAA: I believe that gsed was required for using brew, as it was documented in the UG, but when installing with conda, it's sed. 

In this case, I think the conda version of coreutils (instead of brew) also gets us `readlink` instead of `greadlink`. We might be able to simplify some things in the workflow scripts with this new environment.

devbuild.sh
christinaholtNOAA: It activates the base environment so that you have access to all the packages installed in it. Those are needed for all the package installing and environment solving steps.

christinaholtNOAA:
If it's helpful to take a look at the rendered documentation, I've started building this branch on Read the Docs here: https://srw-app-fork.readthedocs.io/en/latest/index.html

docs/UsersGuide/source/BuildingRunningTesting/ContainerQuickstart.rst
```diff
@@ -271,9 +271,9 @@ The ``wflow_<platform>`` modulefile will then output instructions to activate th
 .. code-block:: console
 
    Please do the following to activate conda:
-       > conda activate workflow_tools
+       > conda activate srw_app
```
gspetro-NOAA: @christinaholtNOAA We might want to leave this chapter as-is because your instructions will only work if/when the containers are updated, and I don't think we do that automatically yet.

docs/UsersGuide/source/BuildingRunningTesting/VXCases.rst
```diff
@@ -93,7 +93,7 @@ First, navigate to the ``ufs-srweather-app/ush`` directory. Then, load the workf
 
 Users running a csh/tcsh shell would run ``source /path/to/etc/lmod-setup.csh <platform>`` in place of the first command above. 
 
-After loading the workflow, users should follow the instructions printed to the console. Usually, the instructions will tell the user to run ``conda activate regional_workflow``. 
+After loading the workflow, users should follow the instructions printed to the console. Usually, the instructions will tell the user to run ``conda activate srw_app``.
```
gspetro-NOAA, line 96: @christinaholtNOAA This is another one we should probably leave as-is. The VX chapter always uses the most recent release code because the develop branch is always changing, whereas the release provides a consistent baseline. Since the v2.2.0 release will have workflow_tools, this chapter probably should, too. (The SI team is about to do an update to bring it in line w/v2.2.0, so it won't say regional_workflow much longer).

MichaelLueken (commented):
@christinaholtNOAA - While attempting to build the SRW App using the Jenkins `.cicd/scripts/srw_build.sh` script, the build fails due to the `tests/build.sh` script looking for the built executables in `install_$compiler/exec`.  In `devbuild.sh`, the `BUILD_DIR` and `INSTALL_DIR` defaults have been moved to a point where these updated options have already been set, resetting these to the default settings.  This means that the built executables are in `exec`, rather than `install_$compiler/exec`.

To correct this behavior, please either move the `BUILD_DIR` and `INSTALL_DIR` default options back to lines 102 and 103 in `devbuild.sh`, or update the `tests/build.sh` script to look for the executables in `exec`, rather than `install_$compiler/exec`.

I also noted a minor spelling issue in `devbuild.sh` as well.

devbuild.sh
```diff
@@ -187,11 +188,48 @@ while :; do
...
+if [ "${BUILD_CONDA}" = "on" ] ; then
+  if [ ! -d "${CONDA_BUILD_DIR}" ] ; then
+    os=$(uname)
+    test $os == Darwin && os=MacOSX
+    hardware=$(uname -m)
+    installer=Miniforge3-${os}-${hardware}.sh
+    curl -L -O "https://github.com/conda-forge/miniforge/releases/latest/download/${installer}"
+    bash ./${installer} -bfp "${CONDA_BUILD_DIR}"
+    rm ${installer}
+  fi
```
mkavulich: It seems like this logic could break unpredictably if a new miniforge release has different installer script settings. Could we make this a static version that gets updated as needed?

docs/UsersGuide/source/BuildingRunningTesting/AQM.rst
```diff
@@ -68,7 +68,7 @@ If the SRW-AQM builds correctly, users should see the standard executables liste
    * - nexus
      - Runs the NOAA Emission and eXchange Unified System (:ref:`NEXUS <nexus>`) emissions processing system
 
-Load the ``workflow_tools`` Environment
+Load the ``srw_app`` Environment
```
mkavulich, line 71: I see that @gspetro-NOAA has already commented on this, but I'd like to go a bit further: I'd prefer to keep the environment named the same regardless of how it was installed; having two different environment names for the same functional thing makes the documentation potentially confusing.

Is the reason for this name change to avoid collisions with the existing, built-in `workflow_tools` environment? Would a better way be to check for an existing environment named `workflow_tools` and stop the custom installation if it already exists? If the user really wants to install their own they can specify a conda-dir to disambiguate from the existing system environment.

Actually, thinking about it more, since the existing `workflow_tools` environment is not in the default conda path, collisions shouldn't occur at all, right?

devbuild.sh
```diff
@@ -187,11 +188,48 @@ while :; do
   shift
 done
 
+# build conda and conda environments, if requested.
+set -x
```
mkavulich:
```suggestion
```

devbuild.sh
```diff
@@ -117,6 +114,7 @@ VERBOSE=false
 
 # Turn off all apps to build and choose default later
 DEFAULT_BUILD=true
+BUILD_CONDA="off"
```
mkavulich: There are some verification capabilities that currently don't run any of the UFS, UFS_UTILS, or UPP components, but do make use of the python environment. Could we add a simple flag here like "build_only_conda" or similar option so that we could use this new capability without needing to build all the rest of the code? This would be a great benefit to verification tasks on non-tier-1 machines!

devbuild.sh
christinaholtNOAA: I moved these down here because `${SRW_DIR}` can't be determined on a standard MacOS platform since `readlink` is unavailable by default. Once the conda base environment is installed, the utility will be available, making it more portable to use this script on MacOS without a prior step of users on macs to install supporting software manually.

christinaholtNOAA:
@MichaelLueken Thanks for reporting on the Jenkins issues! In the solution I just pushed, I left the variables where they were but did not overwrite the user-defined settings. Let me know if that's an acceptable alternative to your suggestions.

mkavulich (changes requested):
A few suggested changes, let me know if you have any questions. 

I was able to test this on a random Linux machine and all but three of the python unit tests passed. The three failures all gave the same message:

`ModuleNotFoundError: No module named 'scripts.templater'`

Do you have any thoughts on what might be causing those failures? It's not necessarily a dealbreaker but it seems to me like all python tests *should* pass even if there's no built code.

devbuild.sh
christinaholtNOAA: Good call.

docs/UsersGuide/source/BuildingRunningTesting/AQM.rst
christinaholtNOAA: I decided to go with a name change here since `srw_app` is now a bit more descriptive than `workflow_tools`. Technically `workflow_tools` is tied to the requirements we had for the `workflow-tools` repository when we first integrated with SRW. Now that it's evolved into its own python package, it seems more consistent to track to an environment name that not only has the `uwtools` package, but also a couple of others that are exactly what the SRW workflow needs.

With this new conda paradigm, we'll now have `srw_app` and `srw_graphics` as two environments that go with the version of SRW you are currently using.

As we deprecate the need for pre-installed, supported environments on Tier I platforms, I don't anticipate that we'd want to have both `workflow_tools` and `srw_app`; we'd just have the user-installed `srw_app` environment.

devbuild.sh
christinaholtNOAA: That's easy enough. I can add a `conda_only` target that will turn off the default build of app components in addition to turning on the conda build. I'd prefer not to turn off the default build when `conda` is specified because that drastically changes the original behavior of `devbuild.sh`.

docs/UsersGuide/source/BuildingRunningTesting/AQM.rst
christinaholtNOAA: I might also mention that with the deprecation of supported platform conda environments for SRW, we might want to consider making the conda installation default. Let's plan to bring this up in the code manager meeting since there are lots of pros/cons to each strategy.

christinaholtNOAA:
@mkavulich I've pushed a change to the README in the tests directory that might help with the failures you see for the python tests on a generic linux platform. Let me know if that does the trick.

graphics_environment.yml
```diff
@@ -0,0 +1,17 @@
...
+  - f90nml 1.4*
+  - jinja2=3.0*
+  - matplotlib=3.5*
+  - pillow=9.2*
+  - pygrib=2.1*
+  - pylint=2.16*
+  - pyproj=2.6*
+  - pytest=7.1*
+  - pyyaml=6.0*
+  - pyyaml=6.0*
```
gspetro-NOAA: It looks like `pyyaml=6.0*` is listed twice here.

gspetro-NOAA:
I tried to build on Derecho following the instructions in the [docs](https://srw-app-fork.readthedocs.io/en/latest/BuildingRunningTesting/BuildSRW.html#) that @christinaholtNOAA linked, and it claims to have built, but there were build errors in the printed log. 
```
gpetro@derecho5:/glade/work/gpetro/srwholt> ./devbuild.sh --platform=derecho conda
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100     9  100     9    0     0     46      0 --:--:-- --:--:-- --:--:--    46
./Miniforge3-Linux-x86_64.sh: line 1: Not: command not found
./devbuild.sh: line 209: conda/etc/profile.d/conda.sh: No such file or directory
./devbuild.sh: line 214: conda: command not found
./devbuild.sh: line 215: conda: command not found
./devbuild.sh: line 216: mamba: command not found
./devbuild.sh: line 218: conda: command not found
./devbuild.sh: line 219: mamba: command not found
PLATFORM(MACHINE)=derecho
WARNING: Setting default COMPILER=intel for new platform derecho
COMPILER=intel
MODULE_FILE=build_derecho_intel
Running "module reset". Resetting modules to system default. The following $MODULEPATH directories have been removed: None
... Load MODULE_FILE and create BUILD directory ...

Lmod is automatically replacing "intel/2023.0.0" with "intel-classic/2023.0.0".


Due to MODULEPATH changes, the following have been reloaded:
  1) netcdf/4.9.2

The following have been reloaded with a version change:
  1) hdf5/1.12.2 => hdf5/1.14.0


Currently Loaded Modules:
  1) cmake/3.26.3                    19) g2/3.4.5
  2) ncarenv/23.06              (S)  20) g2tmpl/1.10.2
  3) craype/2.7.20                   21) ip/3.3.3
  4) hpc/1.2.0                       22) sp/2.3.3
  5) intel-classic/2023.0.0          23) w3emc/2.9.2
  6) hpc-intel-classic/2023.0.0      24) gftl-shared/v1.5.0
  7) cray-mpich/8.1.25               25) yafyaml/v0.5.1
  8) hpc-cray-mpich/8.1.25           26) mapl/2.35.2-esmf-8.4.2
  9) ncarcompilers/1.0.0             27) nemsio/2.5.4
 10) mkl/2023.0.0                    28) sfcio/1.4.1
 11) jasper/2.0.25                   29) sigio/2.3.2
 12) zlib/1.2.11                     30) w3nco/2.4.1
 13) libpng/1.6.37                   31) wrf_io/1.2.0
 14) pio/2.5.10                      32) hdf5/1.14.0
 15) esmf/8.4.2                      33) netcdf/4.9.2
 16) fms/2023.01                     34) wgrib2/2.0.8
 17) bacio/2.4.1                     35) srw_common
 18) crtm/2.4.0                      36) build_derecho_intel

  Where:
   S:  Module is Sticky, requires --force to unload or purge
 ...
```

When I try to load the environment, it tells me to do `conda activate srw_app` but then fails to load. Perhaps the Derecho modulefiles need an update, too? Also possible it's human error, since I'm new to Derecho, but I wanted to try on a machine that hasn't already had some testing. 
```
gpetro@derecho5:/glade/work/gpetro/srwholt> module use modulefiles/
gpetro@derecho5:/glade/work/gpetro/srwholt> module load wflow_derecho 
Please do the following to activate conda:
       > conda activate srw_app
-bash: /glade/work/gpetro/srwholt/conda/etc/profile.d/conda.sh: No such file or directory
gpetro@derecho5:/glade/work/gpetro/srwholt> conda activate srw_app
If 'conda' is not a typo you can use command-not-found to lookup the package that contains it, like this:
    cnf conda
gpetro@derecho5:/glade/work/gpetro/srwholt> cnf conda
conda: searching ... 
Warning: no repos found - make sure your repositories are configured.
 conda: command not found 
```

MichaelLueken:
> @MichaelLueken Thanks for reporting on the Jenkins issues! In the solution I just pushed, I left the variables where they were but did not overwrite the user-defined settings. Let me know if that's an acceptable alternative to your suggestions.

@christinaholtNOAA Thank you very much for pushing your solution!  I was able to successfully build the SRW without issue.

gspetro-NOAA:
@christinaholtNOAA I have tried building your `build_conda` branch on Hera, Jet, and Orion, but there seems to be some issue. I used the following commands: 
```
git clone -b build_conda https://github.com/christinaholtNOAA/ufs-srweather-app.git srwholt
cd srwholt
./devbuild.sh --platform=name_of_machine conda
```
All the machines seem to give the same starting errors as on Derecho. I know it was building on Hera before, so maybe the fix for Jenkins broke the manual builds? If you want to check out my attempted build, it's at `/scratch2/NAGAPE/epic/Gillian.Petro/srwholt`. Interestingly, the build just stops on Orion, whereas on the other systems, it continues... not sure why there's a difference, but I think I prefer the Orion behavior. If someone runs the build command and goes to do something else while it builds, they might think the faulty build has been successful. I think that's a broader issue with SRW tho and not specific to this PR.

christinaholtNOAA:
My apologies. It looks like there's some issue with the curl command. I'm investigating and will get back to you!

christinaholtNOAA:
@gspetro-NOAA Thanks for the report on this...would you mind grabbing the latest to see if that fixes the issue you're seeing?

EdwardSnyder-NOAA:
With the latest commits, I was able to build the SRW App using the `devbuild.sh` script on PW GCP and Derecho without any problems. 

@gspetro-NOAA when rerunning the `devbuild.sh` file, make sure you remove the `conda` directory under `ufs-srweather-app`. The script only downloads conda if the `conda` directory isn't under `ufs-srweather-app`. See [line 199](https://github.com/ufs-community/ufs-srweather-app/pull/938/files#diff-6793940b2bd1bdc88e7b09d2510f19b344661a6e4459955a9a56e1655733063dR199) of the `devbuild.sh` file. I've ran into this problem myself in the past.

christinaholtNOAA:
I've just pushed changes and am currently re-running fundamental tests on Hera (the build went as expected). Those changes address Gillian's latest feedback and the transition to installing conda with no conda target by default. I changed that default in the test build script, too.

Now, the conda installation is only turned off for the AQM buid.

gspetro-NOAA:
@christinaholtNOAA I decided to check whether `./devclean.sh -a` works to clean up the conda/SRW App build, and it looks like it doesn't (I still have a `conda_loc` directory in my SRW App repo; haven't looked for less obvious conda remnants). I will go ahead and test your updates, but it would be nice to put in a fix for `devclean.sh` so that the PR doesn't break that functionality.

docs/UsersGuide/source/BuildingRunningTesting/RunSRW.rst
```diff
@@ -111,23 +111,37 @@ The first two steps depend on the platform being used and are described here for
...
 
-.. _Load-WF-L1:
 
-Loading the Workflow Environment on Level 1 Systems
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Installing miniconda and creating environments
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In addition to building the SRW App components, the ``devbuild.sh`` script will install miniconda,
+specifically miniforge, and install the SRW-required environments if the ``conda`` target is
```
gspetro-NOAA: Should probably remove mention of "conda" if "default" or nothing is now appropriate.

gspetro-NOAA: Although I'm actually wondering if this section should be removed? The information is already in the BuildSRW chapter. We probably don't want people to rebuild by accident.

docs/UsersGuide/source/BuildingRunningTesting/AQM.rst
christinaholtNOAA: The result from the developer meeting was to go with conda installed by default with no target. It is turned off by default when building AQM.

docs/UsersGuide/source/BuildingRunningTesting/RunSRW.rst
christinaholtNOAA: I will remove the whole section. It is kind of in an odd spot now that this happens in devbuild.sh.

gspetro-NOAA:
Conda build looks good on Derecho! Ran `grid_RRFS_CONUScompact_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16` last night. Will try to run fundamental tests on another platform since Derecho is down for maintenance this week. This looks promising tho! :)

gspetro-NOAA:
Hmmm... running into some difficulties on Orion & Hercules. On Orion, it seemed to build fine and load the env, but I get a ModuleNotFound error when trying to run tests: 
```
(srw_app) Orion-login-4[8] gpetro$ ./run_WE2E_tests.py -t fundamental -m orion -a epic
  File "/work/noaa/epic/gpetro/srwholt-orion/tests/WE2E/./run_WE2E_tests.py", line 13, in <module>
    from generate_FV3LAM_wflow import generate_FV3LAM_wflow
  File "/work/noaa/epic/gpetro/srwholt-orion/tests/WE2E/../../ush/generate_FV3LAM_wflow.py", line 38, in <module>
    from scripts.templater import set_template
ModuleNotFoundError: No module named 'scripts'
```
On Hercules, when I try to run manage_externals, it can't seem to find python at all: 
```
hercules-login-1[5] gpetro$ ./manage_externals/checkout_externals
/usr/bin/env: ‘python’: No such file or directory
```
I saw that in the `build_orion_intel` file, line 13 `load("stack-python/3.9.7")` was removed, but `build_hercules_intel` still seems to have a similar line (`load("stack-python/3.9.14")`). Maybe it should be removed? 
Not sure what to do about the Orion testing error tho.

MichaelLueken:
> 
> On Hercules, when I try to run manage_externals, it can't seem to find python at all:
> 
> ```
> hercules-login-1[5] gpetro$ ./manage_externals/checkout_externals
> /usr/bin/env: ‘python’: No such file or directory
> ```
> 
> I saw that in the `build_orion_intel` file, line 13 `load("stack-python/3.9.7")` was removed, but `build_hercules_intel` still seems to have a similar line (`load("stack-python/3.9.14")`). Maybe it should be removed? Not sure what to do about the Orion testing error tho.

@gspetro-NOAA - I can't speak on the issue you encountered on Orion, but for both Hercules and Gaea C5, you will need to have a version of python loaded in your environment before you will be able to use `./manage_externals/checkout_externals`.  If you add `module load python` to your .`bashrc` or equivalent file on these machines, or use `module load python` before using `./manage_externals/checkout_externals`, then you should be able to checkout the necessary SRW App components.

mkavulich:
> @gspetro-NOAA - I can't speak on the issue you encountered on Orion, but for both Hercules and Gaea C5, you will need to have a version of python loaded in your environment before you will be able to use `./manage_externals/checkout_externals`. If you add `module load python` to your .`bashrc` or equivalent file on these machines, or use `module load python` before using `./manage_externals/checkout_externals`, then you should be able to checkout the necessary SRW App components.

I would advise against loading modules in .bashrc or other login scripts, as this can cause unexpected behavior for submitted jobs. A safer approach would be to add python as a default module: from a fresh login, `module load python` (or whatever specific module you'd like), then `module save`.

gspetro-NOAA:
@mkavulich @MichaelLueken The `module load python` / `module save` approach seems to have worked! (didn't try modifying .bashrc for the reasons stated). That said, it seems problematic to me that users would need to take any additional step on a Level 1 system. I will add a note to the release docs and pull it into develop post-release. 
On a related note, I discovered that Orion runs just fine when I source `lmod-setup.sh` before loading the environment. Most systems don't seem to require that, but a few do (I was aware of Gaea and Odin only). We should also document that. ;) Do you know of other systems that require this now?
@christinaholtNOAA I ran the fundamental tests on Orion and all passed except for `grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_GFS_v16`. I'll take a look later to see why.

mkavulich (approved):
Thanks for making those changes, I confirmed the unit tests now pass on the tested Linux platform.

tests/README.md
```diff
@@ -46,10 +46,12 @@ The unit tests in the test_python/ directory test various parts of the workflow
 
 ### Set PYTHONPATH
 
-First, you will need to set the PYTHONPATH environment variable to include the ush/ directory:
+First, you will need to set the PYTHONPATH environment variable to include the ush/ directory and
+the a few of the workflow-toools subdirectories. From the top level of the ufs-srweather-app clone
```
MichaelLueken: Minor grammatical issues.

```suggestion
a few of the workflow-tools subdirectories. From the top level of the ufs-srweather-app clone
```

docs/UsersGuide/source/BuildingRunningTesting/BuildSRW.rst
```diff
@@ -200,6 +200,19 @@ On Level 1 systems for which a modulefile is provided under the ``modulefiles``
 
 where ``<machine_name>`` is replaced with the name of the platform the user is working on. Valid values include: ``cheyenne`` | ``gaea`` | ``hera`` | ``jet`` | ``linux`` | ``macos`` | ``noaacloud`` | ``orion`` 
 
+Starting with SRW v2.2.0 will install miniconda and SRW environments as part
```
MichaelLueken: Since this update won't be in the SRW v2.2.0 release, should this line read:

```suggestion
Directly following the release of SRW v2.2.0, the App will install miniconda and SRW environments as part
```

docs/UsersGuide/source/BuildingRunningTesting/AQM.rst
MichaelLueken, line 71: Since the conda installation is turned off for AQM, should the environments still be `workflow_tools`, rather than `srw_app`?

devbuild.sh
```diff
@@ -215,9 +256,11 @@ if [ "${APPLICATION}" = "ATMAQ" ]; then
...
     BUILD_NEXUS="on"
     BUILD_AQM_UTILS="on"
+    BUILD_CONDA="off"
   fi
   if [ "${PLATFORM}" = "wcoss2" ]; then
     BUILD_POST_STAT="on"
+    BUILD_CONDA="off"
   else
     BUILD_POST_STAT="off"
   fi
```
MichaelLueken: While building the SRW App for AQM, I noted that the conda environment is still being built, even though `BUILD_CONDA` is set to off for AQM.  Maybe lines 254 - 267 (or, at least, the lines setting `BUILD_CONDA` to off) need to be moved to before line 195, since `BUILD_CONDA` is set to on by default, so the conda environment will already be built by the time `BUILD_CONDA` is set to off on line 259.

devbuild.sh
```diff
@@ -187,11 +192,47 @@ while :; do
...
+  conda activate
+  if ! conda env list | grep -q "^srw_app\s" ; then
+    mamba env create -n srw_app --file environment.yml
+  fi
+  if ! conda env list | grep -q "^srw_graphics\s" ; then
+    mamba env create -n srw_graphics --file graphics_environment.yml
+  fi
+else
+  source ${CONDA_BUILD_DIR}/etc/profile.d/conda.sh
+  conda activate
```
MichaelLueken, line 248: Since we don't want to build conda for AQM, should the noted else statement (line 219) and the following two lines (lines 220-221) be removed?  Otherwise, the `devbuild.sh` script will attempt to activate the `srw_app` environment.

devbuild.sh
```diff
@@ -187,11 +192,47 @@ while :; do
...
+fi
+
+# Conda environment should have linux utilities to perform these tasks on macos.
+SRW_DIR=$(cd "$(dirname "$(readlink -f -n "${BASH_SOURCE[0]}" )" )" && pwd -P)
+MACHINE_SETUP=${SRW_DIR}/src/UFS_UTILS/sorc/machine-setup.sh
+BUILD_DIR="${BUILD_DIR:-${SRW_DIR}/build}"
+INSTALL_DIR=${INSTALL_DIR:-$SRW_DIR}
+CONDA_BUILD_DIR="$(readlink -f "${CONDA_BUILD_DIR}")"
+echo ${CONDA_BUILD_DIR} > ${SRW_DIR}/conda_loc
```
MichaelLueken, line 257: A `conda_loc` file will be created, regardless of whether or not `BUILD_CONDA` is on or off.  This file should only be created when `BUILD_CONDA` is on.

MichaelLueken (changes requested):
@christinaholtNOAA - I noted a few grammatical issues and have pointed them out in this latest review.

There are also issues with the AQM.  Since the `BUILD_CONDA` is off for AQM, the `AQM.rst` chapter shouldn't switch the conda environment from `workflow_tools` to `srw_app`.  Additionally, due to the way that the `devbuild.sh` script is set up, even though `BUILD_CONDA` is set to off for AQM (and WCOSS2, for that matter), the `srw_app` conda environment will still be built.  The necessary logic to set `BUILD_CONDA` to off needs to be added before line 195.  Other issues associated with conda related features in `devbuild.sh` have also been noted in this review.

modulefiles/tasks/hera/python_srw.lua
```diff
@@ -1,4 +1,2 @@
-prepend_path("MODULEPATH","/scratch1/NCEPDEV/nems/role.epic/miniconda3/modulefiles")
-load(pathJoin("miniconda3", os.getenv("miniconda3_ver") or "4.12.0"))
-
-setenv("SRW_ENV", "workflow_tools")
+load("conda")
+setenv("SRW_ENV", "srw_app")
```
MichaelLueken: While only noting this issue on Hera, it will be present on all other platforms as well, this modification requires that `BUILD_CONDA` is set to true.  Otherwise, `conda` isn't present and neither is the `srw_app` environment.  This will cause the AQM WE2E test - `aqm_grid_AQM_NA13km_suite_GFS_v16` - to fail.  The `modulefiles/tasks/${machine}/python_srw.lua` files will need to be updated so that they can be used with the `srw_app` environment, and the current `workflow_tools` environment.

modulefiles/wflow_hera.lua
```diff
@@ -7,12 +7,10 @@ whatis([===[Loads libraries needed for running the UFS SRW App on Hera ]===])
 
 load("rocoto")
 load("set_pythonpath")
-
-prepend_path("MODULEPATH","/scratch1/NCEPDEV/nems/role.epic/miniconda3/modulefiles")
-load(pathJoin("miniconda3", os.getenv("miniconda3_ver") or "4.12.0"))
+load("conda")
```
MichaelLueken, line 10: Similar to the `modulefiles/tasks/${machine}/python_srw.lua` files, the `modulefiles/wflow_$machine.lua` files will need to be updated to allow the AQM test to run.  If `BUILD_CONDA` is set to off, as is the case for AQM, then neither `conda` nor `srw_app` will be available, leading to the failure of the AQM test.

MichaelLueken (commented):
@christinaholtNOAA - While attempting to run the AQM WE2E test, `aqm_grid_AQM_NA13km_suite_GFS_v16`, the test failed because there was no conda module to load.  Both the `modulefiles/wflow_${machine}.lua` files and the `modulefiles/tasks/${machine}/python_srw.lua` files will need to be updated in such a way that they can be run using the `srw_app` conda environment, and the current `workflow_tools` environment.

modulefiles/build_orion_intel.lua
```diff
@@ -10,7 +10,6 @@ prepend_path("MODULEPATH", "/work/noaa/da/role-da/spack-stack/modulefiles")
 
 load("stack-intel/2022.0.2")
 load("stack-intel-oneapi-mpi/2021.5.1")
-load("stack-python/3.9.7")
```
MichaelLueken, line 13: Why was the loading of `stack-python/3.9.7` removed from the `build_orion_intel.lua` modulefile?  While running the fundamental test suite on Orion, I saw the same behavior as @gspetro-NOAA, everything passed except for `grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_GFS_v16`.  This test failed in the `run_MET_Pb2nc_obs` task with the following error message:
`/work/noaa/epic/role-epic/spack-stack/spack-stack-1.4.1/envs/unified-env/install/intel/2022.0.2/met-10.1.1-eiz3e5i/bin/pb2nc: error while loading shared libraries: libpython3.9.so.1.0: cannot open shared object file: No such file or directory`
This error message usually occurs when a module that was loaded for the build is different from the module used for running.  When I added this load back into the modulefile, the test successfully ran through to completion.

MichaelLueken (commented):
@christinaholtNOAA - I found the same behavior that @gspetro-NOAA noted on Orion, the `grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_GFS_v16` test failed, while the rest of the fundamental tests passed.  The `run_MET_Pb2nc_obs` task failed with the following error message:
`/work/noaa/epic/role-epic/spack-stack/spack-stack-1.4.1/envs/unified-env/install/intel/2022.0.2/met-10.1.1-eiz3e5i/bin/pb2nc: error while loading shared libraries: libpython3.9.so.1.0: cannot open shared object file: No such file or directory`
Adding back the `load("stack-python/3.9.7")` into the `build_orion_intel.lua` modulefile allows the test to successfully run through to completion.

christinaholtNOAA:
@MichaelLueken After giving this AQM issue some thought, I've reached out to @chan-hoo via email to ask some questions about whether AQM can tolerate this sort of change in develop. I will come back to the solution once we've had a chance to work out those kinks. Sorry for the delay here, and I'm glad you caught this!

MichaelLueken:
@christinaholtNOAA - Given @chan-hoo's reply to your email, I was able to successfully build your build_conda branch on Hera with the `-a=ATMAQ` option and was able to run the `aqm_grid_AQM_NA13km_suite_GFS_v16` test:
```
----------------------------------------------------------------------------------------------------
Experiment name                                                  | Status    | Core hours used 
----------------------------------------------------------------------------------------------------
aqm_grid_AQM_NA13km_suite_GFS_v16                                  COMPLETE            1083.12
----------------------------------------------------------------------------------------------------
Total                                                              COMPLETE            1083.12
```
What you currently have should be fine, but we will need to make sure that nothing is done to build the `srw_app` conda environment on WCOSS2.

MichaelLueken:
> @christinaholtNOAA - Given @chan-hoo's reply to your email, I was able to successfully build your build_conda branch on Hera with the `-a=ATMAQ` option and was able to run the `aqm_grid_AQM_NA13km_suite_GFS_v16` test:
> 
> ```
> ----------------------------------------------------------------------------------------------------
> Experiment name                                                  | Status    | Core hours used 
> ----------------------------------------------------------------------------------------------------
> aqm_grid_AQM_NA13km_suite_GFS_v16                                  COMPLETE            1083.12
> ----------------------------------------------------------------------------------------------------
> Total                                                              COMPLETE            1083.12
> ```
> 
> What you currently have should be fine, but we will need to make sure that nothing is done to build the `srw_app` conda environment on WCOSS2.

@christinaholtNOAA - It should be noted that this test was ran using the current settings in `modulefiles/tasks/hera/miniconda_regional_workflow_cmaq.lua`.

I have attempted to run the AQM WE2E test once again, replacing:

```
prepend_path("MODULEPATH","/scratch1/NCEPDEV/nems/role.epic/miniconda3/modulefiles")
load(pathJoin("miniconda3", os.getenv("miniconda3_ver") or "4.12.0"))

setenv("SRW_ENV", "regional_workflow_cmaq")
```
with:
```
load("conda")
setenv("SRW_ENV", "srw_app")
```
This time, the test failed in the following tasks:
`point_source`, `nexus_emission_00`, `nexus_emission_01`, and `nexus_emission_02`

The following traceback is in the `point_source` log files:
```
  File "/scratch2/NAGAPE/epic/Michael.Lueken/ufs-srweather-app/sorc/AQM-utils/python_utils/stack-pt-merge.py", line 12, in <module>
    import netCDF4 as nc
  File "/scratch1/NCEPDEV/nems/role.epic/spack-stack/spack-stack-1.4.1/envs/unified-env/install/intel/2021.5.0/py-netcdf4-1.5.3-ofq7pt3/lib/python3.9/site-packages/netCDF4/__init__.py", line 3, in <module>
    from ._netCDF4 import *
ModuleNotFoundError: No module named 'netCDF4._netCDF4'
```
The following traceback is in the `nexus_emission_*` log files:
```
  File "/scratch2/NAGAPE/epic/Michael.Lueken/ufs-srweather-app/sorc/arl_nexus/utils/python/make_nexus_output_pretty.py", line 144, in <module>
    raise SystemExit(main(**parse_args()))
                     ^^^^^^^^^^^^^^^^^^^^
  File "/scratch2/NAGAPE/epic/Michael.Lueken/ufs-srweather-app/sorc/arl_nexus/utils/python/make_nexus_output_pretty.py", line 50, in main
    import netCDF4 as nc
  File "/scratch1/NCEPDEV/nems/role.epic/spack-stack/spack-stack-1.4.1/envs/unified-env/install/intel/2021.5.0/py-netcdf4-1.5.3-ofq7pt3/lib/python3.9/site-packages/netCDF4/__init__.py", line 3, in <module>
    from ._netCDF4 import *
ModuleNotFoundError: No module named 'netCDF4._netCDF4'
```
So, continuing to use the current `regional_workflow_cmaq` conda environment works, but updating to `srw_app` leads to failures due to missing netCDF4 in the conda environment.

christinaholtNOAA:
My apologies for the huge delay on getting back to the comments here. I have modified the code to install a `srw_aqm` environment when the AQM application is being built. I tested that on Hera and found that the test no longer worked -- data was unavailable. I'm now second-guessing that test though since it was an older version. I did confirm that the appropriate environment was being loaded for the tasks that were run.

I updated my branch to the top of develop, did a fresh build, and successfully ran the fundamental tests on Hera again.

christinaholtNOAA:
I'm actually going to re-try that AQM test now.

christinaholtNOAA:
I was able to more successfully run the AQM test and fixed the issue that @MichaelLueken reported above. My test is now failing in the `aqm_lbcs` task because it can't find data on Hera. Many other AQM tasks ran successfully and loaded the appropriate environment with NetCDF4.

scripts/exregional_make_grid.sh
```diff
@@ -278,6 +278,7 @@ generation executable (exec_fp):
 #
 # Call the python script to create the namelist file.
 #
+echo PYTHON: $(which python)
```
MichaelLueken: This appears to be a debugging print.  Should this be removed?

MichaelLueken (approved):
@christinaholtNOAA -

Thank you very much for addressing my concerns regarding WCOSS2 and the AQM WE2E test!  Unfortunately, it looks like the old WE2E staged data was overwritten with a much newer case, so the `aqm_lbcs` task will fail.  As you have noted, however, following your latest updates, all of the rest of the AQM-specific tasks appear to be running smoothly now, so I'm fine with these changes.  I have noted one update to the `scripts/exregional_make_grid.sh` script that appears to have been made for debugging purposes.  If this is the case, then it would be a good idea to go ahead and remove this line from the script.

I will go ahead and approve this work now and I will submit the Jenkins tests in the morning.

MichaelLueken (changes requested):
@christinaholtNOAA -

Unfortunately, while working through the conversations and resolving those that I started, I noted the failure of the `grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_GFS_v16` fundamental WE2E test on Orion.  A rerun of this test on Orion this morning shows that the test is still failing.  It should be noted that this is only an issue on Orion, the rest of the RDHPCS machines successfully run the fundamental WE2E tests without issue.

The test is failing in the `run_MET_Pb2nc_obs` task with the following error message in the log:

`/work/noaa/epic/role-epic/spack-stack/spack-stack-1.4.1/envs/unified-env/install/intel/2022.0.2/met-10.1.1-eiz3e5i/bin/pb2nc: error while loading shared libraries: libpython3.9.so.1.0: cannot open shared object file: No such file or directory`

This error message is indicating that python needs to be used while running the verification tasks.  Adding `load("stack-python/3.9.7")` back into `build_orion_intel.lua` allows the test to run through to completion without issue.

The current `develop` branch fundamental WE2E tests pass on Orion, so I will not be able to merge this PR until the issue with this test is resolved.  Once this correction has been made, I will re-approve this PR and submit the automated Jenkins tests.

christinaholtNOAA:
@MichaelLueken Thanks for all the help with debugging other platforms.

I've added back the stack-python that was removed for Orion, but did it in the run_vx local module file for just those tasks. I was noticing that the stack python environment was interacting poorly with the conda environment on some tasks (when netcdf is expected, for example), so want to limit the interaction between stack and conda environments as much as possible. It looks necessary for verification from this failure. 

If you don't mind re-running that test again, it would be super helpful. Thanks!

MichaelLueken:
Thanks, @christinaholtNOAA!  Rerunning the test now.

modulefiles/build_orion_intel.lua
MichaelLueken: The modification that you made at hash [387373e](https://github.com/ufs-community/ufs-srweather-app/pull/938/commits/387373e85b5fea32760138835b8bf109c158f81e) has corrected the failure of the `grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_GFS_v16` WE2E test on Orion.

Resolving conversation now.

MichaelLueken (approved):
@christinaholtNOAA -

Thank you very much for working with me to get these final changes in!  The `grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_GFS_v16` WE2E test is now successfully running on Orion:
```
----------------------------------------------------------------------------------------------------
Experiment name                                                  | Status    | Core hours used 
----------------------------------------------------------------------------------------------------
grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta    COMPLETE              10.41
nco_grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_timeoffset_suite_  COMPLETE              11.26
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2        COMPLETE               8.58
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v17_p8_plot  COMPLETE              16.12
grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_HRRR_suite_HRRR          COMPLETE              26.35
grid_SUBCONUS_Ind_3km_ics_HRRR_lbcs_RAP_suite_WoFS_v0              COMPLETE              13.75
grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_GFS_v16                COMPLETE              21.56
----------------------------------------------------------------------------------------------------
Total                                                              COMPLETE             108.03
```
Once again approving this PR.  I will go ahead and launch the Jenkins tests as well.  I'd like for @gspetro-NOAA to provide one last look to ensure that she is happy, then I'll be able to merge this PR.  Thanks again!

MichaelLueken:
@christinaholtNOAA -

The `Functional UnitTests` have failed on Hera GNU, Hera Intel, and Jet.  The failure appears to be due to the fact that the `.cicd/scripts/srw_unittest.sh` script, which runs the `Functional UnitTests` phase of the pipeline, occurs before the `Build` phase (no `srw_app` conda environment).  It looks like the `Functional UnitTests` phase will need to be moved to after the `Build` phase in `.cicd/Jenkinsfile` in order to ensure that the necessary conda environment is available before running the unit tests.

Additionally, the `Functional WorkflowTaskTests` phase in the pipeline, which is ran using the `.cicd/scripts/srw_ftest.sh` script, failed because the conda activate is hardwired into this script:
`conda activate workflow_tools`
Please replace `workflow_tools` with `srw_app` in this script.

MichaelLueken:
The We2E coverage tests were manually ran on Derecho and all successfully passed:
```
----------------------------------------------------------------------------------------------------
Experiment name                                                  | Status    | Core hours used 
----------------------------------------------------------------------------------------------------
custom_ESGgrid_IndianOcean_6km                                     COMPLETE              24.18
grid_RRFS_CONUS_13km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16_plot     COMPLETE              38.53
grid_RRFS_CONUS_25km_ics_NAM_lbcs_NAM_suite_GFS_v16                COMPLETE              46.45
grid_RRFS_CONUScompact_13km_ics_HRRR_lbcs_RAP_suite_HRRR           COMPLETE              44.70
grid_RRFS_CONUScompact_25km_ics_HRRR_lbcs_RAP_suite_RRFS_v1beta    COMPLETE              18.82
grid_SUBCONUS_Ind_3km_ics_HRRR_lbcs_HRRR_suite_HRRR                COMPLETE              41.38
nco_grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_timeoffset_suite_  COMPLETE              24.88
pregen_grid_orog_sfc_climo                                         COMPLETE              17.04
specify_template_filenames                                         COMPLETE              19.63
----------------------------------------------------------------------------------------------------
Total                                                              COMPLETE             275.61
```

MichaelLueken:
The only failures in the automated Jenkins tests came from Hera Intel:
```
----------------------------------------------------------------------------------------------------
Experiment name                                                  | Status    | Core hours used
----------------------------------------------------------------------------------------------------
custom_ESGgrid_Central_Asia_3km                                    COMPLETE              24.51
get_from_HPSS_ics_FV3GFS_lbcs_FV3GFS_fmt_grib2_2019061200          COMPLETE               6.04
get_from_HPSS_ics_GDAS_lbcs_GDAS_fmt_netcdf_2022040400_ensemble_2  COMPLETE             760.78
get_from_HPSS_ics_HRRR_lbcs_RAP                                    COMPLETE              13.83
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2        DEAD                   4.02
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16_plot     DEAD                   3.94
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_RAP_suite_RAP                 COMPLETE              10.43
grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v15p2        COMPLETE               6.22
grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2         DEAD                 112.91
grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16           COMPLETE             305.82
grid_RRFS_CONUScompact_3km_ics_HRRR_lbcs_RAP_suite_HRRR            COMPLETE             324.28
pregen_grid_orog_sfc_climo                                         COMPLETE               7.04
----------------------------------------------------------------------------------------------------
Total                                                              DEAD                1579.82
```
The use of rocotorewind/rocotoboot allowed the three tests to pass:
```
----------------------------------------------------------------------------------------------------
Experiment name                                                  | Status    | Core hours used 
----------------------------------------------------------------------------------------------------
custom_ESGgrid_Central_Asia_3km                                    COMPLETE              24.51
get_from_HPSS_ics_FV3GFS_lbcs_FV3GFS_fmt_grib2_2019061200          COMPLETE               6.04
get_from_HPSS_ics_GDAS_lbcs_GDAS_fmt_netcdf_2022040400_ensemble_2  COMPLETE             760.78
get_from_HPSS_ics_HRRR_lbcs_RAP                                    COMPLETE              13.83
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2        COMPLETE               9.03
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16_plot     COMPLETE              14.72
grid_RRFS_CONUS_25km_ics_FV3GFS_lbcs_RAP_suite_RAP                 COMPLETE              10.43
grid_RRFS_CONUS_25km_ics_GSMGFS_lbcs_GSMGFS_suite_GFS_v15p2        COMPLETE               6.22
grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v15p2         COMPLETE             236.20
grid_RRFS_CONUS_3km_ics_FV3GFS_lbcs_FV3GFS_suite_GFS_v16           COMPLETE             305.82
grid_RRFS_CONUScompact_3km_ics_HRRR_lbcs_RAP_suite_HRRR            COMPLETE             324.28
pregen_grid_orog_sfc_climo                                         COMPLETE               7.04
----------------------------------------------------------------------------------------------------
Total                                                              COMPLETE            1718.90
```
I would like to check with @gspetro-NOAA to ensure that all of her concerns have been addressed, then I will move forward with merging this work.

docs/UsersGuide/source/BackgroundInfo/TechnicalOverview.rst
```diff
@@ -94,6 +94,11 @@ Optional but recommended prerequisites for all systems:
 * Rocoto Workflow Management System (1.3.1)
 * Python packages ``scipy``, ``matplotlib``, ``pygrib``, ``cartopy``, and ``pillow`` for graphics
 
+.. note::
+  The SRW build will install conda, and necessary conda environments for running the workflow and the
+  graphics automatically when the ``conda`` target is included at build time.
```
gspetro-NOAA: @christinaholtNOAA Is this note still relevant? I thought the conda target wasn't required any more?

docs/UsersGuide/source/BuildingRunningTesting/ContainerQuickstart.rst
```diff
@@ -271,9 +271,9 @@ The ``wflow_<platform>`` modulefile will then output instructions to activate th
 .. code-block:: console
 
    Please do the following to activate conda:
-       > conda activate workflow_tools
+       > conda activate regional_workflow
 
-then the user should run ``conda activate workflow_tools``. This will activate the ``workflow_tools`` conda environment. The command(s) will vary from system to system, but the user should see ``(workflow_tools)`` in front of the Terminal prompt at this point.
+then the user should run ``conda activate regional_workflow``. This will activate the ``regional_workflow`` conda environment. The command(s) will vary from system to system, but the user should see ``(regional_workflow)`` in front of the Terminal prompt at this point.
```
gspetro-NOAA: @christinaholtNOAA This was updated to `workflow_tools` ahead of the release, so you can go ahead and change it back to that.

docs/UsersGuide/source/BackgroundInfo/TechnicalOverview.rst
christinaholtNOAA: Good catch!

docs/UsersGuide/source/BuildingRunningTesting/ContainerQuickstart.rst
christinaholtNOAA: Sure thing!

gspetro-NOAA (approved):
Looks good to me! I ran the fundamental tests on Jet for good measure, and they pass. :)

PR #951, opened 2023-10-20
AQM ecflow workflow for running test parallel

lgannoaa:
As requested by Jianping. This draft PR is created to show the file change made for ecflow workflow from the implementation package. It is by design, remaining in draft mode.
This draft PR is NOT the final for implementation because the ecflow is still under test.


lgannoaa:
Development continue. This draft is now closed.

PR #1266, opened 2025-04-09
[release/public-v3.0.0] Added CCPP updates to RTD index so it tracks

jdkublnick:
Quick Update to get CCPP info into RTD


MichaelLueken (approved):
I now see the CCPP Updates page in RTD.  Approving.

ulmononian:
thanks for the quick fix here guys

